PR #23, opened 2020-10-29
Switch to AGPL 3.0

jack-signal:
Also adds license headers onto the Swift code, so I think after this is merged we are good to switch this public.


teeler:
Just curious but why was this done?

PR #41, opened 2020-11-09
Fix a new clippy warning from nightly

jack-signal:


jrose-signal (approved):

jack-signal:
Oh looks like nightly has a bunch of new clippys for us, updating locally ...

jrose-signal (approved):

PR #42, opened 2020-11-09
Pin nightly to a specific date

jack-signal:
I picked 11/1 simply because that was what has been used for Java builds so far.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -16,11 +16,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Install nightly rust
+    - uses: actions/checkout@v2
+
+    - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: nightly
-        override: true
         profile: minimal
 
     - name: Install Android toolchains
@@ -28,8 +28,6 @@ jobs:
 
     - name: Install NDK tool
       run: cargo install cargo-ndk
-
-    - uses: actions/checkout@v2
 
     - name: Build libsignal-client-java
       run: ./gradlew build
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,7 +20,6 @@ jobs:
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
-        override: true
         profile: minimal
         components: rustfmt, clippy
     - name: Check for duplicate dependencies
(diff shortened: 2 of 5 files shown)


jack-signal:
Oh fmt and clippy aren't available for this nightly, probably need those... will pick another date

jack-signal:
Per https://rust-lang.github.io/rustup-components-history/index.html 2020-11-09 is the only day this week that has had all components available.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly
+nightly-2020-11-01
```
jrose-signal: I don't think the GitHub Action is respecting this, but it shouldn't be too hard to read from this file to do that (cf. me [doing the same for Node](https://github.com/signalapp/libsignal-client/pull/20/files#diff-4e88a8cc63d2540befa3412886ac9b9db5cd10bce852960298517ea4bd7f7bb7R26-R31)).

.github/workflows/java.yml
```diff
@@ -19,7 +19,6 @@ jobs:
     - name: Install nightly rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: nightly
```
jrose-signal, line 22: Oh, that's great that that just works! I didn't even check.

jrose-signal (approved):

.github/workflows/java.yml
jack-signal: It's documented as working for the action, but isn't finding the `rust-toolchain` file for some reason, further debugging required.

jrose-signal: Looks like we've just got the steps in the wrong order—checkout first, then install. We haven't had to do that before now.

jack-signal: I wasn't able to get it to read a rust-toolchain even with a simple repo so I opened an issue with actions-rs https://github.com/actions-rs/toolchain/issues/123. I'm going to wait to hear back from them, or if nothing then I'll try the approach you used for node.

.github/workflows/rust.yml
```diff
@@ -16,13 +16,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Install nightly rust
+    - uses: actions/checkout@v2
+    - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
         override: true
```
jrose-signal, line 23: Nitpick: Missed `override: true` (though it shouldn't matter)

PR #45, opened 2020-11-11
Fixes for Android

jack-signal:
- Allow info to be null/empty in HKDF arguments. This is used during KBS attestation.
- Allow creating an ECPublicKey from the un-typed key bytes, also used when talking to KBS.


java/java/src/main/java/org/whispersystems/libsignal/ecc/ECPublicKey.java
```diff
@@ -24,6 +24,13 @@ public ECPublicKey(byte[] serialized) {
     this.handle = Native.ECPublicKey_Deserialize(serialized, 0);
   }
 
+  static public ECPublicKey fromPublicKey(byte[] key) {
```
jrose-signal, line 27: I'd suggest putting "Bytes" in the name to mirror your accessor. (Alternately, it could be `fromDjbPublicKey` or something.)

jrose-signal (approved):

PR #47, opened 2020-11-11
Fix CocoaPod build to use rust-toolchain

jrose-signal:
Also stop checking out libsignal-ffi.git for no reason.


jack-signal (approved):

jrose-signal:
I was hoping GitHub would handle that rename. Oh well.

PR #49, opened 2020-11-12
Automatically generate Native.java from the exported symbols of the Rust bridge

jack-signal:
Working on the sealed sender code made me realize how obnoxious it is to keep Native.java vs JNI lib.rs in sync.

This is ugly, but it works.

The changes in bridge/jni just change ABI from "system" to "C" as cbindgen doesn't recognize "system" fns for whatever reason. The ABIs should be the same for our purposes.

### DIFF
--- a/rust/bridge/jni/bin/gen_java_decl.py
+++ b/rust/bridge/jni/bin/gen_java_decl.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+
+#
+# Copyright (C) 2020 Signal Messenger, LLC.
+# SPDX-License-Identifier: AGPL-3.0-only
+#
+
+import os
+import subprocess
+import re
+import sys
+
+our_abs_dir = os.path.dirname(os.path.realpath(__file__))
+
+cbindgen = subprocess.Popen(['cbindgen'], cwd=os.path.join(our_abs_dir, '..'), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+(stdout, stderr) = cbindgen.communicate()
+
+stdout = str(stdout.decode('utf8'))
+stderr = str(stderr.decode('utf8'))
+
+ignore_this_warning = re.compile("WARN: Can't find .*\. This usually means that this type was incompatible or not found\.")
+
+unknown_warning = False
+
+for l in stderr.split('\n'):
+    if l == "":
+        continue
+
+    if ignore_this_warning.match(l):
+        continue
+
+    print(l)
+    unknown_warning = True
+
+if unknown_warning:
+    sys.exit(1)
+
+java_decl = re.compile('([a-zA-Z]+) Java_org_signal_client_internal_Native_([A-Z][a-zA-Z]+)_1([A-Za-z0-9]+)\(JNIEnv .?env, JClass class_(, .*)?\);')
+
+def translate_to_java(typ):
+    # jobject is not given here; instead use a type
+    type_map = {
+        "void": "void",
+        "jstring": "String",
+        "JString": "String",
+        "jbyteArray": "byte[]",
+        "ObjectHandle": "long",
+        "jint": "int",
+        "jlong": "long",
+        "jboolean": "boolean",
+    }
+
+    if typ in type_map:
+        return type_map[typ]
+
+    # Assume anything prefixed with Java refers to an object
+    if typ.startswith('Java'):
+        return typ[4:]
+
+    raise Exception("Don't know what to do with a", typ)
+
+cur_type = None
+decls = []
+
+for line in stdout.split('\n'):
+    if line == '':
+        continue
+
+    match = java_decl.match(line)
+    if match is None:
+        raise Exception("Could not understand", line)
+
+    (ret_type, this_type, method_name, args) = match.groups()
+
+    # Add newlines between groups of functions for readability
+    if cur_type is None or this_type != cur_type:
+        decls.append("")
+        cur_type = this_type
+
+    java_fn_name = '%s_%s' % (this_type, method_name)
+    java_ret_type = translate_to_java(ret_type)
+    java_args = []
+
+    if args != None:
+        for arg in args.split(', ')[1:]:
+            (arg_type,arg_name) = arg.split(' ')
+            java_arg_type = translate_to_java(arg_type)
+            java_args.append('%s %s' % (java_arg_type, arg_name))
+
+    decls.append("  public static native %s %s(%s);" % (java_ret_type, java_fn_name, ", ".join(java_args)))
+
+template_file = open(os.path.join(our_abs_dir, 'Native.java.in')).read()
+
+contents = template_file.replace('\n  // INSERT DECLS HERE', "\n".join(decls))
+
+native_java = os.path.join(our_abs_dir, '../../../../java/java/src/main/java/org/signal/internal/Native.java')
+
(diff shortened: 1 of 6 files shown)


rust/bridge/jni/bin/gen_java_decl.py
```diff
@@ -0,0 +1,152 @@
...
+    if typ == "ObjectHandle":
+        return "long"
+    if typ == "jint":
+        return "int"
+    if typ == "jlong":
+        return "long"
+    if typ == "jboolean":
+        return "boolean"
+    if typ == "JString":
+        return "String"
```
jrose-signal, line 109: Is there really both jstring and JString?

rust/bridge/jni/bin/gen_java_decl.py
```diff
@@ -0,0 +1,152 @@
...
+
+for line in stdout.split('\n'):
+    if line == '':
+        continue
+
+    match = java_decl.match(line)
+    if match is None:
+        raise Exception("Could not understand", line)
+
+    if cur_type is None or match.group(2) != cur_type:
```
jrose-signal, line 125: To make this code a little easier to read I suggest expanding the groups immediately:

```python
(ret_type, this_type, method_name, args) = match.groups()
```

rust/bridge/jni/bin/gen_java_decl.py
```diff
@@ -0,0 +1,152 @@
...
+contents += "}\n"
+
+native_java = os.path.join(our_abs_dir, '../../../../java/java/src/main/java/org/signal/internal/Native.java')
+
+if not os.access(native_java, os.F_OK):
+    raise Exception("Didn't find Native.java where it was expected")
+
+fh = open(native_java, 'w')
+fh.write(contents)
+fh.close()
```
jrose-signal, line 151: If this does get moved into the build system, it'd be easier to write to stdout and not worry about the path on disk.

rust/bridge/jni/bin/gen_java_decl.py
```diff
@@ -0,0 +1,152 @@
...
+# SPDX-License-Identifier: AGPL-3.0-only
+#
+
+import os
+import subprocess
+import re
+
+our_abs_dir = os.path.dirname(os.path.realpath(__file__))
+
+cbindgen = subprocess.Popen(['cbindgen'], cwd=os.path.join(our_abs_dir, '..'), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jrose-signal, line 14: Nitpick: I suggest passthrough stderr for this.

jrose-signal (commented):
Some thoughts, though we don't have to do them all immediately:

- Losing the types is rough. How do you feel about using `type JavaCiphertextMessage = jobject` and such in the Rust file so cbindgen can still present them?

- I'd personally prefer if the file template were separated from the generation script, `Native.java.in` with a `// SUBSTITUTE HERE` comment or similar.

- Thus far we've managed to avoid having out-of-date generated files. Can we call this in a build.rs for libsignal-jni? Or in the script that calls `cargo build -p libsignal-jni`?

rust/bridge/jni/bin/gen_java_decl.py
jack-signal: Yes though this something coming from the Rust JNI crate. Basically jstring == the underlying type from Java, and JString a wrapper struct with lifetime annotations. We have to use JString for arguments, but return jstrings.

rust/bridge/jni/bin/gen_java_decl.py
jack-signal: That makes a lot of noise

WARN: Can't find JClass. This usually means that this type was incompatible or not found.
... <497 more of these>


OTOH current version means all failures are silently hidden. How would you feel about capturing stderr into a pipe, but filtering it such that we ignore "This usually means that this type was incompatible or not found." but if any other output goes to stderr we halt immmediately?

jack-signal:
> Losing the types is rough. How do you feel about using type JavaCiphertextMessage = jobject and such in the Rust file so cbindgen can still present them?

Good call. Done.

> I'd personally prefer if the file template were separated from the generation script, Native.java.in with a // SUBSTITUTE HERE comment or similar.

Yeah that's better. Done.

rust/bridge/jni/bin/gen_java_decl.py
jack-signal: Done

rust/bridge/jni/bin/gen_java_decl.py
jrose-signal: What happens with cbindgen `[parse]` options `parse_deps = true` and `include = ["jni"]`? Is that sufficient to avoid the incompatible types?

jrose-signal: If not, yeah, filtering out this particular warning seems sad-but-reasonable.

rust/bridge/jni/bin/gen_java_decl.py
```diff
@@ -13,31 +13,45 @@
...
+        "ObjectHandle": "long",
+        "jint": "int",
+        "jlong": "long",
+        "jboolean": "boolean",
+        "JavaSessionStore": "SessionStore",
+        "JavaIdentityKeyStore": "IdentityKeyStore",
+        "JavaPreKeyStore": "PreKeyStore",
+        "JavaSignedPreKeyStore": "SignedPreKeyStore",
+        "JavaSenderKeyStore": "SenderKeyStore",
+        "JavaCiphertextMessage": "CiphertextMessage",
```
jrose-signal, line 57: Can we pattern-match these? "If the type starts with 'Java', drop that and assume it's valid."

rust/bridge/jni/bin/gen_java_decl.py
jack-signal: Done

rust/bridge/jni/bin/gen_java_decl.py
jack-signal: Makes sense. Each type still has to be named in Native.java.in so it's not much of a savings but every bit helps.

rust/bridge/jni/bin/gen_java_decl.py
jack-signal: Interesting idea. Doesn't seem to help though, I assume because `JClass` and such are not declared `repr(C)`

jrose-signal (approved):

java/java/src/main/java/org/whispersystems/libsignal/SessionCipher.java
```diff
@@ -70,10 +70,10 @@ public SessionCipher(SignalProtocolStore store, SignalProtocolAddress remoteAddr
    */
   public CiphertextMessage encrypt(byte[] paddedMessage) throws UntrustedIdentityException {
     synchronized (SESSION_LOCK) {
-       return Native.SessionCipher_EncryptMessage(paddedMessage,
-                             this.remoteAddress.nativeHandle(),
-                             sessionStore,
-                             identityKeyStore);
+       return (CiphertextMessage)Native.SessionCipher_EncryptMessage(paddedMessage,
```
jrose-signal, line 73: Oh, this cast shouldn't be needed anymore either.

jack-signal:
Rebased and squashed

PR #50, opened 2020-11-12
Update Cargo.lock

jack-signal:
Maybe we should start using `--frozen` in CI to catch this?


jrose-signal (approved):
Danger of long-running PRs: I forget what conventions we've changed since I've started. >< Thanks, Jack. (And I agree about --frozen.)

PR #51, opened 2020-11-12
Update GH actions to ignore changes to bridges

jack-signal:
Swift doesn't care if JNI bridge changed, similarly for Java and the C ABI

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -3,10 +3,10 @@ name: Node
 on:
   push:
     branches: [ master ]
-    paths-ignore: ['swift/**', '*.podspec', 'java/**', 'rust/bridge/**']
+    paths-ignore: ['swift/**', '*.podspec', 'java/**', 'rust/bridge/jni/**', 'rust/bridge/ffi/**']
   pull_request:
     branches: [ master ]
-    paths-ignore: ['swift/**', '*.podspec', 'java/**', 'rust/bridge/**']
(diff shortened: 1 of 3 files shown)


.github/workflows/node.yml
```diff
@@ -3,10 +3,10 @@ name: Node
 on:
   push:
     branches: [ master ]
-    paths-ignore: ['swift/**', '*.podspec', 'java/**']
+    paths-ignore: ['swift/**', '*.podspec', 'java/**', 'rust/bridge/**']
   pull_request:
     branches: [ master ]
-    paths-ignore: ['swift/**', '*.podspec', 'java/**']
+    paths-ignore: ['swift/**', '*.podspec', 'java/**', 'rust/bridge/**']
```
jrose-signal, line 9: rust/bridge/node/ is relevant, so we'll have to be more fine-grained. Or switch to an include/exclude model (https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns-1).

I do wish we could share variables across workflows, oh well.

jack-signal: Oh my mistake, I forgot there is a `rust/bridge/node` now. I should also update the Swift and Java builds to ignore that and modify this to ignore ffi/jni but not node.

jack-signal: Fixed

jrose-signal (approved):

PR #55, opened 2020-11-12
Fix some Swift lint warnings

jack-signal:
Can we configure the Swift linter to fail on all warnings?


jrose-signal:
Oops, yeah, worth adding `--strict` to our GitHub invocation.

jrose-signal (approved):

PR #57, opened 2020-11-12
Have each language ignore changes to other workflow scripts

jack-signal:
I noticed in #55 when I added `--strict` to the options in `swift.yml` it triggered builds of Node and Java when it didn't need to.

I am not loving the combinatorial explosion going on here but don't know of a good way to get around at. At least we are fairly unlikely to add many more supported languages.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -3,10 +3,10 @@ name: Java
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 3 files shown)


jrose-signal (approved):
I figure there's going to be some omissions all the time, but sure, we can add this one. Another thing to check would be if we can use symlinks for some of these files so that the things that change are in the usual nested directories, but that gets tricky with development on Windows (which we'll want for Desktop if nothing else).

PR #59, opened 2020-11-13
Swift: preserve files needed for the Rust build in podspec

jrose-signal:


jack-signal (approved):
IIUC this would have prevented using the pod outside of this repository? If so, is there a good way to test this in CI to prevent regressions?

PR #62, opened 2020-11-13
Import the libsignal-metadata-java tests

jack-signal:
Accidentally omitted from import in #21


jack-signal:
I deleted libsignal-metadata-tests SessionCipherTest.java which is an exact duplicate of libsignal-protocol-java. Otherwise unmodified.

PR #67, opened 2020-11-17
Enforce that our scripts are Shellcheck-clean

### LINKED ISSUE #37: Adopt shellcheck
Shellcheck is a nice static analyzer for shell scripts - we should run it in CI against all of our shell scripts.

jrose-signal:
Resolves #37.


jack-signal (approved):
Cocoapods build failure seems unrelated "WARN  | url: The URL (https://github.com/signalapp/libsignal-client) is not reachable."

PR #69, opened 2020-11-18
Use a helper to expect our "async" calls to complete synchronously

jrose-signal:
The underlying APIs are async only because it supports async callbacks; if all the callbacks complete synchronously, the top-level API will too. We can avoid any sort of thread-local overhead by just expecting that (i.e. panicking if it's not true).

The duplication here is unfortunate but we're still hoping to merge these libraries.


jack-signal (approved):
Nice

PR #72, opened 2020-11-18
Remove some unused protobufs and support for PendingKeyExchange

jack-signal:


jrose-signal:
Is it worth removing the PendingKeyExchange support in the same change?

jack-signal:
I did some digging in Android history. It appears KeyExchange was in fact used up until v2.4.0 when support was removed from the Android client. I think it is something vestigial from the old TextSecure crypto-over-SMS design, unclear if it was ever used in "Signal" proper. Anyway seems safe to remove to me but I'm going to cc @ehrenkret-signal

ehrenkret-signal (approved):
Yep; removing this should be fine. SMS encryption support was removed long ago: https://signal.org/blog/goodbye-encrypted-sms/

PR #74, opened 2020-11-18
Swift: expose SessionRecord.archiveCurrentState()

jrose-signal:
This is necessary for the "reset session" user-level operation. I believe it's also our first mutating operation!


jack-signal (approved):
On Android this operation is still done in Java with direct access to the protobufs, need to fix that.

PR #76, opened 2020-11-19
Begin merging bridge libraries (ffi and jni)

jrose-signal:
This starts with the simplest examples: macros that appear in both libraries without custom logic.


jrose-signal:
Hmm…the Java failure is a java.lang.UnsatisfiedLinkError, but none of the native declarations changed, and `nm` isn't showing any differences. Maybe it behaves differently on Linux?

rust/bridge/shared/src/jni/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    };
+}
+
+macro_rules! jni_bridge_deserialize {
+    ( $typ:ident::$fn:path as None ) => {};
+    ( $typ:ident::$fn:path as $jni_name:ident ) => {
+        paste! {
+            #[no_mangle]
+            pub unsafe extern "C" fn [<Java_org_signal_client_internal_Native_ $jni_name _1Deserialize>](
```
jack-signal, line 186: I did not know about the paste crate, that is handy.

jack-signal (commented):
Approach looks fine. I think before this is merged I'd like to be convinced it works outside of the limited cases covered so far. Even if just the currently macro-ed getters and leaving the one-off stuff for future work. [I don't see any reason why the getters wouldn't work in basically the same way you have done so far but if there is a problem I'd like to understand it before we commit to this]

rust/bridge/shared/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#![allow(clippy::missing_safety_doc)]
+#![deny(warnings)]
+
+use aes_gcm_siv::Aes256GcmSiv;
+use libsignal_protocol_rust::*;
+use std::convert::TryFrom;
+
+#[cfg(not(any(feature = "ffi", feature = "jni")))]
+compile_error!("Either feature \"ffi\" or \"jni\" must be enabled for this crate.");
```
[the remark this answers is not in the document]
jrose-signal: Yeah, I went back and forth on this too, but without it you can't do things like `cargo clippy --all`, because `libsignal-ffi` and `libsignal-jni` depend on different features. It's not such a burden to support in the end, and I don't think we're likely to ship a combined lib by accident since the ways we build for clients always use scripts.

jack-signal (approved):

rust/bridge/shared/src/lib.rs
```diff
@@ -0,0 +1,195 @@
+//
+// Copyright 2020 Signal Messenger, LLC.
+// SPDX-License-Identifier: AGPL-3.0-only
+//
+
+#![allow(clippy::missing_safety_doc)]
+// #![deny(warnings)]
```
jrose-signal, line 7: Oops, I'll put this back later.

PR #77, opened 2020-11-19
Sealed Sender

jack-signal:
Still todo:
 - [x] SealedSessionCipher tests in Rust
 - [x] Rewrite the Java tests so as to not require the protobufs be visible in Java anymore
 - [x] C ABI
 - [ ] Swift wrapper

One thing to note is the weird split interface where we have sealed sender decrypt -> UnidentifiedSenderMessageContent and then a second stage which goes from UnidentifiedSenderMessageContent to an actual message. That's to accommodate exception handling with the existing Java API. There is a big block of catch/throws like

```
    try {
       ...
    } catch (InvalidMessageException e) {
        throw new ProtocolInvalidMessageException(e, content.getSenderCertificate().getSender(), content.getSenderCertificate().getSenderDeviceId());                                                                                    
    } catch (InvalidKeyException e) {                                                                                                                                                                                                
      throw new ProtocolInvalidKeyException(e, content.getSenderCertificate().getSender(), content.getSenderCertificate().getSenderDeviceId());                                                                                      
 ...
```

But these are checked exceptions, which in Java have no relevance for runtime but are checked at compile time. And we do not have any throws declarations on our JNI wrappers (it would be nice to fix this but would require some additional cbindgen hacks, as I certainly do not want to give up the generated nature of `Native.java`) so the compiler will reject a JNI call because this block isn't suppose to throw any of the caught exceptions (even though it actually can, because we're invoking a JNI call which bypasses the compile time checked exception mechanism). As a result, the most expedient method was to first decrypt to a UnidentifiedSenderMessageContent  and then retain some of the decrypt logic in Java though mostly it's just invoking `SessionCipher` which will drop right back into Rust.

I think Swift can avoid all this though I haven't looked at what SignalMetadataKit is doing right now.


rust/protocol/src/sealed_sender.rs
```diff
@@ -0,0 +1,784 @@
...
+
+        let shared_secret = our_private.calculate_agreement(their_public)?;
+        let kdf = HKDF::new(3)?;
+        let derived_values = kdf.derive_salted_secrets(&shared_secret, &salt, &[], 96)?;
+
+        Ok(Self { derived_values })
+    }
+
+    fn cipher_key(&self) -> Result<&[u8]> {
+        Ok(&self.derived_values[32..64])
```
jrose-signal, line 582: The first 32 bytes are just thrown away?

rust/protocol/src/sealed_sender.rs
```diff
@@ -0,0 +1,784 @@
...
+    if (is_local_e164 || is_local_uuid) && usmc.sender()?.sender_device_id()? == local_device_id {
+        return Err(SignalProtocolError::SealedSenderSelfSend);
+    }
+
+    let mut rng = rand::rngs::OsRng;
+
+    let remote_address = usmc.sender()?.preferred_address(session_store, ctx).await?;
+
+    let message = match usmc.msg_type()? {
+        2 => {
```
jrose-signal, line 745: Absolutely these should use named constants. :-)

rust/bridge/jni/src/lib.rs
```diff
@@ -1662,3 +1653,197 @@ pub unsafe extern "C" fn Java_org_signal_client_internal_Native_SessionState_1In
...
+    contents: jbyteArray,
+) -> ObjectHandle {
+    run_ffi_safe(&env, || {
+        let sender = native_handle_cast::<SenderCertificate>(sender)?;
+        let contents = env.convert_byte_array(contents)?;
+        let msg_type = match msg_type {
+            1 => Ok(1u8),
+            2 => Ok(2u8),
+            3 => Ok(1u8),
+            x => Err(SignalJniError::IntegerOverflow(format!(
```
jrose-signal, line 1751: Nitpick: is this really an integer overflow?

jrose-signal (commented):
I figure this is a direct translation from the Java code in most cases, so I'm mostly assuming the logic is correct and looking instead at organization and such.

rust/protocol/src/sealed_sender.rs
jack-signal: Yes, which is odd and doesn't have any positive effect that I can see but that's how it works. I'll leave a comment to make it explicit.

(I considered throwing away the first 32 bytes in `new`, which would save a little memory at the cost of a second allocation, didn't seem worth it)

jrose-signal:
C stuff looks good. I can take care of the Swift side if you want.

swift/Sources/SignalClient/SealedSender.swift
```diff
@@ -0,0 +1,164 @@
...
+        try checkError(signal_sender_certificate_validate(&result, nativeHandle, trust_root.nativeHandle, time))
+
+        if result == 1 {
+            return true
+        } else {
+            return false
+        }
+    }
+
+    // FIXME missing signal_sender_certificate_preferred_address (is it needed?)
```
jrose-signal, line 142: The current logic I have is always preferring a UUID if present. Something like the preferred address logic may or may not make sense based on whether or not ProtocolAddress can be an opaque key.

jack-signal: Preferring UUID definitely makes sense to me. Possibly this logic is also wrong for Android, or at the least should be left client-specific in some way.

jack-signal:
I started on the Swift binding but it is incomplete, in particular missing sealed sender decryption function and could use some more thorough tests. It is also missing UnidentifiedSenderMessge/UnidentifiedSenderMessageContent, but it's quite possible that iOS doesn't need them anyway. Android doesn't directly touch them, I just didn't want to modify (at least at this juncture) the existing libsignal-metadata-java interface.

jrose-signal (approved):
I'll follow up with the logic moving sender cert verification out of `decrypt_to_usmc`.

PR #79, opened 2020-11-20
Fix compile issue when CARGO_BUILD_TARGET=aarch64-apple-ios

qhhonx:


jack-signal:
Thank you for the contribution, change itself looks good.

I'm looking into what our process will be for accepting external contributions to this repo so there may be some delay before we can merge.

jrose-signal:
Looks like we'll need you to sign our CLA (https://signal.org/cla/) but then we should be good to take this.

qhhonx:
> Looks like we'll need you to sign our CLA ([signal.org/cla](https://signal.org/cla/)) but then we should be good to take this.

Done.

jack-signal (approved):

PR #81, opened 2020-11-21
replace "Rust" with "Rustup" in build_helper's tips

DivineDominion:
I was confused when `swift/build_ffi.sh -d` failed with

```
error: rustup not found in PATH
note: Rust can be installed from https://rustup.rs/
```

... because I do have Rust installed, and the note to me read like `rustup.rs` would install rust itself. (I mean, the tool _does_, but that's not what it says :))

I'm not experienced with using Rust, so if anyone can add a hint why we rely on rustup rather than on e.g. the locally installed rust version, I'd happily add that to the printed note.

### DIFF
--- a/bin/build_helpers.sh
+++ b/bin/build_helpers.sh
@@ -15,7 +15,7 @@ check_rust() {
 
   if ! which rustup > /dev/null; then
     echo 'error: rustup not found in PATH' >&2
-    echo 'note: Rust can be installed from https://rustup.rs/' >&2
+    echo 'note: Rustup can be installed from https://rustup.rs/' >&2
     exit 1
   fi
 


jrose-signal:
Hmm. The root thing we're checking for here is that you have the correct pinned toolchain, and we're using a rustup-based [rust-toolchain file](https://rust-lang.github.io/rustup/overrides.html) for that. That's why we're expecting you to have `rustup`, and honestly this message is most likely to be seen from within Xcode from someone trying to build Signal-iOS who's never installed Rust before.

You're right, though, that a custom (nightly) Rust compiler could be used here as well. Maybe we should treat this as a warning instead of an error if `cargo` is still in your path, then?

jrose-signal:
(For the record, the reason we have a pinned toolchain is because we need to be on nightly for various features, but don't want to deal with supporting every possible nightly release. The build shouldn't stop you from using a different one, though.)

DivineDominion:
> (For the record, the reason we have a pinned toolchain is because we need to be on nightly for various features, but don't want to deal with supporting every possible nightly release. The build shouldn't stop you from using a different one, though.)

I admit that the conventions of pinning rustc versions in the community go over my head due to inexperience, but in general, I find this sounds sensible: offer to install a version that works, guaranteed, but also look for other `rustc` occurrences in PATH?

jrose-signal:
Resolved in #92.

PR #82, opened 2020-11-23
JNI: fix exception checks

jrose-signal:
One of the error states for `JNIEnv::call_method` is "threw an exception", so if we want to handle exceptions we have to do it before propagating the JNI error. On top of that, PreKeyStore and SignedPreKeyStore use exceptions to indicate the load of a pre-key that doesn't exist, so we need to handle that explicitly rather than treating it as a generic failure.

### DIFF
--- a/rust/bridge/jni/src/util.rs
+++ b/rust/bridge/jni/src/util.rs
@@ -315,12 +315,13 @@ pub fn jlong_from_u64(value: Result<u64, SignalProtocolError>) -> Result<jlong, 
     }
 }
 
-pub fn call_method_checked<'a>(
+pub fn call_method_with_exception_as_null<'a>(
     env: &JNIEnv<'a>,
     obj: impl Into<JObject<'a>>,
     fn_name: &'static str,
     sig: &'static str,
     args: &[JValue<'_>],
+    exception_to_treat_as_null: Option<&'static str>,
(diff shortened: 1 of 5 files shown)


jack-signal (approved):
lgtm

rust/bridge/jni/src/util.rs
```diff
@@ -343,9 +354,17 @@ pub fn exception_check(env: &JNIEnv, fn_name: &'static str) -> Result<(), Signal
...
+            if env.is_instance_of(throwable, exception_to_treat_as_null)? {
+                return Ok(JValue::Object(JObject::null()));
+            }
+        }
+
         let getmessage_sig = "()Ljava/lang/String;";
 
         let exn_type = exception_class_name(env, throwable).ok();
+        // Clear again in case we got an exception looking up the class name.
+        env.exception_clear()?;
```
alan-signal, line 367: If there was an error looking up the class name, this will eat that error? Same as the one below.

I'd be tempted to suggest falling over, otherwise we don't know there's any problem.

jrose-signal: We're dropping the precise exception, but we'll get "(None)" or whatever in the output for the class name, and a dummy string (see below) for the message. Honestly I expect these to never occur; I don't know how I would even write a test case for it.

PR #83, opened 2020-11-25
Swift: Enable line table debug info for the Rust libraries

jrose-signal:


jack-signal (approved):
Seems fine but can you give some background on "line tables" which is not a term I am familiar with. I understand this is enabling debug info in release mode which is totally sensible but wondering if there is some other implications here. And searching "Swift line tables" / "Rust line tables" brings up nothing relevant.

jrose-signal:
Line tables are the subset of debug info that allow for backtraces and breakpoints, but not inspecting variables or types. It tends to be a lot smaller and is less additional compile time (compared to "full" debug info). We could also just turn on full debug info in release builds, but backtraces were most important for me in debugging iOS issues.

PR #90, opened 2020-12-01
Swift: connect Rust's 'log' to SignalCoreKit's OWSLogger

jrose-signal:
We're not actually logging anything yet, but this will let us do so. The logging is initialized using a static constructor so that clients of SignalCoreKit don't have to do any additional setup. This requires an ObjC file instead of a Swift one. (When running as a Swift package, logs will just go to stderr via NSLog.)

### DIFF
--- a/rust/bridge/ffi/src/logging.rs
+++ b/rust/bridge/ffi/src/logging.rs
@@ -1,0 +1,105 @@
+//
+// Copyright 2020 Signal Messenger, LLC.
+// SPDX-License-Identifier: AGPL-3.0-only
+//
+
+use libc::c_char;
+use std::ffi::CString;
+
+#[repr(C)]
+pub enum LogLevel {
+    Error = 1,
+    Warn,
+    Info,
+    Debug,
+    Trace,
+}
+
+impl From<log::Level> for LogLevel {
+    fn from(level: log::Level) -> Self {
+        use log::Level::*;
+        match level {
+            Error => Self::Error,
+            Warn => Self::Warn,
+            Info => Self::Info,
+            Debug => Self::Debug,
+            Trace => Self::Trace,
+        }
+    }
+}
+
+impl From<LogLevel> for log::Level {
+    fn from(level: LogLevel) -> Self {
+        use LogLevel::*;
+        match level {
+            Error => Self::Error,
+            Warn => Self::Warn,
+            Info => Self::Info,
+            Debug => Self::Debug,
+            Trace => Self::Trace,
+        }
+    }
+}
+
+pub type LogCallback = extern "C" fn(
+    target: *const c_char,
+    level: LogLevel,
+    file: *const c_char,
+    line: u32,
+    message: *const c_char,
+);
+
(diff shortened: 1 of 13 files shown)


rust/bridge/ffi/src/logging.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    fn flush(&self) {
+        (self.flush)()
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn signal_init_logger(max_level: LogLevel, logger: FfiLogger) {
+    log::set_logger(Box::leak(Box::new(logger))).unwrap();
```
jack-signal, line 96: Can / should we detect multiple calls to this function, since each invocation causes a leak?

jrose-signal: `log::set_logger` will return an error if it's called multiple times, which I've turned into a panic with the `unwrap()`. For setup code like this I think that's reasonable, especially when there aren't even explicit calls to it on the client side (when using SignalCoreKit).

jack-signal: Ah, I had missed that. But unwinding a Rust panic into C/Swift will cause undefined behavior. We should either return an error code or do a whole process abort if we don't like what's happening.

[My preference would be to have this fn return an int error code if this is the 2nd registration, then have the static constructor abort if the function errors, since in that context there really isn't anything else to do. Just aborting here is fine to. But UB, no bueno]

jrose-signal: Ugh, right, I keep forgetting that Rust panics unwind by default instead of aborting by default (which we decided is the behavior we want for this library). Thanks for catching this.

Imperiopolis-Signal (approved):
LGTM

jack-signal (approved):

swift/Sources/SignalClient/Logging.m
```diff
@@ -0,0 +1,72 @@
...
+static void logMessage(const char *_Nonnull target,
+    SignalLogLevel level,
+    const char *_Nullable modulePath,
+    uint32_t line,
+    const char *_Nonnull message)
+{
+    // We're not using OWSLog* directly because we don't want log() to be the source of the log.
+    NSString *formattedMessage;
+    if (modulePath) {
+        formattedMessage = [NSString stringWithFormat:@"[%s:%u] %s", modulePath, line, message];
```
charlesmchen-signal, line 36: Maybe this is a waste of time, but there might be a perf benefit to exiting early _before_ `[NSString stringWithFormat:...]` if we're not going to log this `level`. There's functions like `ShouldLogVerbose()` and we could add a `ShouldLogLevel(level)` function too.

charlesmchen-signal (approved):
LGTM

swift/Sources/SignalClient/Logging.m
jrose-signal: I guess it would be easy enough to call `isEnabled` before doing any other work! Some of the Rust examples even do it that way.

PR #92, opened 2020-12-02
Tweak error message to account for Rust without rustup

jrose-signal:
Clarify the use of a specific Rust toolchain as discussed in #81.

### DIFF
--- a/bin/build_helpers.sh
+++ b/bin/build_helpers.sh
@@ -14,9 +14,15 @@ check_rust() {
   fi
 
   if ! which rustup > /dev/null; then
-    echo 'error: rustup not found in PATH' >&2
-    echo 'note: Rust can be installed from https://rustup.rs/' >&2
-    exit 1
+    if ! which cargo > /dev/null; then
+      echo 'error: cargo not found in PATH; do you have Rust installed?' >&2
+      echo 'note: we recommend installing Rust via rustup from https://rustup.rs/' >&2
+      exit 1
+    fi
+
+    echo 'warning: rustup not found in PATH; using cargo at' "$(which cargo)" >&2
+    echo 'note: this project uses Rust toolchain' "'$(cat ./rust-toolchain)'" >&2
+    return
   fi
 
   if [[ -n "${CARGO_BUILD_TARGET:-}" ]] && ! (rustup target list --installed | grep -q "${CARGO_BUILD_TARGET:-}"); then


jrose-signal:
@DivineDominion, does this assuage your concerns?

bin/build_helpers.sh
```diff
@@ -14,9 +14,15 @@ check_rust() {
   fi
 
   if ! which rustup > /dev/null; then
-    echo 'error: rustup not found in PATH' >&2
-    echo 'note: Rust can be installed from https://rustup.rs/' >&2
-    exit 1
+    if ! which cargo > /dev/null; then
+      echo 'error: cargo not found in PATH; do you have Rust installed?' >&2
+      echo 'note: Rust can be installed from https://rustup.rs/' >&2
```
DivineDominion, line 19: I think it can help to specify that this is not just _the_ or _a_ Rust installation, but that we recommend rustup in particular?

       echo 'note: We recommend installing Rust via rustup from https://rustup.rs/' >&2

DivineDominion:
I cannot give proper feedback because I'm not accustomed to the tooling :)

To me it seems like the nesteof `if`s should be reversed? I did have `cargo` on my Mac, but not `rustup`. So I would now get the new error

>       echo 'warning: rustup not found in PATH; using cargo at' "$(which cargo)" >&2 

... but is that actually true? Because the script aborts right afterwards. (Does another script fall back to `cargo`?)

And if I have `rustup`, it wouldn't need to check for cargo anymore, would it? Because it can install rust/cargo on its own?

jrose-signal:
> ... but is that actually true? Because the script aborts right afterwards. (Does another script fall back to `cargo`?)

Yeah, this is a file of helper functions, so the `return` just ends the `check_rust()` function, not the whole program.

> And if I have `rustup`, it wouldn't need to check for cargo anymore, would it? Because it can install rust/cargo on its own?

Yep, and it doesn't. There are three possibilities now (after the first block, which just adjusts PATH):

1. You have `rustup`: continue on to possibly checking for a cross-compilation target.
2. You don't have `rustup` but you do have `cargo`: emit the new warning and don't check anything else (relying on `cargo` to emit error messages about cross-compilation later).
3. You don't have `rustup` or `cargo`: point to https://rustup.rs and exit with an error.

I'll put in your adjusted wording, but I think these are a sensible set of cases!

jack-signal (approved):

DivineDominion:
Gotcha, thanks for the excellent support and response times!

PR #93, opened 2020-12-03
Fix clean target and use it during Docker builds

jack-signal:


java/build.gradle
```diff
@@ -25,6 +25,16 @@ task makeJniLibrariesDesktop(type:Exec) {
   commandLine './build_jni.sh', 'desktop'
 }
 
+task clean(type: Delete) {
+  description 'Clean JNI libs'
+  delete fileTree('./android/src/main/jniLibs') {
+    include '**/*.so'
+  }
+  delete fileTree('./java/src/main/resources') {
+    include '**/*.so'
```
jrose-signal, line 34: Nitpick: may as well include .dylib and .dll now too.

jrose-signal (approved):

PR #94, opened 2020-12-04
Swift: Remove 'throws', using the Java API as a reference 

jrose-signal:
While the FFI level of the Rust client is very conservative and can fail for things like null pointers, Swift restores a lot of invariants that C can't guarantee. In order to strike a balance between recovery and ergonomic use of the API, this commit removes 'throws' from APIs based on whether the corresponding Java API throws a checked exception.

Note that these changes do not *precisely* match the Java API; there are a few places where the Java API is declared as throwing because a precondition was previously lazily enforced, but libsignal-client enforces it eagerly. Example: `PreKeySignalMessage.getIdentityKey` (the validity of the identity key is now checked on deserialization).

Accessors that no longer throw have also been converted to properties.

Builds on #90 so that failures log in the same way as the rest of the iOS app.


swift/Sources/SignalClient/SenderKeyName.swift
```diff
@@ -6,23 +6,24 @@
...
-        signal_sender_key_name_destroy(handle)
+    internal override class func destroyNativeHandle(_ handle: OpaquePointer) -> SignalFfiErrorRef? {
+        return signal_sender_key_name_destroy(handle)
     }
 
     internal override class func cloneNativeHandle(_ newHandle: inout OpaquePointer?, currentHandle: OpaquePointer?) -> SignalFfiErrorRef? {
         return signal_sender_key_name_clone(&newHandle, currentHandle)
     }
 
+    // FIXME: Do we need both of these entry points?
```
jrose-signal, line 17: @jack-signal?

Imperiopolis-Signal (approved):

swift/Sources/SignalClient/SenderKeyName.swift
jack-signal: Not necessarily needed I guess. Depends on what is going to be convenient for iOS. My take would be to leave this one then have the version taking a `ProtocolAddress` call `self.init` so underlying logic isn't duplicated, because if some name isn't already in a `ProtocolAddress`, creating one involves several new heap allocations which would be otherwise completely avoidable.

jack-signal: (I don't have strong feelings about this and if it turns out that one form or another would be more convenient when we actually get to this in iOS it would be easy to just add it back)

jack-signal:
I'm fine with this for getters and destructors, or any functions which should always succeed modulo memory exhaustion. However I'm concerned that asserting the `xxx_new` functions never fail boxes us in in terms of checking arguments. For example we currently (in Rust) allow creating a `SenderKeyDistributionMessage` with a bogus chain key, say empty or 100 bytes. Such messages will fail to deserialize. But it would IMO be better if we did check those arguments in `new`, rather than allowing to create an invalid/unusable object. If we assert now these functions can't fail in Swift then we are faced with two unfortunate options - never improve our input checks, or for Swift accept that we'll immediately crash if an invalid argument is used. Similarly, we couldn't add logic to `ProtocolAddress` to verify UUID usage, etc.

jrose-signal:
That's a good point. I'll put it back on all the initializers and initializer-adjacent things.

swift/Sources/SignalClient/SenderKeyName.swift
jrose-signal: Hm, I like types but I also like avoiding heap allocations! I suppose another option is to make ProtocolAddress an eagerly-converted type like Fingerprint, since we're pretty sure it'll never grow another field or something. (But that does make it more awkward to validate the name.)

I'll go with your suggestion for now.

swift/Sources/SignalClient/Error.swift
```diff
@@ -92,3 +96,19 @@ internal func checkError(_ error: SignalFfiErrorRef?) throws {
         throw SignalError.unknown(errType, errStr)
     }
 }
+
+internal func assertNoError(_ error: SignalFfiErrorRef?) {
+    assertNoError { try checkError(error) }
```
charlesmchen-signal, line 101: Feel free to ignore, but I personally would rename this to something like `failOnError`.

charlesmchen-signal (approved):
lgtm

jrose-signal:
I'm going to merge this now to avoid merge conflicts. Updates for the iOS frameworks will come later today.

PR #95, opened 2020-12-04
Node: Switch to Neon's napi-runtime and JsBox

jrose-signal:
We lose the ability to define classes in Rust, but we probably want to be doing that in TypeScript anyway. A big advantage here is we're no longer dependent on the Node version.

### DIFF
--- a/node/index.ts
+++ b/node/index.ts
@@ -6,6 +6,16 @@
 import bindings = require('bindings'); // eslint-disable-line @typescript-eslint/no-require-imports
 import * as SignalClient from './libsignal_client';
 
-export const { PrivateKey } = bindings(
-  'libsignal_client'
-) as typeof SignalClient;
+const SC = bindings('libsignal_client') as typeof SignalClient;
+
+export class PrivateKey {
+  private readonly nativeHandle: SignalClient.PrivateKey;
+
+  constructor() {
+    this.nativeHandle = SC.PrivateKey_generate();
+  }
+
+  serialize(): Buffer {
+    return SC.PrivateKey_serialize(this.nativeHandle);
+  }
+}
--- a/rust/bridge/node/src/lib.rs
+++ b/rust/bridge/node/src/lib.rs
@@ -6,40 +6,48 @@
 use libsignal_protocol_rust::*;
 use neon::context::Context;
 use neon::prelude::*;
+use std::ops::Deref;
 
-fn borrow_this<'a, V, T, F>(cx: &mut MethodContext<'a, V>, f: F) -> T
-where
-    V: Class,
-    F: for<'b> FnOnce(neon::borrow::Ref<'b, &mut <V as Class>::Internals>) -> T,
-{
-    let this = cx.this();
-    cx.borrow(&this, f)
-}
+struct DefaultFinalize<T>(T);
 
-declare_types! {
-    pub class JsPrivateKey for PrivateKey {
-        init(_cx) {
-            // FIXME: guard against calling this directly
-            let mut rng = rand::rngs::OsRng;
-            let keypair = KeyPair::generate(&mut rng);
-            Ok(keypair.private_key)
-        }
+impl<T> Finalize for DefaultFinalize<T> {}
 
-        method serialize(mut cx) {
-            let bytes = borrow_this(&mut cx, |k| {
-                k.serialize()
-            });
-            // FIXME: check for truncation
-            let mut buffer = cx.buffer(bytes.len() as u32)?;
-            cx.borrow_mut(&mut buffer, |raw_buffer| {
-                raw_buffer.as_mut_slice().copy_from_slice(&bytes);
-            });
-            Ok(buffer.upcast())
-        }
+impl<T> Deref for DefaultFinalize<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
     }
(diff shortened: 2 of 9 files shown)


node/index.ts
```diff
@@ -6,6 +6,18 @@
 import bindings = require('bindings'); // eslint-disable-line @typescript-eslint/no-require-imports
 import * as SignalClient from './libsignal_client';
 
-export const { PrivateKey } = bindings(
-  'libsignal_client'
-) as typeof SignalClient;
+const SC = bindings('libsignal_client') as typeof SignalClient;
+
+const nativeHandle: unique symbol = Symbol();
```
EvanHahn-Signal, line 11: tl;dr: this symbol-based approach doesn't achieve true privacy. I'd probably just rely on TypeScript here.

---

I didn't know about `unique symbol`. Cool!

It seems like you're using this symbol as a way to achieve a private class field. This approach is reasonable, but I think there are two better options:

1. Give up on trying to achieve privacy at runtime, and only rely on TypeScript to achieve member privacy. This is what we've done in Signal Desktop and RingRTC, as far as I know. Here's a sketch of how that'd look:

   ```typescript
   export class PrivateKey {
     private readonly nativeHandle: SignalClient.PrivateKey;
     constructor() {
       this.nativeHandle = SC.PrivateKey_generate();
     }
     serialize(): Buffer {
       return SC.PrivateKey_serialize(this.nativeHandle);
     }
   }
   ```

   Of course, this doesn't achieve true privacy, because you could easily do something like this in JavaScript (or TypeScript with enough `any`s and `@ts-ignore`s):

   ```js
   const privateKey = new PrivateKey();
   console.log(privateKey.nativeHandle);
   ```

   Though it's harder, you can still "break into" your symbol-based solution:

   ```js
   const privateKey = new PrivateKey();
   const stolenSymbol = Object.getOwnPropertySymbols(privateKey)[0];
   console.log(privateKey[stolenSymbol]);
   ```
2. Use [JavaScript's private class fields](https://www.valentinog.com/blog/private/) to achieve privacy at runtime and compile time. Here's a sketch:

   ```typescript
   export class PrivateKey {
     readonly #nativeHandle: SignalClient.PrivateKey;
     constructor() {
       this.#nativeHandle = SC.PrivateKey_generate();
     }
     serialize(): Buffer {
       return SC.PrivateKey_serialize(this.#nativeHandle);
     }
   }
   ```

I don't feel strongly about which of those two solutions we use. I'd probably use the latter if I were the tzar of Signal's JavaScript code, but I'd realistically choose the first because we don't use private fields anywhere else (as far as I know).

If I'm mistaken and you're using this symbol for another reason, feel free to ignore this suggestion.

rust/bridge/node/src/lib.rs
```diff
@@ -6,40 +6,48 @@
...
+    let mut rng = rand::rngs::OsRng;
+    let keypair = KeyPair::generate(&mut rng);
+    Ok(boxed(cx, keypair.private_key).upcast())
+}
+
+#[allow(non_snake_case)]
+fn PrivateKey_serialize(mut cx: FunctionContext) -> JsResult<JsValue> {
+    let value = cx.argument::<DefaultJsBox<PrivateKey>>(0)?;
+    let bytes = value.serialize();
+    let mut buffer = cx.buffer(bytes.len() as u32)?;
```
EvanHahn-Signal, line 41: tl;dr, could you change this to `cx.array_buffer` if possible? Also means updating the TypeScript types.

---

Like much of the JavaScript world, Signal Desktop has historically been inconsistent about which byte sequence type to use. Should we use the [Node-native `Buffer`](https://nodejs.org/api/buffer.html)? The [standard `ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)? A [userland wrapper](https://github.com/signalapp/Signal-Desktop/blob/a5657473c3b2b36afa1cade62323ed88464be400/components/bytebuffer/dist/ByteBufferAB.js)?

I've asked the rest of the Desktop team what we'd like to standardize on. My preference is `ArrayBuffer` because it is (1) standard across runtimes, unlike `Buffer` (2) not tied to a specific "interpretation" of what the bytes represent, like `Uint8Array`.

For now, I think `ArrayBuffer` is a better choice. If you'd like to wait for our discussion to resolve, I'm fine to leave it as `Buffer`, but we might have to make a small breaking change.

EvanHahn-Signal (changes requested):
The JS parts look great to me other than my comments. I only skimmed the Rust code because I know so little about it.

node/index.ts
jrose-signal: I was thinking it made sense to use the same symbol for all of the native-wrapping objects, but I guess they're still opaque on the JavaScript side and not accessible on the Rust side (without the Rust side being the one to access them). And Neon is doing its own type-tagging too. I'll switch to a normal private property.

rust/bridge/node/src/lib.rs
jrose-signal: I think I had picked Buffer to match what (I think) libtextsecure is currently doing, but I'm happy to switch it now or later.

EvanHahn-Signal: `Buffer` is all good!

jack-signal (approved):

PR #96, opened 2020-12-04
Link to Swift in the README

flipsasser:
Otherwise you just have to scroll back up, but this is Markdown and we're all pro users here

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ $ make java_test
 
 ## Swift
 
-To learn about the Swift build process see ``swift/README.md``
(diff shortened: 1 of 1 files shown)


jrose-signal:
Ah, I didn't know GitHub would automatically adjust the link based on where you were viewing the readme! I'm not sure I agree with "we're all pro users here" but if anything that's more reason to be as expressive as possible with links. Thank you!

flipsasser:
Well maybe we're not all pro users of the hidden intricacies of Github-flavored Markdown, but we're at least technical users :)

PR #104, opened 2020-12-09
Swift: Use a marker protocol for store "context" parameters

jrose-signal:
...instead of a bare UnsafeMutableRawPointer.

The "most correct" thing to do would be to have the context be an associated type of each store, and then clients would set them to the correct type, and they'd all have to be compatible when using multiple stores in an operation. However, this falls down in practice because Swift doesn't allow values to have protocol type for protocols with associated type, and the iOS framework SignalMetadataKit (and plenty of tests) do in fact want to reference the store types without having to use concrete types. Additionally, preserving that type through the C function shims used in each with*Store function would require a lot of extra work, since Swift generics are not monomorphized (resolved down to concrete types at compile time).

Instead, use a marker protocol, which is easier to work with than a pointer and less error-prone than Any.

### DIFF
--- a/swift/Sources/SignalClient/DataStoreInMemory.swift
+++ b/swift/Sources/SignalClient/DataStoreInMemory.swift
@@ -3,6 +3,11 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+/// A dummy StoreContext usable with InMemorySignalProtocolStore.
+public struct NullContext: StoreContext {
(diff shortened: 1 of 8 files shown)


Imperiopolis-Signal (approved):

jack-signal (approved):

charlesmchen-signal (approved):
LGTM

PR #106, opened 2020-12-09
Node: Upgrade Neon to v0.6.0. neon-build is no longer required.

kjvalencik:
`neon-build` was specified, but was unused. This may have caused issues on macOS and Windows. However, as of Neon v0.6.0, `neon-build` for N-API only outputs the library as `index.node`.

Since the file is copied by an external build script, `neon-build` is not necessary.

Neon `v0.6.0` includes dynamic loading for N-API symbols which incurs a very small runtime costs, but has several benefits:
* Portable across Node and Electron versions (build once, use in both)
* Build without Internet
* Cross compiling to Windows


kjvalencik:
@jrose-signal The test is failing because of a N-API symbol that isn't present in older versions of Node/Electron. Are the versions being tested the minimum versions supported?

I might need to add feature flags into Neon for controlling the N-API version if there is demand for supporting N-API versions lower than 6.

jrose-signal:
Right now yes, we're following the version used in https://github.com/signalapp/Signal-Desktop. It is absolutely an old version of Electron (running a 12.13.x Node, apparently, which gives it N-API 5), but I don't know our timeline for moving forward.

One way around this without feature flags would be if dynamically loaded symbols were stored as Options. Then you'd get the panic when a client tried to *use* a feature that Neon didn't provide, rather than when failing to bind them.

(I don't actually think the dynamic binding change is a good idea on Linux and Mac; it's absolutely worse for load time and possibly dirty memory use as well. But I see that it's practical and makes configuration a lot easier.)

kjvalencik:
@jrose-signal I had considered that, but I thought that was worse than failing on load. The concern is that the API might only be used in some untested corner case of an app; failing at an unexpected time at runtime.

I'm going to work on implement feature flags for selecting the desired N-API version and disabling Neon features that require it. Currently, Neon has exactly _one_ N-API function that isn't in Version 1.

We aren't loading a lot of symbols, so I don't think the loading time is significant (admittedly, I haven't measured). My hunch is that it will be significantly faster than parsing a typical JavaScript file that is required.

Can you explain what you mean by dirty memory use?

jrose-signal:
> @jrose-signal I had considered that, but I thought that was worse than failing on load. The concern is that the API might only be used in some untested corner case of an app; failing at an unexpected time at runtime.

Makes sense. It does match the previous behavior though.

> I'm going to work on implement feature flags for selecting the desired N-API version and disabling Neon features that require it. Currently, Neon has exactly _one_ N-API function that isn't in Version 1.

Aw, if only it hadn't been exposed in the first place! :-)

> We aren't loading a lot of symbols, so I don't think the loading time is significant (admittedly, I haven't measured). My hunch is that it will be significantly faster than parsing a typical JavaScript file that is required.
> 
> Can you explain what you mean by dirty memory use?

Heh. Both of these were knee-jerk thoughts from my past work on Swift, where concerns at that level were significant by virtue of being magnified across many users and use cases. The dirty memory thing was mostly a hunch, on the grounds that the OS knows that a page of resolved symbols is read-only once resolved but can't say the same about an arbitrary global struct of pointers. But that's at most a page of overhead (if there was spare space on the rest of the page with resolvable symbols); it's not like the resolved symbols can be shared across processes or lazily pulled in from disk like true constant data. And we're talking about JavaScript, which for all the amazing GC improvements over the past years is still not the environment you'd pick if you were memory-constrained.

I do still think slowing down load time is Bad but you're probably right that adding more JS dependencies could slow it down just as much. So, comments withdrawn.

kjvalencik:
@jrose-signal I appreciate the feedback! I'll documented some of these thoughts and file them in an issue. We could use some macro magic to choose dynamic loading conditionally at build time.

There are _very_ [few users](https://github.com/search?l=TOML&q=%22napi-runtime%22&type=Code) of the Neon n-api runtime and I would love to know your experience, needs and suggestions.

jrose-signal:
To be fair, we are not users yet! The other library where we're using Neon is https://github.com/signalapp/ringrtc, and there we're still using it in v8 mode. Nothing's been released using N-API yet.

jrose-signal:
Superseded by #151.

PR #108, opened 2020-12-10
Swift: Propagate callback errors directly

jrose-signal:
This turned out to be easier to do in Swift than in Java, since we *already* have a Swift-side adapter that catches callback-thrown errors. I have the solution for Java too (#5), but that one's a Rust-side change on top of #76, so I'll put it up later.


jrose-signal:
\*sigh\* This is what I get for trying to decompose a larger change. I'll come back to this later.

PR #109, opened 2020-12-10
Make most of Rust's SessionState pub(crate)

jack-signal:
Remove some functions which were not used within the crate or the bindings.

Also fix some type errors in the JNI binding (introduced in #107) - SessionState was being used instead of SessionRecord, and this happened to work because the first element of a SessionRecord is an `Option<SessionState>`

End goal is removing `SessionState` from the crate exports entirely but we need to do a new release and update Android first.

### DIFF
--- a/rust/protocol/tests/ratchet.rs
+++ b/rust/protocol/tests/ratchet.rs
@@ -48,29 +48,29 @@ fn test_ratcheting_session_as_bob() -> Result<(), SignalProtocolError> {
 
     let alice_base_public_key = PublicKey::deserialize(&alice_base_public)?;
 
-    let alice_identity_public = IdentityKey::decode(&alice_identity_public)?;
-
     let bob_parameters = BobSignalProtocolParameters::new(
         bob_identity_key_pair,
         bob_signed_prekey_pair,
         None, // one time pre key pair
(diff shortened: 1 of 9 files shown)


rust/protocol/tests/ratchet.rs
```diff
@@ -48,29 +48,29 @@ fn test_ratcheting_session_as_bob() -> Result<(), SignalProtocolError> {
...
     );
 
     let bob_session = initialize_bob_session(&bob_parameters)?;
 
+    let bob_record = SessionRecord::new(bob_session);
+
     assert_eq!(
-        bob_session.local_identity_key()?,
-        *bob_identity_key_pair.identity_key()
+        hex::encode(bob_record.local_identity_key_bytes()?),
```
jrose-signal, line 63: I assume the `hex::encode` is to make for prettier output when the assertion fails?

jrose-signal (approved):

rust/protocol/tests/ratchet.rs
jack-signal: Exactly, not that it should ever fail, but trying to eyeball compare hex is a little easier than the array notation.

PR #114, opened 2020-12-10
Propagate callback exceptions/errors when they fail the whole operation

### LINKED ISSUE #5: jni: capture the Throwable
We currently only capture the name and message if a storage callback throws an exception. Instead capture the entire `Throwable` and provide it as a the `cause` field on the relevant exception we throw on the other side.

jrose-signal:
Resolves #5, as well as providing the equivalent feature for Swift.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -848,6 +848,7 @@ dependencies = [
  "jni",
  "libc",
  "libsignal-protocol-rust",
+ "log",
  "paste",
 ]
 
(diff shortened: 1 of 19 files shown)


jrose-signal:
Fixed Rust tests in 203e5d5, which can't use `==` for errors anymore.

jack-signal (approved):
lgtm and sorry about the wait

PR #115, opened 2020-12-10
Support generic bridge functions for FFI/JNI

jrose-signal:
This is *one* approach to generic bridge functions that have the same body in FFI and JNI, with the only difference being input and output types. Here's what it looks like:

```rust
#[bridge_fn(ffi = "publickey_compare")]
fn ECPublicKey_Compare(key1: &PublicKey, key2: &PublicKey) -> i32 {
    match key1.cmp(&key2) {
        std::cmp::Ordering::Less => -1,
        std::cmp::Ordering::Equal => 0,
        std::cmp::Ordering::Greater => 1,
    }
}

#[bridge_fn(ffi = "publickey_verify")]
fn ECPublicKey_Verify(
    key: &PublicKey,
    message: &[u8],
    signature: &[u8],
) -> Result<bool, SignalProtocolError> {
    key.verify_signature(&message, &signature)
}
```

This has a few moving pieces:
- `bridge_fn` is a proc-macro that expands into an FFI entry point and a JNI entry point, each of which call the original function. It classifies arguments as "by value", "by ref", or "array" (pointer + length in FFI).
- Individual input and output types are mapped by "rules" macros in the `ffi` and `jni` submodules, so that cbindgen (which only does a syntactic walk) can output the right types.
- There are specific traits (ArgTypeInfo, RefArgTypeInfo, SizedArgTypeInfo, ResultTypeInfo) for arbitrary conversion logic (what was formerly handled by things like `jint_to_u32`).

Thus, adding a new function is easy; adding a new type is not too bad (about the same as the ad hoc conversion functions from before); adding a new *kind* of type is tricky but won't happen often.

### DIFF
--- a/rust/bridge/shared/src/jni/convert.rs
+++ b/rust/bridge/shared/src/jni/convert.rs
@@ -4,7 +4,7 @@
 //
 
 use jni::JNIEnv;
-use jni::objects::JString;
+use jni::objects::{AutoByteArray, JString, ReleaseMode};
 use jni::sys::{JNI_FALSE, JNI_TRUE};
 use libsignal_protocol_rust::*;
 use std::borrow::Borrow;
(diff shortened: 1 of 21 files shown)


rust/bridge/shared/src/support.rs
```diff
@@ -19,6 +33,15 @@ macro_rules! expr_as_fn {
     };
 }
 
+// macro_rules! bridge_fn {
```
jack-signal, line 36: Dead code?

rust/bridge/shared/src/jni/convert.rs
```diff
@@ -0,0 +1,196 @@
...
+        impl ResultTypeInfo for $typ {
+            type ResultType = ObjectHandle;
+            fn convert_into(self, _env: &JNIEnv) -> Result<Self::ResultType, SignalJniError> {
+                box_object(Ok(self))
+            }
+        }
+    };
+}
+
+native_handle!(PublicKey);
```
jack-signal, line 135: shared/src/lib.rs

jack-signal (approved):

PR #116, opened 2020-12-10
Rust: Move deny-warnings behavior into CI

jrose-signal:
Warnings don't need to be immediately fixed locally, and they make it harder to test with newer nightlies. CI is the place to enforce it.


jack-signal (approved):
Yes

PR #117, opened 2020-12-11
AES: Fix duplicated aarch64-apple-ios helper method

jrose-signal:


jrose-signal:
I'm not sure of the best way to test this. We could add a cross-compile step for the Rust code, but cross-compiling the Swift code in a realistic way means stepping beyond the default `pod lib lint` provided by CocoaPods. Maybe I should make a manual test case after all (cf. #59).

jack-signal (approved):
Hrm. I had assumed the Cocoapods lint step was already cross compiling for iOS. At the very least we should cross compile the Rust code. At some point I tried setting that up from Linux and with https://github.com/tpoechtrager/osxcross it seems to be possible but kind of a pain. I would imagine we can cross compile for iOS from Mac quite easily though.

jrose-signal:
> Hrm. I had assumed the Cocoapods lint step was already cross compiling for iOS.

I had hoped so too, but it looks like it's just compiling for iOS sim, not device. :-/

> I would imagine we can cross compile for iOS from Mac quite easily though.

Yep, if we don't care about the Swift wrapper (which, admittedly, shouldn't have anything arch-specific), we can just do `cargo build --target aarch64-apple-ios`.

PR #118, opened 2020-12-11
Add the logic for handling revoked server certificates

jack-signal:
Reserve one ID for testing

ServerCertificates are created by
https://github.com/signalapp/Signal-Server-Private/blob/master/service/src/main/java/org/whispersystems/textsecuregcm/workers/CertificateCommand.java I'm going to open a PR reserving this value there so we never have a tragic collision between test and prod values


jrose-signal (approved):

jrose-signal:
Looks like this broke the tests since I changed the error enums way back.

PR #119, opened 2020-12-14
Cross build the Rust code for iOS

jack-signal:
Hopefully avoiding bugs like #117

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -19,6 +19,7 @@ jobs:
       java: ${{ steps.filter.outputs.java }}
       node: ${{ steps.filter.outputs.node }}
       swift: ${{ steps.filter.outputs.swift }}
+      rust_ios: ${{ steps.filter.outputs.rust_ios }}
 
     steps:
     - uses: actions/checkout@v2
@@ -57,9 +58,11 @@ jobs:
           - 'package.json'
           - 'rust/bridge/node/**'
           - 'yarn.lock'
-          swift:
+          rust_ios: &rust_ios
           - *all
           - 'rust/bridge/ffi/**'
+          swift:
+          - *rust_ios
           - 'swift/**'
           - 'SignalClient.podspec'
           ignored:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -107,6 +107,30 @@ jobs:
...
       run: cargo clippy --all -- -D warnings
 
+  rust_ios:
+    name: Build Rust for iOS
+
+    runs-on: macos-latest
+
+    needs: changes
+
+    if: ${{ needs.changes.outputs.rust == 'true' }}
```
jrose-signal, line 117: This can be `needs.changes.outputs.swift` to avoid changes that aren't part of libsignal-ffi.

.github/workflows/build_and_test.yml
```diff
@@ -107,6 +107,30 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        components: rustfmt, clippy
```
jrose-signal, line 126: Nitpicks: this custom action also has `target: aarch64-apple-ios` as a config option (if you only need one target).

(Also, we're not using Clippy in this build.)

.github/workflows/build_and_test.yml
jack-signal: OTOH with that, we rebuild even on changes that only touch Swift code. I added a new filter covering just the Rust code that is relevant to iOS

.github/workflows/build_and_test.yml
```diff
@@ -57,6 +57,9 @@ jobs:
           - 'package.json'
           - 'rust/bridge/node/**'
           - 'yarn.lock'
+          ios:
+          - *all
+          - 'rust/bridge/ffi/**'
           swift:
           - *all
           - 'rust/bridge/ffi/**'
```
jrose-signal, line 66: If you add this, let's make it a reference inside `swift`, so there's no duplication.

jrose-signal (approved):

PR #121, opened 2020-12-14
Use the correct handle

### LINKED ISSUE #120: SenderKeyRecord.serialize() error
When trying to serialize SenderKeyRecord i'm getting the error "nullParameter("null pointer")".
Tried changing "handle" with "nativeHandle" inside serialize method and it worked but i'm not sure if that is the solution.

jack-signal:
Fixes #120

I am not happy that this is not caught by the compiler. A grep seems to indicate this was the only such mistake.

Possibly ClonableOwnerHandle.handle should be `fileprivate`?

### DIFF
--- a/swift/Sources/SignalClient/state/SenderKeyRecord.swift
+++ b/swift/Sources/SignalClient/state/SenderKeyRecord.swift
@@ -15,8 +15,6 @@ public class SenderKeyRecord: ClonableHandleOwner {
         return signal_sender_key_record_clone(&newHandle, currentHandle)
     }
 
-    private var handle: OpaquePointer?
-
     public init<Bytes: ContiguousBytes>(bytes: Bytes) throws {
         let handle: OpaquePointer? = try bytes.withUnsafeBytes {
             var result: OpaquePointer?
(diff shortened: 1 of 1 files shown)


jrose-signal:
It's already fully `private`. You accidentally shadowed it in the subclass (scroll up), which is all right because a subclass shouldn't know about its parent's implementation details (okay), and because neither Swift nor Swiftlint has a warning for "variable was never assigned to" (bad). Or are you saying `handle` should be `internal` so that we'll get a complaint for shadowing it?

jack-signal:
Ah I had missed the extra handle. OK at least that makes more sense - I didn't understand why Swift would willingly convert an `enum` to `OpaquePointer?`. Seems this happened in 934cb60a5cae593 the original member just never got deleted when `ClonableOwnerHandle` was introduced.

jrose-signal (approved):

jack-signal:
I suggested `fileprivate` because  (based on my assumption that we were referencing `ClonableHandleOwner`'s private `handle` member within `SenderKeyRecord`)  I guessed that Swift's `private` was more akin to C++ `protected`, and we needed `fileprivate` to prevent access from subclasses outside of the file, but on further reading I see this is not the case at all.

I'm not sure if `internal` helps or not here. The most likely way for this kind of bug to happen is (as here) where some code using `handle` is converted to using `ClonableHandleOwner`), and someone forgets to update all of the functions to use the correct type *and* leaves the member variable declared. But more likely we'd just use `ClonableHandleOwner` out of the gate for types where it makes sense, now that it is available? And there doesn't seem to be any legit reason for anything to ever touch `handle` itself.

jrose-signal:
Yeah, that sounds right to me as well. It does feel a little funny to me that ClonableHandleOwner has been used for non-clonable handles, but I suppose there's not much of a downside.

PR #122, opened 2020-12-15
FFI: drop build.rs that invokes cbindgen

jrose-signal:
Using cbindgen in a Cargo build script is convenient, but it means cbindgen can't reuse the results of the actual build for macro expansion, which results in doing a good chunk of a *second* build. Move that logic to swift/build_ffi.sh instead.

Blocked on a cbindgen release that contains support for eqrion/cbindgen#603. (Or we ask people to `cargo install` a specific hash, which seems sketchy. Building it on demand takes too long, though—longer than libsignal-client itself.)


jrose-signal:
…really need to figure out a good way to cache the built `cbindgen` (since the version provided by GitHub Actions isn't very new).

jack-signal (approved):

.github/workflows/build_and_test.yml
```diff
@@ -258,6 +263,12 @@ jobs:
           profile: minimal
           target: x86_64-apple-ios
 
+      - name: Install cbindgen
+        # FIXME: Wait until there's a released version of cbindgen!
+        run: cargo install cbindgen --git https://github.com/eqrion/cbindgen
+        env:
+          CARGO_TARGET_DIR: cargo-install-cache
```
jrose-signal, line 270: Did not work the way I wanted it to: https://github.com/signalapp/libsignal-client/pull/122/checks?check_run_id=1572842893

jrose-signal:
Discovered complication: because cbindgen expands libsignal-bridge, it's invalidated as a dependency for libsignal-ffi, which means we never reach a clean build state.

PR #124, opened 2020-12-17
Add PublicKey and PrivateKey for Node

jack-signal:
~Mostly good to go except I need to work out the error handling strategy, right now there is a lot of `unwrap`~

### DIFF
--- a/node/index.ts
+++ b/node/index.ts
@@ -8,14 +8,67 @@ import * as SignalClient from './libsignal_client';
 
 const SC = bindings('libsignal_client') as typeof SignalClient;
 
+export class PublicKey {
+  private readonly nativeHandle: SignalClient.PublicKey;
+
+  private constructor(handle: SignalClient.PublicKey) {
+    this.nativeHandle = handle;
+  }
+
+  static fromNativeHandle(handle: SignalClient.PublicKey): PublicKey {
+    return new PublicKey(handle);
+  }
+
+  static deserialize(buf: Buffer): PublicKey {
+    return new PublicKey(SC.PublicKey_deserialize(buf));
+  }
+
+  serialize(): Buffer {
+    return SC.PublicKey_serialize(this.nativeHandle);
+  }
+
+  verify(msg: Buffer, sig: Buffer): boolean {
+    return SC.PublicKey_verify(this.nativeHandle, msg, sig);
+  }
+
+  _unsafeGetNativeHandle(): SignalClient.PublicKey {
+    return this.nativeHandle;
+  }
+}
+
 export class PrivateKey {
   private readonly nativeHandle: SignalClient.PrivateKey;
 
-  constructor() {
-    this.nativeHandle = SC.PrivateKey_generate();
+  private constructor(handle: SignalClient.PrivateKey) {
+    this.nativeHandle = handle;
+  }
+
+  static generate(): PrivateKey {
+    return new PrivateKey(SC.PrivateKey_generate());
+  }
+
+  static deserialize(buf: Buffer): PrivateKey {
+    return new PrivateKey(SC.PrivateKey_deserialize(buf));
   }
 
   serialize(): Buffer {
     return SC.PrivateKey_serialize(this.nativeHandle);
+  }
+
+  sign(msg: Buffer): Buffer {
+    return SC.PrivateKey_sign(this.nativeHandle, msg);
+  }
+
+  agree(other_key: PublicKey): Buffer {
+    return SC.PrivateKey_agree(
+      this.nativeHandle,
+      other_key._unsafeGetNativeHandle()
+    );
+  }
+
+  getPublicKey(): PublicKey {
+    return PublicKey.fromNativeHandle(
+      SC.PrivateKey_getPublicKey(this.nativeHandle)
+    );
   }
 }
--- a/node/libsignal_client.d.ts
+++ b/node/libsignal_client.d.ts
@@ -8,9 +8,25 @@
 // FIXME: Eventually we should be able to autogenerate this.
 
 // Newtype pattern from https://kubyshkin.name/posts/newtype-in-typescript/
+interface PublicKey {
+  readonly __type: unique symbol;
+}
+
+export function PublicKey_deserialize(buf: Buffer): PublicKey;
+export function PublicKey_serialize(key: PublicKey): Buffer;
+export function PublicKey_verify(
+  key: PublicKey,
+  msg: Buffer,
+  signature: Buffer
+): boolean;
+
 interface PrivateKey {
   readonly __type: unique symbol;
 }
 
 export function PrivateKey_generate(): PrivateKey;
+export function PrivateKey_deserialize(buf: Buffer): PrivateKey;
 export function PrivateKey_serialize(key: PrivateKey): Buffer;
+export function PrivateKey_sign(key: PrivateKey, msg: Buffer): Buffer;
+export function PrivateKey_agree(key: PrivateKey, other_key: PublicKey): Buffer;
+export function PrivateKey_getPublicKey(key: PrivateKey): PublicKey;
--- a/node/test/PublicAPITest.ts
+++ b/node/test/PublicAPITest.ts
@@ -7,14 +7,71 @@ import { assert } from 'chai';
 import * as SignalClient from '../index';
 
 describe('SignalClient', () => {
-  it('can generate and serialize PrivateKeys', () => {
-    const a = new SignalClient.PrivateKey();
-    const b = new SignalClient.PrivateKey();
-    assert.equal(a.serialize().length, 32, 'correct length');
-    assert(a.serialize().equals(a.serialize()), 'repeatable');
-    assert(
-      !a.serialize().equals(b.serialize()),
+  it('ECC signatures work', () => {
+    const priv_a = SignalClient.PrivateKey.generate();
+    const priv_b = SignalClient.PrivateKey.generate();
+    assert.lengthOf(priv_a.serialize(), 32, 'private key serialization length');
+    assert.deepEqual(priv_a.serialize(), priv_a.serialize(), 'repeatable');
+    assert.notDeepEqual(
+      priv_a.serialize(),
+      priv_b.serialize(),
       'different for different keys'
     );
+
+    const pub_a = priv_a.getPublicKey();
+    const pub_b = priv_b.getPublicKey();
+
+    const msg = Buffer.from([1, 2, 3]);
+
+    const sig_a = priv_a.sign(msg);
+    assert.lengthOf(sig_a, 64, 'signature length');
+
+    assert(pub_a.verify(msg, sig_a));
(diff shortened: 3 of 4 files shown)


jack-signal:
Important design question here:

To get access to values eg in a `JsBuffer`, we have to use `lock` or `borrow` both of which lock the interpreter. We then while holding the lock perform operations which can be relatively slow eg signature generation or verification, protobuf parsing, etc. Would it be better is to just always copy the inputs into the Rust heap, then perform the core operations without holding the lock?

node/index.ts
```diff
@@ -8,14 +8,64 @@ import * as SignalClient from './libsignal_client';
...
 const SC = bindings('libsignal_client') as typeof SignalClient;
 
+export class PublicKey {
+  private readonly nativeHandle: SignalClient.PublicKey;
+
+  private constructor(handle: SignalClient.PublicKey) {
+    this.nativeHandle = handle;
+  }
+
+  static from_native_handle(handle: SignalClient.PublicKey): PublicKey {
```
jrose-signal, line 18: Nitpick: JavaScript camelCase, please. (Surprised this didn't show up in `yarn lint` or `yarn format`.)

node/index.ts
```diff
@@ -8,14 +8,64 @@ import * as SignalClient from './libsignal_client';
...
+
+  sign(msg: Buffer): Buffer {
+    return SC.PrivateKey_sign(this.nativeHandle, msg);
+  }
+
+  agree(other_key: PublicKey): Buffer {
+    return SC.PrivateKey_agree(this.nativeHandle, other_key.getNativeHandle());
+  }
+
+  public_key(): PublicKey {
```
jrose-signal, line 66: Is this more idiomatic as a no-argument method or a get-only property? (https://www.typescriptlang.org/docs/handbook/classes.html#accessors) Would probably want one of the Desktop folks to decide.

rust/bridge/node/src/lib.rs
```diff
@@ -22,32 +22,146 @@ impl<T> Deref for DefaultFinalize<T> {
...
+
+fn return_binary_data<'a, T: AsRef<[u8]>>(
+    cx: &mut FunctionContext<'a>,
+    bytes: Result<T, SignalProtocolError>,
+) -> JsResult<'a, JsValue> {
+    match bytes {
+        Ok(bytes) => {
+            let bytes = bytes.as_ref();
+            let bytes_len = bytes.len() as u32;
+            if bytes_len as usize != bytes.len() {
```
jrose-signal, line 53: Nitpick: how about `u32::try_from`?

rust/bridge/node/src/lib.rs
```diff
@@ -22,32 +22,146 @@ impl<T> Deref for DefaultFinalize<T> {
...
 }
 
 #[neon::main]
 fn main(mut cx: ModuleContext) -> NeonResult<()> {
     cx.export_function("PrivateKey_generate", PrivateKey_generate)?;
+    cx.export_function("PrivateKey_deserialize", PrivateKey_deserialize)?;
     cx.export_function("PrivateKey_serialize", PrivateKey_serialize)?;
+    cx.export_function("PrivateKey_sign", PrivateKey_sign)?;
+    cx.export_function("PrivateKey_agree", PrivateKey_agree)?;
+    cx.export_function("PrivateKey_get_publickey", PrivateKey_get_publickey)?;
```
jrose-signal, line 161: I was thinking these should probably be JavaScript convention, `PrivateKey_getPublicKey`. Among other choices, it'll be easier to unify with our JNI approach.

jrose-signal:
> To get access to values in a `JsBuffer`, we have to use `lock` or `borrow` both of which lock the interpreter. We then while holding the lock perform operations which can be relatively slow (say .1 ms or so) like signature generation or verification. Would it be better to just always copy the inputs into the Rust heap, then perform the core operations without holding the lock?

This is JavaScript we're talking about here, so I don't think it matters: everything's executed on a single thread, even `async` functions (they're put into a run-loop-style task queue). If we do any of our own concurrency, though, it might make sense.

jrose-signal (commented):
Looks pretty good! Only a few comments from my side.

binding.gyp
```diff
@@ -14,7 +14,6 @@
                     'action': [
                         'env',
                         'CONFIGURATION_NAME=<(CONFIGURATION_NAME)',
-                        'CARGO_BUILD_TARGET_DIR=<(INTERMEDIATE_DIR)/rust',
```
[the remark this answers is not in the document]
jack-signal: At least for me I was seeing `yarn` wipe out this dir between rebuilds which forced a complete rebuild of the crate and all dependencies, which was rather inconvenient. I'm ok with reverting this and just keeping it locally for development.

node/libsignal_client.d.ts
```diff
@@ -8,9 +8,25 @@
 // FIXME: Eventually we should be able to autogenerate this.
```
EvanHahn-Signal, line 8: A broader discussion that we shouldn't have in a driveby PR comment:

TypeScript doesn't have a sound type system (succinctly described in [this blog post][0] if you're curious). That means that declaration files could contain lies, even if they're autogenerated. The soundest thing you can do is to declare most things as `unknown` and then use runtime type assertions. Onerous, but sound. (This [bit us at least once in RingRTC][1], though autogeneration would've fixed that.)

I think autogeneration is probably a reasonable middle ground, but I'd love to discuss our goals here. How important are sound types? Do we want to add runtime type assertions to make sure we're in the clear?

Again, probably not worth discussing here, but something I wanted to flag as it extends beyond this PR.

[0]: https://blog.logrocket.com/is-typescript-worth-it/
[1]: https://github.com/signalapp/ringrtc/commit/8c7d9ae0cba6dc596d86de2835c90bae312564e6#diff-d4fb1204a18470143330eacf6a818ddc5a550610e8e57409c102d44d253b03eaR1519

node/index.ts
EvanHahn-Signal: I'd call this `getPublicKey` and keep it as a function. I rarely see accessors used in JavaScript-land. (This is anecdotal, and other Desktop devs might have different experiences.)

(Some pedantry that I hope is educational and not nitpicky: accessors are a JavaScript feature, not a TypeScript feature.)

node/test/PublicAPITest.ts
```diff
@@ -7,14 +7,76 @@ import { assert } from 'chai';
...
-  it('can generate and serialize PrivateKeys', () => {
-    const a = new SignalClient.PrivateKey();
-    const b = new SignalClient.PrivateKey();
-    assert.equal(a.serialize().length, 32, 'correct length');
-    assert(a.serialize().equals(a.serialize()), 'repeatable');
+  it('ECC signatures work', () => {
+    const priv_a = SignalClient.PrivateKey.generate();
+    const priv_b = SignalClient.PrivateKey.generate();
+    assert.equal(priv_a.serialize().length, 32, 'correct length');
+    assert(priv_a.serialize().equals(priv_a.serialize()), 'repeatable');
```
EvanHahn-Signal, line 14: If you wish:

```suggestion
    assert.deepEqual(priv_a.serialize(), priv_a.serialize(), 'repeatable');
```

You can do this any time you're comparing buffers. It can give a clearer error message, showing you the diff instead of "assertion failed".

You can use this elsewhere in this file, too.

node/test/PublicAPITest.ts
```diff
@@ -7,14 +7,76 @@ import { assert } from 'chai';
...
+
+  it('decoding invalid ECC key throws an error', () => {
+    const invalid_key = Buffer.alloc(33, 0xab);
+
+    try {
+      SignalClient.PrivateKey.deserialize(invalid_key);
+      assert(false, 'deserialized invalid key');
+    } catch (err) {
+      assert.equal(err.message, 'bad key length <33> for key with type <Djb>');
+    }
```
EvanHahn-Signal, line 73: I think this is clearer, shorter, and more idiomatic:

```suggestion
    assert.throws(() => {
      SignalClient.PrivateKey.deserialize(invalid_key);
    }, 'bad key length <33> for key with type <Djb>');
```

Also applies to the assertions below.

node/index.ts
```diff
@@ -8,14 +8,64 @@ import * as SignalClient from './libsignal_client';
...
+
+  serialize(): Buffer {
+    return SC.PublicKey_serialize(this.nativeHandle);
+  }
+
+  verify(msg: Buffer, sig: Buffer): boolean {
+    return SC.PublicKey_verify(this.nativeHandle, msg, sig);
+  }
+
+  getNativeHandle(): SignalClient.PublicKey {
```
EvanHahn-Signal, line 34: Should we add tests for this?

EvanHahn-Signal (approved):
LGTM other than my comments. Probably good to get other peoples' eyes on this, though.

node/libsignal_client.d.ts
jrose-signal: The types in this file are not real JavaScript types; they're all going to be Node "[externals](https://nodejs.org/api/n-api.html#n_api_napi_create_external)". So lies are exactly what we want (to use TypeScript's type system to protect us). The autogeneration comment is mostly about avoiding writing all these functions out when it's already in the Rust files.

node/index.ts
jack-signal: @EvanHahn-Signal yes please any help on the JS tooling/linting side is most welcome

EvanHahn-Signal: I'll make a note to follow up with y'all on this in 2021.

node/test/PublicAPITest.ts
jack-signal: Nice thanks!

[You are looking at literally the first JS/TS I have ever written so any suggestions on style, easier ways of doing things etc are super useful]

node/index.ts
jack-signal: There isn't anything that JavaScript world can meaningfully do with the result. The only reason it is not private is that operations that need to take in a JS PublicKey and pass it to a Rust operation needs to be able to get at the underlying handle (see for example in PrivateKey.agree). Ideally this would be package private; nothing outside of this file should ever call it. But I couldn't find any way of doing that.

Maybe, given that this will be accessible to anyone using the library, it should be given a scarier name like `_unsafeGetNativeHandle`.

binding.gyp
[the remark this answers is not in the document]
jack-signal: Removed this from git

node/test/PublicAPITest.ts
EvanHahn-Signal: This is a _lot_ better than the first JS I wrote...and not far off the JS I write today!

node/index.ts
EvanHahn-Signal: I'd prefer that name—makes it clearer that it's only public because of a weakness of JavaScript.

jack-signal: Done

node/test/PublicAPITest.ts
jack-signal: Changed thanks

PR #126, opened 2020-12-21
Node: list files for `yarn pack`; add an alias `yarn build` for faster Rust turnaround

jrose-signal:
My hope is that `yarn pack` (on each build platform) will do the copying we need for the artifact repo, making it a drop-in change to use the source repo instead, but for now it's just there. `yarn build` addresses Jack's complaint in #124 that `yarn install` always calls `yarn node-gyp rebuild`, which does a clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "build/Release/*.node"
   ],
   "scripts": {
+    "build": "node-gyp build",
     "tsc": "tsc -p node && cp node/*.d.ts node/dist",
     "clean": "rimraf node/dist build",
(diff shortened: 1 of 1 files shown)


scottnonnenberg-signal:
Can't see anything wrong with this, but I'd need to test the resultant assets before being totally sure.

EvanHahn-Signal:
Removing myself as a reviewer because I think Scott's got this one, but let me know if that's wrong.

jrose-signal:
I'm going to merge, then. We can always tweak it if/when `yarn pack` becomes something we actually depend on, rather than just something I'm experimenting with.

PR #127, opened 2020-12-22
Implement IdentityKeyPair Deserialize

AsamK:
This PR adds jni glue code for the `IdentityKeyPair` `deserialize` method and makes the `serialize` method public, as it was previously in the libsignal-protocol-java library.
The `IdentityKeyPair` consists of a private and public key, so the deserialize method needs to return a handle for each key. I've implemented the tuple as a simple java array with two elements. It's not very type safe but seemed the simplest solution here.

Is this something you'd consider merging?
I use these two methods in my signal-cli project.

### DIFF
--- a/rust/bridge/jni/src/lib.rs
+++ b/rust/bridge/jni/src/lib.rs
@@ -7,7 +7,7 @@
 
 use async_trait::async_trait;
 use jni::objects::{JClass, JObject, JString, JValue};
-use jni::sys::{jboolean, jbyteArray, jint, jlong, jobject, jstring};
+use jni::sys::{jboolean, jbyteArray, jint, jlong, jlongArray, jobject, jstring};
 use jni::JNIEnv;
 use std::convert::TryFrom;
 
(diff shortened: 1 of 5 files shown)


jrose-signal:
The best person to review this would be @jack-signal, but he's already on winter vacation. I'll make sure we get to this when he gets back in January!

rust/bridge/jni/src/lib.rs
```diff
@@ -156,6 +156,28 @@ pub unsafe extern "C" fn Java_org_signal_client_internal_Native_IdentityKeyPair_
...
 
+#[no_mangle]
+pub unsafe extern "C" fn Java_org_signal_client_internal_Native_IdentityKeyPair_1Deserialize(
+    env: JNIEnv,
+    _class: JClass,
+    data: jbyteArray,
+    offset: jint,
+) -> jlongArray {
+    run_ffi_safe(&env, || {
+        let offset = jint_to_u32(offset)? as usize;
```
jack-signal, line 167: offset doesn't seem to be required here since it is always zero.

We support an offset for public key deserialization because Android client needed/wanted that functionality in one particular spot.

jack-signal (changes requested):
Change is fine except that `offset` argument should be removed.

I'd caution that generally we are implementing this library for Signal's use case only. But here the change is simple and restores interfaces which were removed from libsignal-protocol-java, so seems ok to me.

rust/bridge/jni/src/lib.rs
AsamK: I've removed the offset.

AsamK:
Thanks for reviewing! I've made the requested change.

jack-signal (approved):

PR #131, opened 2021-01-06
protocol: Remove spurious session state archiving processing a prekey

jrose-signal:
`promote_state`, called below, is already going to archive the previous session state.

### DIFF
--- a/rust/protocol/src/session.rs
+++ b/rust/protocol/src/session.rs
@@ -109,8 +109,6 @@ async fn process_prekey_v3(
         *message.base_key(),
     );
 
-    session_record.archive_current_state()?;
-
     let mut new_session = ratchet::initialize_bob_session(&parameters)?;
 
     new_session.set_local_registration_id(identity_store.get_local_registration_id(ctx).await?)?;


rust/protocol/src/session.rs
```diff
@@ -109,8 +109,6 @@ async fn process_prekey_v3(
         *message.base_key(),
     );
 
-    session_record.archive_current_state()?;
-
     let mut new_session = ratchet::initialize_bob_session(&parameters)?;
 
     new_session.set_local_registration_id(identity_store.get_local_registration_id(ctx).await?)?;
```
jrose-signal, line 114: The one thing I'm not sure about is error exits. If we want to make sure we always archive if we get this far, we should change the use of `promote_state` below instead.

jack-signal: The current logic matches Java, but I don't think it matters much. Archiving early does mean we bump off the very oldest (40th) session state immediately rather than later. If an error occurs somewhere below, we end up with no active state in the session, which is fine since we'll check the old states anyway during processing, and then promote whatever we used to the active state.

jack-signal (approved):

PR #133, opened 2021-01-06
Swift: add SessionRecord.hasCurrentState

jrose-signal:
If a session has just been reset (`archiveCurrentState()`), then there won't be an active state in the session and sends will fail until a new session is properly established by the client.

### DIFF
--- a/rust/bridge/ffi/src/lib.rs
+++ b/rust/bridge/ffi/src/lib.rs
@@ -147,6 +147,21 @@ pub unsafe extern "C" fn signal_session_record_archive_current_state(
     })
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn signal_session_record_has_current_state(
+    result: *mut bool,
+    session_record: *const SessionRecord,
+) -> *mut SignalFfiError {
+    run_ffi_safe(|| {
+        if result.is_null() {
+            return Err(SignalFfiError::NullPointer);
+        }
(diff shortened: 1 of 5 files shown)


jrose-signal:
I don't love the name of the accessor here. "hasActiveState"? "canSend"? @Imperiopolis-Signal, @charlesmchen-signal, thoughts?

rust/bridge/ffi/src/lib.rs
```diff
@@ -260,6 +260,21 @@ pub unsafe extern "C" fn signal_session_record_archive_current_state(
...
+pub unsafe extern "C" fn signal_session_record_has_current_state(
+    result: *mut bool,
+    session_record: *const SessionRecord,
+) -> *mut SignalFfiError {
+    run_ffi_safe(|| {
+        if result.is_null() {
+            return Err(SignalFfiError::NullPointer);
+        }
+        let session_record = native_handle_cast::<SessionRecord>(session_record)?;
+        *result = session_record.current_session.is_some();
```
jack-signal, line 273: The `current_session` isn't accessible anymore, I think you'll need to add an equivalent accessor on `SessionRecord` (or you could use `session_record.session_state().is_ok()` knowing that the only reason that would fail is because there is no current session, but that's a little subtle)

jack-signal (approved):

Imperiopolis-Signal (approved):

PR #134, opened 2021-01-07
Resolve difference in behavior between Java and Rust SessionRecord

jack-signal:
In libsignal-protocol-java, SessionRecord holds a SesssionState struct which is the "active" session plus a list of old states. If the record is freshly created, there is still a SessionState, but it is an uninitialized/new protobuf structure which causes all fields to be empty/zero/false.

So in the original Java logic you can call for example hasSenderChain, and on an empty/fresh record it will return false. However in Rust, in this case the Option is empty and we return an error instead.

For hasSenderChain, it seems reasonable to return false if there is no active session, since if there is no session there is certainly no chain.

Android also expects the session version to be == 0 on such sessions, but this makes less sense, so have this logic only in the Java binding and not in the Rust library proper.

### DIFF
--- a/rust/bridge/jni/src/lib.rs
+++ b/rust/bridge/jni/src/lib.rs
@@ -1518,8 +1518,11 @@ jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetLocalR
 jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetRemoteRegistrationId(SessionRecord) using SessionRecord::remote_registration_id);
 
 // For historical reasons Android assumes this function will return zero if there is no session state
-jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetSessionVersion(SessionRecord) using
-                 |s: &SessionRecord| Ok(s.session_version().unwrap_or(0)));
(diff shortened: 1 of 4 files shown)


rust/bridge/jni/src/lib.rs
```diff
@@ -1516,7 +1516,10 @@ pub unsafe extern "C" fn Java_org_signal_client_internal_Native_SessionRecord_1F
 
 jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetLocalRegistrationId(SessionRecord) using SessionRecord::local_registration_id);
 jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetRemoteRegistrationId(SessionRecord) using SessionRecord::remote_registration_id);
-jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetSessionVersion(SessionRecord) using SessionRecord::session_version);
+
+// For historical reasons Android assumes this function will return zero if there is no session state
+jni_fn_get_jint!(Java_org_signal_client_internal_Native_SessionRecord_1GetSessionVersion(SessionRecord) using
+                 |s: &SessionRecord| Ok(s.session_version().unwrap_or(0)));
```
jrose-signal, line 1522: This will swallow *any* errors, including any we might add in the future. I'd rather be more conservative here.

jrose-signal (approved):

PR #135, opened 2021-01-07
Fix handling when attempting to decrypt with a session that isn't found

jack-signal:
There were two discrepancies between the logic here and the original logic of libsignal-protocol-java.

First, if the session record had an uninitialized active session, in Java this would still attempt decryption with the old session states, but Rust would stop immediately without trying the old states. [I am not sure if this ever happens but it could possibly occur due to use of archiveCurrentState]

Secondly, we returned the wrong error condition. We treated lack of a sender chain as an invalid state (effectively an internal error) but Java treats it as an invalid message, which makes sense in so far as it is a message which we are unable to process with the information we have available. This wrong error type led to an unexpected exception being thrown in Android.

### DIFF
--- a/java/tests/src/test/java/org/whispersystems/libsignal/SessionCipherTest.java
+++ b/java/tests/src/test/java/org/whispersystems/libsignal/SessionCipherTest.java
@@ -71,6 +71,37 @@ public class SessionCipherTest extends TestCase {
     }
   }
 
+  public void testDecryptAfterReset() throws Exception {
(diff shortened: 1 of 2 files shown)


jrose-signal:
It happens with reset sessions but I think it can also happen with delayed messages. `archiveCurrentState()` should make it possible to add a test case for this, though, which we should definitely have.

jrose-signal (approved):

greyson-signal (approved):

PR #136, opened 2021-01-11
Node: name native library as libsignal_client_PLATFORM.node

jrose-signal:
This will make it easier to have a combined artifact root for all supported platforms. Based on what RingRTC is doing, though without as much customization.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -4,9 +4,14 @@
 #
 
 {
+    'conditions': [
+        ['OS=="mac"', {'variables': {'NODE_OS_NAME': 'darwin'}},
+         'OS=="win"', {'variables': {'NODE_OS_NAME': 'win32'}},
+         {'variables': {'NODE_OS_NAME': '<(OS)'}}],
+    ],
     'targets': [
         {
-            'target_name': 'libsignal_client.node',
+            'target_name': 'libsignal_client_<(NODE_OS_NAME).node',
             'type': 'none',
             'actions': [
                 {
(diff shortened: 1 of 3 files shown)


jack-signal:
Nice, yes. I had currently just hacked this with a manual rename in libsignal-client-node asset repo

jack-signal (approved):

binding.gyp
```diff
@@ -4,9 +4,14 @@
 #
 
 {
+    'conditions': [
+        ['OS=="mac"', {'variables': {'NODE_OS_NAME': 'darwin'}},
+         'OS=="win"', {'variables': {'NODE_OS_NAME': 'win32'}},
+         {'variables': {'NODE_OS_NAME': '<(OS)'}}],
```
EvanHahn-Signal, line 10: This seems reasonable based on [the docs](https://gyp.gsrc.io/docs/InputFormatReference.md#predefined-variables), but I could see failing if `OS` is not `win`, `mac`, or `linux`. For example, if it's unset or `freebsd`, we probably want to show some error.

Feel free to ignore—not sure this is a good suggestion.

node/build_node_bridge.sh
```diff
@@ -18,6 +18,8 @@ Usage: $(basename "$0") [-d] [-o DIR/]
 Options:
 	-d -- debug build (default is release, follows \$CONFIGURATION_NAME)
 	-o -- where to copy the built module (default: build/\$CONFIGURATION_NAME)
+
+NODE_OS_NAME must also be set to the value of Node's \`os.platform()\`.
```
EvanHahn-Signal, line 22: Could be worth mentioning that this is an environment variable for a newbie, but maybe not worth it.

EvanHahn-Signal (approved):
Looks reasonable to me! I asked a couple of small questions but I don't think either should block merge.

binding.gyp
jrose-signal: The last line defaults to passing through gyp's OS as the NODE_OS_NAME. That happens to be [correct](https://nodejs.org/api/os.html#os_os_platform) for `freebsd` and `openbsd`, and we can handle more unusual platforms as they come.

node/build_node_bridge.sh
jrose-signal: I think in practice it'll be rare to invoke this explicitly at all, but maybe *that* note should be there too.

binding.gyp
jrose-signal: Ah, the gyp docs are really bad, but their test cases include the BSDs, and the setting *is* populated automatically when run by node-gyp.

P.S. I cringe every time I write this tool's name…

PR #139, opened 2021-01-11
Pin cargo-ndk tool to 1.0.0

jack-signal:
2.0.0 is incompatible and behaves differently, would need work to adopt.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -152,7 +152,7 @@ jobs:
       run: rustup target add aarch64-linux-android armv7-linux-androideabi x86_64-linux-android i686-linux-android
 
     - name: Install NDK tool
-      run: cargo install cargo-ndk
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -152,7 +152,7 @@ jobs:
       run: rustup target add aarch64-linux-android armv7-linux-androideabi x86_64-linux-android i686-linux-android
 
     - name: Install NDK tool
-      run: cargo install cargo-ndk
+      run: cargo install cargo-ndk=1.0.0
```
jrose-signal, line 155: Haven't seen this syntax before, only used `--vers 1.0.0`. Maybe that'll work on the builders?

jack-signal: I had just assumed that syntax worked without testing, alas no

jrose-signal (approved):

PR #142, opened 2021-01-15
Fix the session encrypt/decrypt benchmark code

jack-signal:
I considered adding `cargo bench --no-run` to the GH actions build but that would add a whole second (release-mode) build to CI which I'm not sure is worth it.

### DIFF
--- a/rust/protocol/benches/session.rs
+++ b/rust/protocol/benches/session.rs
@@ -22,27 +22,54 @@ pub fn session_encrypt_result(c: &mut Criterion) -> Result<(), SignalProtocolErr
     block_on(alice_store.store_session(&bob_address, &alice_session_record, None))?;
     block_on(bob_store.store_session(&alice_address, &bob_session_record, None))?;
 
-    let message_to_decrypt = block_on(support::encrypt(&mut alice_store, &bob_address, "a short message"))?;
+    let message_to_decrypt = block_on(support::encrypt(
+        &mut alice_store,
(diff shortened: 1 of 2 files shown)


jrose-signal:
`cargo build --benches` or `cargo build --all-targets`, maybe? Or just `cargo check` if we're not going to run them?

…ooh, actually we've been doing this wrong for Clippy too. Let's do `cargo clippy --all-targets` from now on.

rust/protocol/benches/session.rs
```diff
@@ -1,86 +1,125 @@
...
     let alice_address = ProtocolAddress::new("+14159999999".to_owned(), 1);
     let bob_address = ProtocolAddress::new("+14158888888".to_owned(), 1);
 
     let mut alice_store = support::test_in_memory_protocol_store();
     let mut bob_store = support::test_in_memory_protocol_store();
 
-    alice_store.store_session(&bob_address, &alice_session_record)?;
-    bob_store.store_session(&alice_address, &bob_session_record)?;
+    block_on(alice_store.store_session(&bob_address, &alice_session_record, None))?;
+    block_on(bob_store.store_session(&alice_address, &bob_session_record, None))?;
```
jrose-signal, line 23: Since it's a benchmark I wonder if it's better to use `expect_ready` than `block_on`, the latter being a little heavier. We'd have to move it down in the stack though.

jrose-signal (approved):

PR #143, opened 2021-01-16
Migrate a big chunk of FFI/JNI to bridge_fn

jrose-signal:
I suggest reviewing commit-by-commit. In particular the commit that adds `bridge_fn_buffer` has much more going on than the others. It's a little unfortunate that that doesn't fall nicely out of the rest of the `bridge_fn` infrastructure, but the double output parameters for C and avoiding an extra copy for Java make it worth it in the end. Still, you can see it starting to pay off with the diff size.

### DIFF
--- a/java/java/src/main/java/org/signal/internal/Native.java
+++ b/java/java/src/main/java/org/signal/internal/Native.java
@@ -66,20 +66,20 @@ public final class Native {
 
   private Native() {}
 
-  public static native byte[] Aes256GcmSiv_Decrypt(long aesGcmSiv, byte[] msg, byte[] nonce, byte[] associatedData);
+  public static native byte[] Aes256GcmSiv_Decrypt(long aesGcmSiv, byte[] ctext, byte[] nonce, byte[] associatedData);
   public static native void Aes256GcmSiv_Destroy(long handle);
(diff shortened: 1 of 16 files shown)


java/java/src/main/java/org/signal/internal/Native.java
```diff
@@ -97,12 +97,12 @@ private Native() {}
   public static native byte[] HKDF_DeriveSecrets(int version, byte[] inputKeyMaterial, byte[] salt, byte[] info, int outputLength);
 
   public static native long[] IdentityKeyPair_Deserialize(byte[] data);
-  public static native byte[] IdentityKeyPair_Serialize(long publicKeyHandle, long privateKeyHandle);
+  public static native byte[] IdentityKeyPair_Serialize(long privateKey, long publicKey);
 
   public static native void NumericFingerprintGenerator_Destroy(long handle);
   public static native String NumericFingerprintGenerator_GetDisplayString(long handle);
   public static native byte[] NumericFingerprintGenerator_GetScannableEncoding(long handle);
-  public static native long NumericFingerprintGenerator_New(int iterations, int version, byte[] localIdentifier, byte[] localKey, byte[] remoteIdentifier, byte[] remoteKey);
```
jrose-signal, line 105: Whoops, these do not have the same behavior in Java and C at the moment. I'll take it out on Tue.

jack-signal (approved):

PR #146, opened 2021-01-16
Log more information when decryption fails

jack-signal:
Error log looks like this

```
Message from +14151111111:1 failed to decrypt; sender ratchet public key 01d5bbc17518fc09e3aa0cc7f43b627d5bce860f7bb2b1adc4676300569cbd30 message counter 1
Candidate session 0 failed with 'invalid ciphertext message', had 1 receiver chains
Receiver chain with sender ratchet public key 0501d5bbc17518fc09e3aa0cc7f43b627d5bce860f7bb2b1adc4676300569cbd30
Candidate session 1 failed with 'invalid ciphertext message', had 4 receiver chains
Receiver chain with sender ratchet public key 057b8c777abe702f861d6e77bf9d3b0a93a361ac24a2aa07cdbf52e413a4e61576
Receiver chain with sender ratchet public key 0571eab62e9ca2827962b7a41593bd76d2a7976762844af15d6ac1529af456ac0c
Receiver chain with sender ratchet public key 052dca6940acfb4a02da93bd649d400cfedf16156cff65fe0d6f4105b42956be2b
Receiver chain with sender ratchet public key 05d7a0b76243102318ed92a981640dde5ca10d7e3d2179d4ac51f35a55ebe28229
```

### DIFF
--- a/rust/protocol/src/session_cipher.rs
+++ b/rust/protocol/src/session_cipher.rs
@@ -179,7 +179,12 @@ pub async fn message_decrypt_prekey<R: Rng + CryptoRng>(
     )
     .await?;
 
-    let ptext = decrypt_message_with_record(&mut session_record, ciphertext.message(), csprng)?;
+    let ptext = decrypt_message_with_record(
+        &remote_address,
+        &mut session_record,
+        ciphertext.message(),
+        csprng,
+    )?;
 
     session_store
         .store_session(&remote_address, &session_record, ctx)
@@ -205,7 +210,8 @@ pub async fn message_decrypt_signal<R: Rng + CryptoRng>(
         .await?
         .ok_or(SignalProtocolError::SessionNotFound)?;
 
-    let ptext = decrypt_message_with_record(&mut session_record, ciphertext, csprng)?;
+    let ptext =
+        decrypt_message_with_record(&remote_address, &mut session_record, ciphertext, csprng)?;
 
     // Why are we performing this check after decryption instead of before?
     let their_identity_key = session_record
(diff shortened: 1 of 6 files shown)


rust/protocol/src/state/session.rs
```diff
@@ -134,6 +134,16 @@ impl SessionState {
         Ok(self.session.sender_chain.is_some())
     }
 
+    pub(crate) fn get_all_receiver_chains_and_counters(&self) -> Result<Vec<(Vec<u8>, u32)>> {
+        let mut results = vec![];
+        for chain in self.session.receiver_chains.iter() {
+            if let Some(ref chain_key) = chain.chain_key {
+                results.push((chain_key.key.clone(), chain_key.index));
```
jack-signal, line 141: @ehrenkret-signal Please confirm this is what you meant by ratchet key / current counter in the Google doc

Imperiopolis-Signal (approved):
other than the q on if it's the right key for @ehrenkret-signal, LGTM

rust/protocol/src/state/session.rs
ehrenkret-signal: chain_key does not sound like the right variable...

There should be one in the session state called ratchet key that should be a public key.

jack-signal: Yeah that seemed to me what we wanted here except there is no counter associated with that key
https://github.com/signalapp/libsignal-client/blob/master/rust/protocol/src/proto/storage.proto#L11
and the doc mentions the "current counter" associated with the receiver chain so I'm not sure what counter that would mean.

jack-signal: [Discussed with Ehren via Signal] what the doc meant by "ratchet key + current counter" was "sender ratchet public key" + "current chain counter". Doc updated.

rust/protocol/src/session_cipher.rs
```diff
@@ -238,11 +244,60 @@ pub async fn message_decrypt_signal<R: Rng + CryptoRng>(
...
+    ));
+
+    for (idx, (state, err)) in std::iter::once(record.session_state()?)
+        .chain(record.previous_session_states()?)
+        .zip(errs)
+        .enumerate()
+    {
+        let chains = state.all_receiver_chain_logging_info()?;
+        lines.push(format!(
+            "Candidate session {} failed with '{}', had {} receiver chains",
```
ehrenkret-signal, line 270: Ideally would include counters from the chains

rust/protocol/src/session_cipher.rs
```diff
@@ -238,11 +244,60 @@ pub async fn message_decrypt_signal<R: Rng + CryptoRng>(
...
+        ));
+
+        for chain in chains {
+            let chain_idx = match chain.1 {
+                Some(i) => format!("{}", i),
+                None => "missing in protobuf".to_string(),
+            };
+
+            lines.push(format!(
+                "Receiver chain with sender ratchet public key {} chain key index {}",
```
ehrenkret-signal, line 283: is the chain index in here the last counter received + 1 ? I believe that to be so but should be checked

PR #148, opened 2021-01-20
Changes to support building Node library on Windows

jack-signal:


rust/bridge/node/src/lib.rs
```diff
@@ -11,6 +11,19 @@ use std::ops::Deref;
...
+napi_get_all_property_names which was introduced in N-API 6. However
+N-API 6 was first supported in Node 12.70 while the version of
+Electron used in Desktop is only 12.30
+
+On Linux and macOS the link still succeeds and the symbol is simply
+undefined. However on Windows the DLL fails to link when this is
+missing.
+*/
+#[no_mangle]
+extern "C" fn napi_get_all_property_names() { std::process::abort(); }
```
jrose-signal, line 25: Neon 0.7.0 fixes this issue by allowing us to declare that we only need `napi-1` as a Cargo feature. #65 includes that change, but you could extract it here as well.

jrose-signal (approved):

package.json
```diff
@@ -11,7 +11,7 @@
   ],
   "scripts": {
     "build": "node-gyp build",
-    "tsc": "tsc -p node && cp node/*.d.ts node/dist",
```
jrose-signal, line 14: Looks like `tsc -p node --declarationDir node/dist` is close enough (possibly better, even). `lint` and `format` are just going to suffer, but that's fine.

jack-signal: Thanks for pointing that out `--declarationDir` seems much better than the possible alternatives here. And yeah I was content with leaving lint and format broken on Windows.

binding.gyp
```diff
@@ -15,14 +15,15 @@
...
-                    'action_name': 'node/build_node_bridge.sh',
+                    'action_name': 'node/build_node_bridge.py',
                     'action': [
-                        'env',
-                        'CONFIGURATION_NAME=<(CONFIGURATION_NAME)',
-                        'CARGO_BUILD_TARGET_DIR=<(INTERMEDIATE_DIR)/rust',
-                        'NODE_OS_NAME=<(NODE_OS_NAME)',
-                        'node/build_node_bridge.sh',
-                        '-o', '<(PRODUCT_DIR)/'],
+                        'python',
```
jrose-signal, line 20: Oh, this should be python3, shouldn't it. (And in the `#!` below.)

node/scripts/copy_repo.sh
```diff
@@ -9,7 +9,7 @@ set -euo pipefail
 
 SCRIPT_DIR=$(dirname "$0")
 cd "${SCRIPT_DIR}"/..
-. bin/build_helpers.sh
+. ../bin/build_helpers.sh
```
jrose-signal, line 12: Nitpick: the intent of this prefix for the other scripts was that they always run from the root directory, i.e. change the `cd` instead. You don't have to do that but I think it could make things simpler.

package.json
jrose-signal: Ugh, I was wrong about this. We already had the TypeScript-generated .d.ts files going into `dist`; it's libsignal_client.d.ts that's missing.

jrose-signal: You've got it manually in the distribution script, so I guess it's just something I'll have to handle if I want `yarn pack` to replace that script.

PR #149, opened 2021-01-21
Add support for Node to libsignal-bridge

jrose-signal:
This should in theory give us *all* of the functions already in libsignal-bridge for free. There are a few complicated bits:

- Borrowing a JsBuffer requires the JS engine to be locked, but returning a new buffer requires using the JS engine. That means having to incur a copy-in (for `&[u8]` parameters) or a copy-out (for `&[u8]` results) unless we make `bridge_fn` signatures more complicated. For now I've gone with a copy-in because it was easier to implement, but a copy-out probably makes more sense in the long run.

- I wanted to avoid Neon's explicit registration list because it was one more thing to keep in sync, so I've used the [`linkme`](https://docs.rs/linkme/0.2.4/linkme/index.html) crate to have the linker build a static list that can be iterated through at runtime. If you think this is too much magic I'll drop it, though.

~~Still a draft because I built it on top of Neon 0.7.0, which I stole from the futures PR. I'll rebase on something else when we're ready to merge if we still don't have that in. (But meanwhile I'd want to port all the Swift-side tests we have to Node.)~~ Tests can come as we expose the various APIs in TypeScript.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -751,6 +751,7 @@ dependencies = [
  "libsignal-bridge-macros",
  "libsignal-protocol-rust",
  "log",
+ "neon",
  "paste",
  "rand",
 ]
@@ -798,6 +799,7 @@ dependencies = [
 name = "libsignal-node"
 version = "0.1.0"
 dependencies = [
+ "libsignal-bridge",
  "libsignal-protocol-rust",
  "neon",
  "neon-build 0.5.3",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,6 +750,7 @@ dependencies = [
  "libc",
  "libsignal-bridge-macros",
  "libsignal-protocol-rust",
+ "linkme",
  "log",
  "neon",
  "paste",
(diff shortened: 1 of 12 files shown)


jrose-signal:
Guess the `linkme` part isn't working correctly on Linux yet…

jrose-signal:
#159 makes the macro part of this simpler, but I also worked on it on top of #154 and #156. They don't really overlap, though, so what do you think of merging them all into this branch once #154's reviewed by someone on Desktop?

jack-signal (approved):
`linkme` looks pretty magical but worth it here since it saves having to name each and every API endpoint a second time.

jrose-signal:
> `linkme` looks pretty magical but worth it here since it saves having to name each and every API endpoint a second time.

Yep. I'm not super happy about adding more magic but I do know how this mechanism works and it doesn't add runtime overhead. I only went with it in the end because we *are* trying to reduce repetition in the FFI.

PR #151, opened 2021-01-21
Bump Neon to 0.7.0 and use feature napi-1

jack-signal:
Neon by default requires N-API 6 and the version of Electron that is used in Desktop only supports N-API 5. Without this feature, some weird hacks are required to build on Windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,24 +64,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7404febffaa47dac81aa44dba71523c9d069b1bdc50a77db41195149e17f68e5"
-dependencies = [
- "memchr",
-]
-
-[[package]]
-name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jrose-signal (approved):

jrose-signal:
There are a few cleanups this enables (for instance, `neon-build` is no longer necessary), but we can pick those up in #65.

Xashyar:
@jrose-signal 
Hi, can this be reverted now that Desktop uses Electron 11.0 or does N-API 6 require Electron 12.0?

jrose-signal:
I don't think there's any harm in continuing to declare the minimum N-API we need. (It's up to napi-4 by now.) Do you have a specific need for N-API 6?

Xashyar:
Nothing specific, It's just that node.js documentation recommends N-API 3 or higher:

> Node.js 8.0.0 included N-API as experimental. It was released as N-API version 1 but continued to evolve until Node.js 8.6.0. The API is different in versions prior to Node.js 8.6.0. **We recommend N-API version 3 or later.**

https://nodejs.org/api/n-api.html

jrose-signal:
Ah, fair enough. I don't think Neon is even taking this into account, though—[their features are currently inclusive](https://github.com/neon-bindings/neon/blob/main/Cargo.toml#L48-L60). Anyway, since we *are* up to using N-API 4 features at the head of the branch I think we'll be okay.

PR #154, opened 2021-01-26
Node: Add an entry point for logging

jrose-signal:
Sketch out a logging callback for Node clients, like #150 for Java and #90 for Swift.

### DIFF
--- a/node/test/PublicAPITest.ts
+++ b/node/test/PublicAPITest.ts
@@ -1,10 +1,21 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
 import { assert } from 'chai';
 import * as SignalClient from '../index';
+
+SignalClient.initLogger(
+  SignalClient.LogLevel.Trace,
+  (level, target, fileOrNull, lineOrNull, message) => {
+    const targetPrefix = target ? '[' + target + '] ' : '';
+    const file = fileOrNull ?? '<unknown>';
+    const line = lineOrNull ?? 0;
+    // eslint-disable-next-line no-console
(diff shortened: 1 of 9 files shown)


jack-signal (approved):
lgtm

jrose-signal:
Desktop folks, mostly hoping for confirmation that an entry point like `initLogger` will work for integrating into the app.

node/test/PublicAPITest.ts
```diff
@@ -1,11 +1,22 @@
...
 import * as SignalClient from '../index';
 
+SignalClient.initLogger(
+  SignalClient.LogLevel.Trace,
+  (level, target, fileOrNull, lineOrNull, message) => {
+    const targetPrefix = target ? '[' + target + '] ' : '';
+    const file = fileOrNull ?? '<unknown>';
+    const line = lineOrNull ?? 0;
+    // eslint-disable-next-line no-console
+    console.log(targetPrefix + file + ':' + line + ': ' + message);
```
scottnonnenberg-signal, line 16: To work smoothly within our system, you'd want to map your error level to our levels. In the renderer process, we have `window.log.warn`, `window.log.error`, and `window.log.info`.

Note that this will not work if you are loaded in our main process. In that world, `console.log/warn/error` are available.

We could potentially override `console.warn/error` in the renderer process as well. To make this easier.

jrose-signal: This is the local test; we'll be able to call `initLogger` however it makes sense for the actual app. I'm hesitant to have the main package set it up automatically because of load order, though.

scottnonnenberg-signal:
> Desktop folks, mostly hoping for confirmation that an entry point like `initLogger` will work for integrating into the app.

It does look like that callback will be sufficient.

jrose-signal:
We can tweak things further if it turns out there's a better way later, just wanted to make sure this wasn't going to be a totally incompatible approach.

PR #156, opened 2021-01-26
Node: Autogenerate libsignal_client.d.ts like we do Native.java

jrose-signal:
This collects doc comments of the form "ts: <some TS declaration>", which can be written manually *or* generated by the various "bridge" macros. If the declaration looks like a function, it also does some substitution of Rust types for TypeScript types, to make autogeneration easier.


rust/bridge/node/bin/gen_ts_decl.py
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env python3
```
jrose-signal, line 1: This is based on `gen_java_decl.py`. Do you think it's worth me putting in effort to factor out a shared infrastructure?

jack-signal: IMO lets not overthink it at this stage. We can get Desktop onto Rust with everything supported and then think about merging scripts and other cleanups.

It's a different problem from the general issue of the bridge - there the effort of having to support multiple bridges increases as we increase the number of bridged symbols, while these scripts only need to change when we need to support a new in/out type which is much less common.

jack-signal (approved):

PR #157, opened 2021-01-27
JNI: bridge_fn should not care about refs vs values at the macro level

jrose-signal:
That distinction can be made in the main library by always doing a two-step "borrow-then-load", where sometimes the "borrow" operation is a no-op. This makes it easier to support inputs of the form `Option<&PublicKey>`, which shows up in the pre-key operations.


jack-signal (approved):
Nice

PR #158, opened 2021-01-27
Map errors through the bridge more carefully

jack-signal:
Remove several errors that were very once-off.

In Java avoid throwing RuntimeException unless it's an internal error
that we really should crash on.


rust/bridge/jni/src/lib.rs
```diff
@@ -202,20 +202,16 @@ pub unsafe extern "C" fn Java_org_signal_client_internal_Native_PreKeyBundle_1Ne
...
 
-        let prekey = native_handle_cast_optional::<PublicKey>(prekey_handle)?.map(|k| *k);
-
-        let prekey_id = if prekey_id < 0 {
-            None
-        } else {
-            Some(jint_to_u32(prekey_id)?)
+        let prekey = match native_handle_cast_optional::<PublicKey>(prekey_handle)?.map(|k| *k) {
+            None => None,
+            Some(key) => Some((jint_to_u32(prekey_id)?, key)),
```
jrose-signal, line 207: May be clearer as `map`?

rust/protocol/src/curve.rs
```diff
@@ -102,10 +102,7 @@ impl PublicKey {
         match self.key {
             PublicKeyData::DjbPublicKey(pub_key) => {
                 if signature.len() != 64 {
-                    return Err(SignalProtocolError::MismatchedSignatureLengthForKey(
-                        KeyType::Djb,
-                        signature.len(),
-                    ));
+                    return Ok(false);
```
jrose-signal, line 105: This seems like a bad change to me. If clients want to handle invalid signatures the same way as valid-but-incorrect signatures, that should be up to them, but the actual implementation in libsignal-protocol should keep those separate.

jack-signal: Neither libsignal-protocol-java nor SignalProtocolKit make special exception for the wrong length, they just return false. The C library used by Android has an explicit check on the length and jumps immediately to reject. SignalProtocolKit, AFAICT, just assumes the signature is at least 64 bytes and uses whatever random data follows as the signature, which will also cause it to fail to verify.

Also there are a lot of forms of invalid signature (for example r/s can be greater than the group order) I'm not sure why the length should be considered special here.

jrose-signal: All right, withdrawn.

rust/bridge/jni/src/lib.rs
jack-signal: I wasn't able to find a way using `map` that wasn't significantly more complicated/harder to read, maybe missing some obvious approach though?

jrose-signal: Mm, the `?` for `jint_to_u32` does make it tricky, huh. Okay, never mind.

jrose-signal (approved):

PR #160, opened 2021-01-28
Reject SenderCertificates without UUIDs

jrose-signal:
Additionally, never look up a session by e164 when decrypting sealed-sender messages.

This is an API-breaking change for both Java and Swift clients; certain fields and arguments are no longer Optional. On top of that, some tests may need to be updated to provide UUIDs instead of just phone numbers.

I admit I started making this change and it snowballed well beyond "just prefer UUIDs over e164", so if we're not comfortable with this larger alteration I can go back to something nicely self-contained.


Imperiopolis-Signal (approved):
LGTM

jack-signal (approved):

greyson-signal (approved):

PR #163, opened 2021-01-30
Compatibility with cocoapods-binary

jrose-signal:
Three changes that allow libsignal-client to be prebuilt for iOS clients using [cocoapods-binary](https://github.com/leavez/cocoapods-binary):
- Use CocoaLumberjack directly instead of SignalCoreKit's OWSLog
- Move signal_ffi.h into SignalClient.framework (and add a hacky unstable compiler flag to keep the SwiftPM build working)
- Skip build_ffi.sh when the pod is prebuilt

### DIFF
--- a/swift/Sources/SignalClient/Aes256GcmSiv.swift
+++ b/swift/Sources/SignalClient/Aes256GcmSiv.swift
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-import SignalFfi
 import Foundation
 
 public class Aes256GcmSiv: ClonableHandleOwner {
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -255,12 +255,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Check out SignalCoreKit
-        uses: actions/checkout@v2
-        with:
-          repository: signalapp/SignalCoreKit
-          path: SignalCoreKit
-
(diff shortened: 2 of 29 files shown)


jack-signal:
I don't really have the background to review this but this change does seem to break the Linux build, ``swift test`` now gives

```
<unknown>:0: warning: missing submodule 'Dispatch'
<unknown>:0: error: cannot load underlying module for 'Dispatch'
```

I guess it is not strictly necessary for our purposes that Swift be buildable on Linux but it was nice.

jrose-signal:
Hm, I'm not seeing that with a (manually-downloaded) Swift 5.3, though there is another error. What's your setup?

jack-signal:
I'm still on Swift 5.2.4 so that's probably it, will update

jack-signal:
Works fine with 5.3.3 except for one build error fixed in #166

swift/Sources/SignalClient/Aes256GcmSiv.swift
```diff
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-import SignalFfi
+// import SignalFfi
```
charlesmchen-signal, line 6: Remove (here and below)?

jrose-signal: Yeah, I guess. I felt weird not having it be marked at all but it's not serving a purpose in either build system at the moment.

charlesmchen-signal (approved):
LGTM

PR #164, opened 2021-01-31
protocol: upgrade prost to 0.7

svmhdvn:
This fixes a proto compilation error when packaging this as an aport on alpine linux.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,6 +179,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0e4cec68f03f32e44924783795810fa50a7035d8c8ebe78580ad7e6c703fba38"
 
 [[package]]
+name = "bytes"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b700ce4376041dcd0a327fd0097c41095743c4c8af8887265942faf1100bd040"
+
+[[package]]
 name = "cast"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -231,7 +237,7 @@ version = "4.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2809f67365382d65fd2b6d9c22577231b954ed27400efeafbe687bda75abcc0b"
 dependencies = [
- "bytes",
+ "bytes 0.5.6",
  "memchr",
  "pin-project-lite",
 ]
@@ -253,7 +259,7 @@ dependencies = [
  "clap",
  "criterion-plot",
  "csv",
- "itertools 0.9.0",
+ "itertools",
  "lazy_static",
  "num-traits",
  "oorandom",
@@ -275,7 +281,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e022feadec601fba1649cfa83586381a4ad31c6bf3a9ab7d408118b05dd9889d"
 dependencies = [
  "cast",
- "itertools 0.9.0",
+ "itertools",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


jack-signal:
Can you give a little background on what is wrong with prost 0.6 on Alpine?

Cargo.lock
```diff
@@ -184,6 +184,12 @@ version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0e4cec68f03f32e44924783795810fa50a7035d8c8ebe78580ad7e6c703fba38"
 
+[[package]]
+name = "bytes"
```
jrose-signal, line 188: The double 'bytes' dependency is unfortunate; if we do take this we'll probably want to see what depends on 'bytes' 0.5.6.

svmhdvn:
on master HEAD, I am getting the following error when running a `cargo build` from the root:
```
error: failed to run custom build command for `libsignal-protocol v0.1.0 (/home/siva/src/libsignal-client/rust/protocol)`

Caused by:
  process didn't exit successfully: `/home/siva/src/libsignal-client/target/debug/build/libsignal-protocol-47c325af47feaeef/build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', rust/protocol/build.rs:13:52
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

See https://github.com/danburkert/prost/pull/410. This commit is included in the 0.7.0 release.

After upgrading to prost 0.7.0, the build succeeds.

Cargo.lock
svmhdvn: I'm not very experienced with the Rust build infrastructure, what should I do to resolve this?

jrose-signal: `cargo tree --invert bytes` can show all the packages that depend on `bytes`; fortunately…

```
bytes v0.5.6
├── libsignal-protocol v0.1.0 (/Users/jordan/dev/libsignal-client/rust/protocol)
├── prost v0.6.1
│   ├── libsignal-protocol v0.1.0 (/Users/jordan/dev/libsignal-client/rust/protocol)
│   ├── prost-build v0.6.1
│   │   [build-dependencies]
│   │   └── libsignal-protocol v0.1.0 (/Users/jordan/dev/libsignal-client/rust/protocol)
│   └── prost-types v0.6.1
│       └── prost-build v0.6.1 (*)
├── prost-build v0.6.1 (*)
└── prost-types v0.6.1 (*)
```

…the only other user seems to be libsignal-protocol itself! So you can bump that version as well in the Cargo.toml.

svmhdvn: Looks like `cargo tree -d -e normal --workspace` shows that `bytes` is also an indirect dependency for `jni` through `combine` and it's using a much older version. Is the only solution to update it in `combine` upstream?

```
bytes v0.5.6
└── combine v4.3.2
    └── jni v0.17.0
        ├── libsignal-bridge v0.1.0 (/home/siva/src/libsignal-client/rust/bridge/shared)
        │   ├── libsignal-ffi v0.1.2 (/home/siva/src/libsignal-client/rust/bridge/ffi)
        │   ├── libsignal-jni v0.2.3 (/home/siva/src/libsignal-client/rust/bridge/jni)
        │   └── libsignal-node v0.3.0 (/home/siva/src/libsignal-client/rust/bridge/node)
        └── libsignal-jni v0.2.3 (/home/siva/src/libsignal-client/rust/bridge/jni)

bytes v1.0.1
├── libsignal-protocol v0.1.0 (/home/siva/src/libsignal-client/rust/protocol)
│   ├── libsignal-bridge v0.1.0 (/home/siva/src/libsignal-client/rust/bridge/shared) (*)
│   ├── libsignal-ffi v0.1.2 (/home/siva/src/libsignal-client/rust/bridge/ffi)
│   ├── libsignal-jni v0.2.3 (/home/siva/src/libsignal-client/rust/bridge/jni)
│   └── libsignal-node v0.3.0 (/home/siva/src/libsignal-client/rust/bridge/node)
└── prost v0.7.0
    └── libsignal-protocol v0.1.0 (/home/siva/src/libsignal-client/rust/protocol) (*)
```

jrose-signal: Oops, I always forget `--workspace` when I run it manually. It looks like the 4.5.x series of `combine` supports bytes v1, but there's not a new enough `jni` to use that `combine`. So it looks like we have a trade-off here: add duplicate code size to our resulting binary, or don't support Alpine Linux. I'm trying to think of ways to support both but have the right defaults, but they'd all mean threading through Cargo features all over the place.

jack-signal:
Resolved in #222

PR #166, opened 2021-02-01
Avoid using %@ with NSLog

jack-signal:
Unfortunately not implemented on Linux, instead one must convert to a C string and use %s


jrose-signal (approved):
Heh, I made the exact same change and pulled it into #163, but it's been broken in master for a while.

PR #167, opened 2021-02-01
Add int getter support to the bridge

jack-signal:
First commit only converts ProtocolAddress.deviceId I'll push conversions for the rest onto this branch.


jrose-signal:
Superseded by #168, though it's a lot more to review!

PR #168, opened 2021-02-02
Bridge: move all macro-based APIs into bridge/shared

jrose-signal:
The biggest change here is a new `bridge_get`, which constructs a `bridge_fn` around an existing method given only a result type. This reduces duplication and also exposes many more APIs to Node for free.

With this, the only non-store-based APIs left in libsignal-ffi and libsignal-jni are:

- `signal_hkdf_derive`, which is funky because it has an in-place output buffer under FFI (but not JNI)
- `signal_identitykeypair_deserialize`, which has to return multiple values
- `signal_session_record_archive_current_state`, which is our only mutating operation so far
- `signal_session_record_has_current_state`, which…honestly could have been folded in here as an FFI-only `bridge_fn`.
- `signal_fingerprint_new`, which takes PublicKeys in FFI but plain buffers in JNI
- `signal_ciphertext_*`, which are only exposed in FFI and so I missed them, oops.
- `SessionRecord_*` APIs that are only exposed in JNI, also oops.
- `SenderCertificate_PreferredAddress`, which is hopefully being removed in #160
- `UnidentifiedSenderMessageContent_New` and `UnidentifiedSenderMessage_New`, which are JNI-only and only for testing, also oops.

I'll continue working through these (easy ones first).


jack-signal (approved):
Looks good, ship it!

PR #170, opened 2021-02-02
More Node bindings

jack-signal:
Aes256GcmSiv
ProtocolAddress
PreKeyRecord
PreKeyBundle
PreKeySignalMessage
SignalMessage

### DIFF
--- a/node/index.ts
+++ b/node/index.ts
@@ -1,5 +1,5 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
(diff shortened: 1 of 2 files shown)


node/index.ts
```diff
@@ -75,3 +135,229 @@ export class PrivateKey {
...
+    signed_prekey_id: number,
+    signed_prekey: PublicKey,
+    signed_prekey_signature: Buffer,
+    identity_key: PublicKey
+  ) {
+    this.nativeHandle = SC.PreKeyBundle_New(
+      registration_id,
+      device_id,
+      prekey_id,
+      prekey != null ? prekey._unsafeGetNativeHandle() : null,
```
jrose-signal, line 156: TypeScript has optional chaining like Swift, so you can make this `prekey?._unsafeGetNativeHandle()`.

node/index.ts
```diff
@@ -75,3 +135,229 @@ export class PrivateKey {
...
+
+  private constructor(handle: SignalClient.SignalMessage) {
+    this.nativeHandle = handle;
+  }
+
+  _unsafeGetNativeHandle(): SignalClient.SignalMessage {
+    return this.nativeHandle;
+  }
+
+  new(
```
jrose-signal, line 253: Missing `static`.

node/index.ts
```diff
@@ -75,3 +135,229 @@ export class PrivateKey {
...
+}
+
+export class PreKeySignalMessage {
+  private readonly nativeHandle: SignalClient.PreKeySignalMessage;
+
+  private constructor(handle: SignalClient.PreKeySignalMessage) {
+    this.nativeHandle = handle;
+  }
+
+  new(
```
jrose-signal, line 318: Missing `static`

node/index.ts
```diff
@@ -75,3 +135,229 @@ export class PrivateKey {
...
+        preKeyId,
+        signedPreKeyId,
+        baseKey._unsafeGetNativeHandle(),
+        identityKey._unsafeGetNativeHandle(),
+        signalMessage._unsafeGetNativeHandle()
+      )
+    );
+  }
+
+  deserialize(buffer: Buffer): PreKeySignalMessage {
```
jrose-signal, line 340: Missing `static`

node/index.ts
```diff
@@ -75,3 +135,229 @@ export class PrivateKey {
     );
   }
 }
+
+export class PreKeyBundle {
+  private readonly nativeHandle: SignalClient.PreKeyBundle;
+
+  constructor(
```
jrose-signal, line 142: I feel weird having some classes use constructors and others use `static new()`. Maybe they should all use `new()`, with the constructor limited to the "from native handle" use case?

node/index.ts
```diff
@@ -11,6 +11,58 @@ const SC = bindings('libsignal_client_' + os.platform()) as typeof SignalClient;
...
+    handle: SignalClient.ProtocolAddress
+  ): ProtocolAddress {
+    return new ProtocolAddress(handle);
+  }
+
+  static new(name: string, deviceId: number): ProtocolAddress {
+    return new ProtocolAddress(SC.ProtocolAddress_New(name, deviceId));
+  }
+
+  name(): string {
```
jrose-signal, line 57: I think we already checked with the Desktop folks over whether to use method or property syntax for things like this, but it's worth double-checking.

node/index.ts
jack-signal: Unfortunately TypeScript maps null to undefined thus `prekey?._unsafeGetNativeHandle()` is `PublicKey | undefined` not `PublicKey | null` - I assume there is a way around this but haven't found it yet.

node/index.ts
jack-signal: Evan's comment (in #124) on this "I'd call this getPublicKey and keep it as a function. I rarely see accessors used in JavaScript-land. (This is anecdotal, and other Desktop devs might have different experiences.)"

jrose-signal (approved):

PR #171, opened 2021-02-02
Fix compilation problem

jack-signal:
Introduced by merging #118 

Don't really understand what happened here since this compiled fine with the same rustc when the PR was opened.


jrose-signal (approved):
It's that SignalProtocolError isn't Eq anymore, so `assert_eq!` chokes on a Result that might contain one. You could switch to `matches!` too, but this is fine.

jack-signal:
Oh right now we are carrying around a `Box<dyn Error>` now, I forgot about that.

PR #172, opened 2021-02-03
Support &mut arguments and Result<(), Error> results

jrose-signal:
Neon enforces boxed values to be immutable, so I had to jump through some hoops with RefCell. FFI also expects an out-pointer for any declared results, but error-only results shouldn't get one. The easiest way to detect that was to add another macro variant, `bridge_fn_void`.

### DIFF
--- a/rust/bridge/shared/src/node/convert.rs
+++ b/rust/bridge/shared/src/node/convert.rs
@@ -269,6 +269,10 @@ impl<'a, T: Value> ResultTypeInfo<'a> for Handle<'a, T> {
     }
 }
 
+pub(crate) unsafe fn extend_lifetime_to_static<T>(some_ref: &T) -> &'static T {
+    std::mem::transmute::<&'_ T, &'static T>(some_ref)
+}
+
 macro_rules! node_bridge_handle {
     ( $typ:ty as false ) => {};
     ( $typ:ty as $node_name:ident ) => {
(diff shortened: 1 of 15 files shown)


rust/bridge/shared/src/node/convert.rs
```diff
@@ -283,9 +297,67 @@ macro_rules! node_bridge_handle {
...
+            // We know the RefCell can't move because we can't know how many JS references there are referring to the JsBox.
+            type StoredType = (node::Handle<'a, Self::ArgType>, std::cell::Ref<'static, $typ>);
+            fn borrow(
+                _cx: &mut node::FunctionContext,
+                foreign: node::Handle<'a, Self::ArgType>,
+            ) -> node::NeonResult<Self::StoredType> {
+                let cell: &std::cell::RefCell<_> = &***foreign;
+                let cell_with_extended_lifetime = unsafe { node::extend_lifetime_to_static(cell) };
+                Ok((foreign, cell_with_extended_lifetime.borrow()))
+            }
```
jrose-signal, line 316: This "extend lifetime" bit here is a bit sketchy; the alternative would be changing the macro's two-step load into a three-step load so that we could clearly borrow the Handle, the RefCell inside it, and the object inside that. I asked some friends for help poking at this problem and one of them had this to say about my solution:

> Your coworker definitely shouldn't accept that! Being slightly more verbose is never a justification for unsafe code.

> The way to think about unsafe code is that *at best,* writing a single line of unsafe code is lots of work (in justifying and maintaining it)
> So the only time you would want to do it is if the alternative is either imposssible or even more work than that, which is rare.

I promised them I'd include this objection to let you decide if it's worth it. :-) In my mind, the justification work here *was* better than making the macros and traits permanently more complicated, but it's easy for me to say that, having come up with this solution.

jrose-signal: Tagging @kjvalencik as well: I admit I'm putting constraints on myself, but this feels something that would be fixed if you could go from `Handle<'a, JsBox<T>>` to `&'a T`. Is that safe? I don't have 100% solid intuition on what the lifetime means other than preventing escapes, and likewise I don't know whether it's guaranteed that a given Handle (with no Roots) stays alive until the function exits.

kjvalencik: @jrose-signal That _should_ be safe because of the following:

* It's not possible to get an exclusive reference (`&mut T`). This is by design because Neon has no way to prevent JavaScript from aliasing the `JsBox`
* `JsBox` cannot be garbage collected while the `Context` it is associated with is in scope

The `JsBox` itself is just a wrapper around a pointer that can be freely cloned. There's no reference counting or anything else going on. If this _wasn't_ safe, then `js_box.clone()` would also be unsafe.

It seems reasonable to add this `Deref` implementation to Neon.

jrose-signal: We've got a chained Deref already, but it's not actually good enough, because that goes from `&'b Handle<'a, JsBox<T>> -> &'b T`. Still, that's only really a problem in cases like these where you can't also store the Handle.

jack-signal: I am ok with the unsafe here (vs alternative of making the traits 3 phase) on the following logic

- There are a lot of trait implementations, the Cartesian product of the types we support and the languages we support. If removing the unsafe here was a matter of adding say 20 or 100 more lines, here in this one spot, then removing it would be slam dunk. But instead it adds an ongoing complexity tax, even on languages that are not affected, across the whole bridge.
- There is a decent argument that this is unsafe (wrt Rust) but that Electron maintains the invariants we are relying on.
- There is some possibility that in a future version of Neon and/or Electron we can remove the unsafety.

jack-signal (approved):

PR #173, opened 2021-02-03
Add HKDF to Node via bridge

jack-signal:


rust/bridge/shared/src/lib.rs
```diff
@@ -50,6 +50,25 @@ bridge_handle!(SignedPreKeyRecord);
...
+fn HKDF_deriveSecrets(output_length: u32,
+                      version: u32,
+                      input_key_material: &[u8],
+                      label: &[u8]) -> Result<Vec<u8>, SignalProtocolError> {
+    let kdf = HKDF::new(version)?;
+    Ok(kdf.derive_secrets(input_key_material, label, output_length as usize)?.to_vec())
+}
+
+#[bridge_fn(ffi = false, jni = false)]
+fn HKDF_deriveSaltedSecrets(output_length: u32,
```
jrose-signal, line 63: Do we need both variants, or should the salt just be a nullable parameter? Or do we need the unsalted version at all? (With the Swift bindings we just dropped the non-salted entry point.)

jrose-signal (approved):

PR #175, opened 2021-02-04
Bridge: Move nearly all remaining APIs into bridge/shared

jrose-signal:
Update from #168, built on top of #172 (which is complex enough on its own). The only remaining non-store, non-error-or-buffer-related APIs are
- `IdentityKeyPair_Deserialize` (and its FFI equivalent), which has to return multiple values
- `SenderCertificate_PreferredAddress`, which is hopefully being removed in #160.


rust/bridge/shared/src/node/convert.rs
```diff
@@ -279,6 +245,40 @@ impl<'a, T: Value> ResultTypeInfo<'a> for Handle<'a, T> {
...
+        impl SimpleArgTypeInfo for $typ {
+            type ArgType = JsNumber;
+            fn convert_from(
+                cx: &mut FunctionContext,
+                foreign: Handle<Self::ArgType>,
+            ) -> NeonResult<Self> {
+                let value = foreign.value(cx);
+                if !can_convert_js_number_to_int(value, 0.0..=<$typ>::MAX.into()) {
+                    return cx.throw_range_error(format!(
+                        "integer overflow during conversion of {}",
```
jack-signal, line 259: Might be helpful for debugging to include $typ in the message?

jack-signal (approved):

PR #179, opened 2021-02-04
Add Fingerprint API to Node binding

jack-signal:


node/index.ts
```diff
@@ -38,6 +38,82 @@ export class HKDF {
...
 }
 
+export class ScannableFingerprint {
+  private readonly scannable: Buffer;
+
+  private constructor(scannable: Buffer) {
+    this.scannable = scannable;
+  }
+
+  static fromBuffer(scannable: Buffer): ScannableFingerprint {
```
jrose-signal, line 48: Swift doesn't expose this or DisplayableFingerprint.fromString; are they expected to be needed?

jack-signal: There shouldn't be any reason for Desktop proper to call these. It's a similar story to Java, where Android doesn't directly create a DisplayableFingeprint but we expose a constructor because we create the object from Java. And we do have those constructors in Swift but they are `internal`. But sadly no `pub(crate)` equivalent for TS. :(

I could `_` prefix these to give a hint that the app isn't suppose to be using these?

jrose-signal: >< Right, I forgot `private` means "this class" and not "this file". I like the underscore prefix though.

jrose-signal (approved):

PR #180, opened 2021-02-04
Add logging that matches SignalProtocolKit (the ObjC library)

jrose-signal:


rust/protocol/src/session_cipher.rs
```diff
@@ -96,6 +104,10 @@ pub async fn message_encrypt(
         )
         .await?
     {
+        log::warn!(
+            "Previously known identity key for while encrypting for recipient: {}",
```
jack-signal, line 108: I found this phrasing really hard to parse. "for while encrypting"?

How about "Identity key {} is not trusted for remote address {}"?

rust/protocol/src/session_cipher.rs
```diff
@@ -228,6 +240,10 @@ pub async fn message_decrypt_signal<R: Rng + CryptoRng>(
...
```
jack-signal, line 244: OK I see these are directly from AxolotlKit but I still think they are not even English. :)

rust/protocol/src/session_cipher.rs
```diff
@@ -296,21 +312,53 @@ fn decrypt_message_with_record<R: Rng + CryptoRng>(
     ciphertext: &SignalMessage,
     csprng: &mut R,
 ) -> Result<Vec<u8>> {
+    let log_decryption_failure = |state: &SessionState, error: &SignalProtocolError| {
+        log::error!(
+            "Failed to decrypt whisper message with ratchet key: {} and counter: {}. \
+             Session loaded for {}. Local session has base key: {} and counter: {}. {}",
+            hex::encode(ciphertext.sender_ratchet_key().public_key_bytes().unwrap()),
```
jack-signal, line 319: Is this `unwrap` really safe?

Might be worth adding a util that hex encodes an `Optional<AsRef<[u8]>>` that returns something like "{missing}" when `None`?

rust/protocol/src/session_cipher.rs
```diff
@@ -296,21 +312,53 @@ fn decrypt_message_with_record<R: Rng + CryptoRng>(
...
         match result {
             Ok(ptext) => {
+                log::debug!(
+                    "successfully decrypted with current session state (base key {})",
+                    hex::encode(
+                        current_state
+                            .sender_ratchet_key()
+                            .unwrap()
+                            .public_key_bytes()
+                            .unwrap()
```
jack-signal, line 351: Again concerned about the unwraps though these are probably safe since if we were able to decrypt the key is probably there...

rust/protocol/src/state/session.rs
```diff
@@ -210,6 +210,10 @@ impl SessionState {
         self.session.receiver_chains.push(chain);
 
         if self.session.receiver_chains.len() > consts::MAX_RECEIVER_CHAINS {
+            log::info!(
+                "Trimming excessive receiver_chain count: {}",
```
jack-signal, line 214: Worth logging a session id here?

rust/protocol/src/session_cipher.rs
```diff
@@ -342,6 +401,20 @@ fn decrypt_message_with_record<R: Rng + CryptoRng>(
...
+                "No valid session for recipient: {}, current session base key {}, number of previous states: {}",
+                remote_address,
+                hex::encode(current_state.sender_ratchet_key().unwrap().public_key_bytes().unwrap()),
+                record.previous_session_states().unwrap().count(),
+            );
+        } else {
+            log::error!(
+                "No valid session for recipient: {}, (no current session state), number of previous states: {}",
+                remote_address,
+                record.previous_session_states().unwrap().count(),
```
jack-signal, line 415: Worth adding to `SessionState` a

`pub(crate) fn previous_session_state_count() -> usize`?

rust/protocol/src/session_cipher.rs
jack-signal: This specific value is logged regularly - maybe add a `fn` to `SessionState` to handle it?

rust/protocol/src/session_cipher.rs
```diff
@@ -322,17 +370,28 @@ fn decrypt_message_with_record<R: Rng + CryptoRng>(
...
         match result {
             Ok(ptext) => {
+                log::info!(
+                    "successfully decrypted with PREVIOUS session state (base key {})",
+                    hex::encode(
+                        previous
+                            .sender_ratchet_key()
+                            .unwrap()
+                            .public_key_bytes()
+                            .unwrap()
```
jack-signal, line 384: Could (imo should) be a `fn` on `SessionState` to return this as above

jack-signal (commented):
Same few items are pulled from SessionState worth adding a fn to encapulsate that and also any use of `unwrap` makes me nervous.

Especially because I'm not totally sure how much information actually makes it way to the app log if we panic... if all we see is "unwrap failed" we are going to be sad and it will be a pita to debug. At the least we should use `expect` everywhere with a reason of why we believe the value will never be Err/None.

rust/protocol/src/session_cipher.rs
jrose-signal: Honestly either `previous_session_states` shouldn't be failable or it should be handled here.

jack-signal (approved):

PR #182, opened 2021-02-05
Swift: fix IdentityKeyPair.init(bytes:)

### LINKED ISSUE #155: Deserializations bug of IdentityKeyPair in Swift
Swift Struct IdentityKeyPair has a logical error when trying to deserialize IdentityKeyPair from bytes in its init method. the current implementation looks like this: 

```swift
    public init<Bytes: ContiguousBytes>(bytes: Bytes) throws {
        var pubkeyPtr: OpaquePointer?
        var privkeyPtr: OpaquePointer?
        try bytes.withUnsafeBytes {
            try checkError(signal_identitykeypair_deserialize(&pubkeyPtr, &privkeyPtr, $0.baseAddress?.assumingMemoryBound(to: UInt8.self), $0.count))
        }

        publicKey = PublicKey(owned: pubkeyPtr!)
        privateKey = PrivateKey(owned: privkeyPtr!)
    }
```

The order of &privkeyPtr and &pubkeyPtr are mixed up. 

or it could be the order in the `serialize()` method. I think either way will fix it. I tried changing the order in the `init` and it seems to work. I did not try to change the order inside `serialize()` method

you can easily reproduce it by generating a new `IdentityKeyPair` serialize it and then deserialize it and check that public keys do not match as well as private keys.

jrose-signal:
It called the underlying Rust entrypoint with the arguments in the wrong order. Fixes #155.

### DIFF
--- a/swift/Sources/SignalClient/IdentityKey.swift
+++ b/swift/Sources/SignalClient/IdentityKey.swift
@@ -35,7 +35,7 @@ public struct IdentityKeyPair {
         var pubkeyPtr: OpaquePointer?
         var privkeyPtr: OpaquePointer?
         try bytes.withUnsafeBytes {
-            try checkError(signal_identitykeypair_deserialize(&pubkeyPtr, &privkeyPtr, $0.baseAddress?.assumingMemoryBound(to: UInt8.self), $0.count))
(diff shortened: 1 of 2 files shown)


jrose-signal:
Tests forthcoming, oops.

jack-signal (approved):

PR #184, opened 2021-02-05
Avoid using unwrap

jack-signal:
If we must assert that a value is `Ok`/`Some` use `expect` with a string explaining why this must be so.

Most of the changes here are in the tests, where of course `unwrap` is mostly harmless. But updating there as well makes it easy to grep for any other uses.

(This skips `unwrap_or` because those don't panic)


rust/protocol/src/fingerprint.rs
```diff
@@ -136,15 +136,15 @@ impl ScannableFingerprint {
         let same1 = combined
             .local_fingerprint
             .as_ref()
-            .unwrap()
+            .ok_or(SignalProtocolError::FingerprintParsingError)?
```
jrose-signal, line 139: This means you can remove the above `if`. (I like the new form better, honestly.)

rust/protocol/src/identity_key.rs
```diff
@@ -97,7 +97,10 @@ impl IdentityKeyPair {
             private_key: self.private_key.serialize().to_vec(),
         };
         let mut result = Vec::new();
-        structure.encode(&mut result).unwrap();
+
+        // prost documents the only possible encoding error is if there is insufficient
+        // space, which is not a problem when it is allowed to encode into a Vec
+        structure.encode(&mut result).expect("No encoding error");
```
jrose-signal, line 103: danburkert/prost#154 :sob:

rust/protocol/src/protocol.rs
```diff
@@ -625,7 +632,7 @@ mod tests {
     use rand::rngs::OsRng;
     use rand::{CryptoRng, Rng};
 
-    fn create_signal_message<T>(csprng: &mut T) -> SignalMessage
+    fn create_signal_message<T>(csprng: &mut T) -> Result<SignalMessage>
```
jrose-signal, line 635: This one doesn't really make sense to me. It may be a helper function for a test, but it actually cannot fail, and immediately panicking is valid. Can you use `expect` instead?

rust/protocol/src/identity_key.rs
jack-signal: https://github.com/danburkert/prost/pull/378 stalled since October :(

rust/protocol/src/protocol.rs
jack-signal: Sure I guess. Reasoning was to avoid as many `expects` in `src` as possible since those need to be audited for panics, and having tests using unwrap/expect makes it harder to find ones that are in the protocol src proper.

jrose-signal: Eh, my opinion is that auditing is all well and good, but not at the cost of actual type signatures. I can see how being a test helper puts it on the edge, though.

Another option would be to go *back* to use `unwrap` in tests, since the Clippy warning keeps it out of the src files. Any of these three options are all right with me though.

jack-signal:
@jrose-signal ready for another look

rust/bridge/jni/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 //
 
 #![allow(clippy::missing_safety_doc)]
+#![deny(clippy::unwrap_used)]
```
jrose-signal, line 7: This needs to get added to bridge/ffi/src/lib.rs and bridge/node/src/lib.rs as well.

jrose-signal (commented):
We've added poksho to the repository since the original PR. signal-neon-futures is also a separate crate (rust/bridge/node/futures), so we'll need to apply this there as well.

jrose-signal (approved):

PR #185, opened 2021-02-05
Add a script to rewrite Node's package.json

jack-signal:
Currently the `package.json` in https://github.com/signalapp/libsignal-client-node is copied by the script but it uses paths relative to `node` and I hand-hacked it before checkin. Add a script that automates the changes.

### DIFF
--- a/node/scripts/rewrite_package_json.py
+++ b/node/scripts/rewrite_package_json.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2021 Signal Messenger, LLC.
+# SPDX-License-Identifier: AGPL-3.0-only
+#
+
+import sys
+import json
+
+def remove_subdir(l):
+    l = l.replace('node/', '')
+
+    if l == "tsc -b node":
+        l = "tsc"
+
+    return l
+
+def main(args = None):
+    if args is None:
+        args = sys.argv
+
+    if len(args) != 2:
+        raise Exception("Usage: %s <path to package.json>" % (args[0]))
+
+    info = json.loads(open(args[1], 'r').read())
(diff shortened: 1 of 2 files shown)


node/scripts/rewrite_package_json.py
```diff
@@ -0,0 +1,40 @@
...
+#
+
+import sys
+import json
+
+def remove_subdir(l):
+    l = l.replace('node/', '')
+
+    if l == "tsc -b node":
+        l = "tsc"
```
jrose-signal, line 15: There won't be any TypeScript to build, so it shouldn't matter what the shortcut is. Or is it getting run anyway?

Why bother flattening the paths at all? It's going to be embedded inside the app anyway.

jack-signal: TBH logic here is 

I have no #### clue what yarn may or may not do with a `package.json` it pulls from a dependency
With `node/` prefix it obviously doesn't work with respect to the layout in libsignal-client-node
Maybe yarn doesn't look at any of this in which case rewriting it is harmless

jrose-signal: Right, I'm saying that the layout in libsignal-client-node may as well match the layout in libsignal-client. (The "main" entry in the package.json should make that okay; "dist" isn't magic either.)

ThisIsMissEm:
@jack-signal the repo at `https://github.com/signalapp/libsignal-client-node` appears to be archived, so is this still relevant? (I'm just catching up on what's been happening with libsignal since I last looked)

jrose-signal:
One of the things that has happened is that Jack has left Signal! But yes, whoops, this is no longer relevant. Thanks for poking it.

PR #186, opened 2021-02-05
More Node bindings

jack-signal:
~No tests yet~


jrose-signal (approved):
Let's copy Swift's minimal tests at least. :-)

PR #189, opened 2021-02-09
Add SenderKeyStore and related functions for TypeScript

jack-signal:


rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+use signal_neon_futures::*;
+use std::fmt;
+use std::marker::PhantomData;
+use std::panic::AssertUnwindSafe;
+use std::sync::Arc;
 
 pub mod logging;
 
+#[derive(Debug)]
+struct CallbackError {
```
jrose-signal, line 20: This is what signal_neon_futures::PersistentException is for, though admittedly it may need some wrapping or extra traits to really work. (Having to explicitly destroy it might make that unworkable too.)

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+
+fn js_error_to_rust(func: &'static str, err: String) -> SignalProtocolError {
+    SignalProtocolError::ApplicationCallbackError(func, Box::new(CallbackError::new(err)))
+}
+
+pub fn boxed_object<'a, T: 'static + Send>(
+    cx: &mut TaskContext<'a>,
+    value: T,
+) -> JsResult<'a, JsValue> {
+    Ok(cx.boxed(node::DefaultFinalize(value)).upcast())
```
jrose-signal, line 46: This isn't going to be 100% consistent because mutable objects get stuffed in RefCell first. I recommend using ResultTypeInfo's `convert_into` everywhere.

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+        &self,
+        name: SenderKeyName,
+    ) -> Result<Option<SenderKeyRecord>, String> {
+        eprintln!("here...");
+        let store_object_shared = self.store_object.clone();
+        JsFuture::get_promise(&self.js_queue, move |cx| {
+            let store_object = store_object_shared.to_inner(cx);
+            let name: Handle<JsValue> = boxed_object(cx, name.clone())?; // XXX clone
+            eprintln!("now here...");
+            let result = call_method(cx, store_object, "_getSenderKey", std::iter::once(name))?;
```
jrose-signal, line 74: I thought about whether `call_method` should be a macro just to avoid needing to wrap arguments in `once` or `vec!`.

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+
+    async fn do_get_sender_key(
+        &self,
+        name: SenderKeyName,
+    ) -> Result<Option<SenderKeyRecord>, String> {
+        eprintln!("here...");
+        let store_object_shared = self.store_object.clone();
+        JsFuture::get_promise(&self.js_queue, move |cx| {
+            let store_object = store_object_shared.to_inner(cx);
+            let name: Handle<JsValue> = boxed_object(cx, name.clone())?; // XXX clone
```
jrose-signal, line 72: I don't have a good answer for how to avoid *this* clone, since we're handing brand new objects out to JS. I think most of the time it's fine, though: the callbacks all use small-ish objects like addresses and identity keys, or they're doing something like "save session" where we can transfer the object we were working with locally.

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+    }
+
+    async fn do_get_sender_key(
+        &self,
+        name: SenderKeyName,
+    ) -> Result<Option<SenderKeyRecord>, String> {
+        eprintln!("here...");
+        let store_object_shared = self.store_object.clone();
+        JsFuture::get_promise(&self.js_queue, move |cx| {
+            let store_object = store_object_shared.to_inner(cx);
```
jrose-signal, line 71: If you use `into_inner` here, you don't have to `finalize` below (which is good panic protection).

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+#[async_trait(?Send)]
+impl<'a> SenderKeyStore for NodeSenderKeyStore<'a> {
+    async fn load_sender_key(
+        &mut self,
+        sender_key_name: &SenderKeyName,
+        _ctx: libsignal_protocol::Context,
+    ) -> Result<Option<SenderKeyRecord>, SignalProtocolError> {
+        eprintln!("load_sender_key({:?})", sender_key_name);
+        // XXX clone()
+        self.do_get_sender_key(sender_key_name.clone())
```
jrose-signal, line 180: …but we shouldn't need to clone *twice.* (I imagine this was moving things around.)

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+        eprintln!("store_sender_key({:?})", sender_key_name);
+        // XXX clone()
+        self.do_save_sender_key(sender_key_name.clone(), record.clone())
+            .await
+            .map_err(|s| js_error_to_rust("saveSenderKey", s))
+    }
+}
+
+#[allow(non_snake_case)]
+#[doc = "ts: export function SenderKeyDistributionMessage_Create(name: SenderKeyName, store: Object): Promise<SenderKeyDistributionMessage>"]
```
jrose-signal, line 200: You can write this as a normal doc comment (see the logging functions for examples).

jrose-signal (commented):
Some hopefully helpful comments for this WIP.

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,289 @@
...
+pub fn SenderKeyDistributionMessage_Create(mut cx: FunctionContext) -> JsResult<JsObject> {
+    // XXX There must be an easier way than this:
+    let name_arg = cx.argument::<<&SenderKeyName as node::ArgTypeInfo>::ArgType>(0)?;
```
jrose-signal, line 203: One more thought on simplifying this: the type should be inferred by the call to `borrow` on the next line; I only made it explicit in the macro because I wanted less wiggle room in the macro.

rust/bridge/node/src/lib.rs
jack-signal:
```
error[E0507]: cannot move out of an `Arc`
  --> rust/bridge/node/src/lib.rs:70:32
   |
70 |             let store_object = store_object_shared.into_inner(cx);
   |                                ^^^^^^^^^^^^^^^^^^^ move occurs because value has type `libsignal_bridge::node::Root<libsignal_bridge::node::JsObject>`, which does not implement the `Copy` trait
```

?

jrose-signal: Oops. >< My mistake and also something to think about how to simplify later.

node/index.ts
```diff
@@ -909,3 +957,81 @@ export class UnidentifiedSenderMessageContent {
...
+  getSignedPreKey(id: number): SignedPreKeyRecord;
+  saveSignedPreKey(id: number, record: SignedPreKeyRecord): void;
+}
+*/
+
+export abstract class SenderKeyStore {
+  _saveSenderKey(
+    name: SignalClient.SenderKeyName,
+    record: SignalClient.SenderKeyRecord
+  ): Promise<void> {
```
jrose-signal, line 1000: `async` and Promise are interchangeable in JavaScript, so you should be able to write these as `async` functions instead.

node/libsignal_client.d.ts
```diff
@@ -66,11 +68,13 @@ export function SenderCertificate_GetServerCertificate(cert: SenderCertificate):
 export function SenderCertificate_GetSignature(obj: SenderCertificate): Buffer;
 export function SenderCertificate_New(senderUuid: string, senderE164: string | null, senderDeviceId: number, senderKey: PublicKey, expiration: number, signerCert: ServerCertificate, signerKey: PrivateKey): SenderCertificate;
 export function SenderCertificate_Validate(cert: SenderCertificate, key: PublicKey, time: number): boolean;
+export function SenderKeyDistributionMessage_Create(name: SenderKeyName, store: unknown): Promise<SenderKeyDistributionMessage>;
```
jrose-signal, line 71: To avoid this `unknown`, maybe the abstract class should live in libsignal_client.d.ts.in?

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,287 @@
...
+            let store_object = store_object_shared.to_inner(cx);
+            let name: Handle<JsValue> = boxed_object(cx, name)?;
+            let result = call_method(cx, store_object, "_getSenderKey", std::iter::once(name))?;
+            let result = result.downcast_or_throw(cx)?;
+            store_object_shared.finalize(cx);
+            Ok(result)
+        })
+        .then(|cx, result| match result {
+            Ok(value) => {
+                match value.downcast::<JsBox<crate::node::DefaultFinalize<SenderKeyRecord>>, _>(cx)
```
jrose-signal, line 80: I think I made an alias `node::DefaultJsBox` for this.

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,287 @@
...
+        let store_object_shared = self.store_object.clone();
+        JsFuture::get_promise(&self.js_queue, move |cx| {
+            let store_object = store_object_shared.to_inner(cx);
+            let name: Handle<JsValue> = boxed_object(cx, name)?;
+            let record: Handle<JsValue> = boxed_object(cx, record)?;
+            let result = call_method(
+                cx,
+                store_object,
+                "_saveSenderKey",
+                std::iter::once(name).chain(std::iter::once(record)),
```
jrose-signal, line 114: I think `vec!` is fine here; worrying about a small heap allocation when we're going to call into JavaScript is probably overkill.

rust/bridge/node/src/lib.rs
```diff
@@ -3,13 +3,287 @@
...
+            )?
+            .downcast_or_throw(cx)?;
+            store_object_shared.finalize(cx);
+            Ok(result)
+        })
+        .then(|cx, result| match result {
+            Ok(value) => match value.downcast::<JsUndefined, _>(cx) {
+                Ok(_) => Ok(()),
+                Err(_) => Err("name must be a string".into()),
```
jrose-signal, line 123: Wrong error message?

node/index.ts
```diff
@@ -782,10 +803,37 @@ export class SenderCertificate {
...
+  }
+
   private constructor(nativeHandle: SignalClient.SenderKeyDistributionMessage) {
     this.nativeHandle = nativeHandle;
   }
 
+  static async create(
+    name: SenderKeyName,
+    store: SenderKeyStore
+  ): Promise<SenderKeyDistributionMessage> {
```
jrose-signal, line 817: I think we probably *should* hook up the context parameter, even if we make it [optional](https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters), because that'll be groundwork for higher-level transactions like iOS uses.

node/libsignal_client.d.ts
jack-signal: Abstract class has interfaces which refer to the wrapper types, so that doesn't work. What I did instead is split it into an interface with the functions that we invoke in Rust, then the abstract class in index.ts implements that interface in terms of the functions on the abstract class that are operating on the wrapper types.

node/index.ts
```diff
@@ -909,3 +957,44 @@ export class UnidentifiedSenderMessageContent {
...
+    const skr = this.getSenderKey(SenderKeyName._fromNativeHandle(name));
+    if (skr == null) {
+      return null;
+    } else {
+      return skr._unsafeGetNativeHandle();
+    }
+  }
+
+  abstract saveSenderKey(name: SenderKeyName, record: SenderKeyRecord): void;
+  abstract getSenderKey(name: SenderKeyName): SenderKeyRecord | null;
```
jrose-signal, line 983: Whoops, these two should return Promises.

node/index.ts
jack-signal: Makes sense. I'm struggling to see how to approach it though. It feels like `Managed::to_raw` (https://docs.rs/neon/0.7.1/src/neon/handle/mod.rs.html#25) is the right way to go from a `Handle<JsValue>` to a `*mut c_void` which we can stick into the libsignal `Context`. However `Managed::from_raw` requires an `Env` (though it is not actually used for `JsValue`'s from_raw) and I can't find any public API in Neon that returns an `Env`.

Maybe thinking about this wrong?

Would propose since we don't need ctx right now, add it later as an optional when we have everything else sorted.

jrose-signal: Hm, we're going to need to keep it alive somehow, which makes me think we'll need a [Root](https://docs.rs/neon/0.7.1-napi/neon/handle/struct.Root.html) somewhere. We could then do the Swift/C thing and have our low-level Context actually be a reference to the Root.

```rust
let mut root = object.root(&mut cx);
let opaque_context = &mut root as *mut c_void;
```

It's not Rusty at all, though. An alternate approach would be to change the Context type to be `Box<dyn Any>` or `&dyn Any`, which can at least be downcast back safely.

But all right, we can add it later since it's not straightforward.

jrose-signal (approved):

PR #190, opened 2021-02-09
Bridge: Groundwork for async bridge_fns

jrose-signal:


jack-signal (approved):
Cool I kind of started down this path yesterday so I could use `TaskContext` within the promise glad I wasn't totally off base with that.

PR #191, opened 2021-02-10
Log the SignalClient version when the logger is initialized

jack-signal:
Re the discussion on Android/libsignal-client group where the version of the lib being used wasn't clear to the Android team. Just log the version, then there is no ambiguity.

This requires we keep the crate version in sync with the release version but that doesn't seem terribly onerous.

Considered also logging the git hash and commit date using https://docs.rs/vergen/3.2.0/vergen/index.html but probably just a version is sufficient.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -101,6 +101,9 @@ jobs:
     - name: Rustfmt check
       run: cargo fmt --all -- --check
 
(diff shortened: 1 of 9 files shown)


jrose-signal:
Can we add a GitHub check that the crate versions stay in sync with the individual language package versions? (perhaps as a follow-up)

jrose-signal (approved):

PR #192, opened 2021-02-10
Don't promote old sessions

jack-signal:
libsignal-protocol-java does this, SignalProtocolKit does not

AND-1221


jack-signal:
Per Ehren's doc we are doing something different here, closing this.

PR #193, opened 2021-02-10
Bridge: Convert store-based APIs over to async bridge_fns

jrose-signal:
I suggest reviewing commit-by-commit. The store implementations are copied over verbatim from what's already there, so no need to look into their specifics.


rust/bridge/shared/src/node/convert.rs
```diff
@@ -94,6 +126,37 @@ where
...
 
+pub(crate) struct FinalizableOption<T: Finalize>(Option<T>);
+
+impl<T: Finalize> Finalize for FinalizableOption<T> {
+    fn finalize<'a, C: Context<'a>>(self, cx: &mut C) {
+        if let Some(value) = self.0 {
+            value.finalize(cx)
+        }
+    }
+}
```
jrose-signal, line 137: This will be unnecessary in the next version of Neon: neon-bindings/neon#680

rust/bridge/shared/macros/src/lib.rs
```diff
@@ -228,15 +242,132 @@ fn jni_name_from_ident(ident: &Ident) -> String {
...
+
+    quote! {
+        // Use a RefCell so that the early-exit cleanup functions can reference the context
+        // without taking ownership.
+        let cx = std::cell::RefCell::new(cx);
+        #(#input_saving)*
+        #(#input_unwrapping)*
+        Ok(signal_neon_futures::promise(
+            &mut cx.into_inner(),
+            std::panic::AssertUnwindSafe(async move {
```
jrose-signal, line 343: I'm not happy about AssertUnwindSafe, even though we're kind of implicitly relying on it just by working with Neon. I want to come back and see if I can do something safer, but since the stores are all passed `&mut` it might be impossible.

For what it's worth, I put this restriction in the signature of `signal_neon_futures::promise` myself, and I suppose I'm the one taking it back out by putting AssertUnwindSafe into the macro here. :-/

jrose-signal:
Planned follow-up work (unordered, not all high-priority):
- Document the heck out of `bridge_fn` and the infrastructure behind it.
- Add some more organization in libsignal-bridge and libsignal-bridge-macros.
- Spend some time, but not too much, on the AssertUnwindSafe issue above.
- Make it easier to reuse the type conversions here in manually-written FFI/JNI/Neon functions.
- See if there's an easy way to unify `bridge_get_bytearray` with the regular `bridge_get`.
- Try to clean up the Node stores, including being able to avoid clones.
- Persist thrown errors in the Node stores.
- Possibly clean up the Java stores as well.
- I'm sure there's even more but we'll go with that.

jack-signal (approved):

PR #195, opened 2021-02-12
Add node multi-arch/cross-compilation support

dennisameling:
Needed for Apple Silicon/Windows arm64/Windows ia32

Ref https://github.com/signalapp/Signal-Desktop/issues/3745#issuecomment-777837619
Ref https://github.com/signalapp/Signal-Desktop/issues/1636#issuecomment-778153694
Ref https://github.com/signalapp/Signal-Desktop/issues/4461

**How to test**:

`npm_config_arch` is only needed when you're cross-compiling from a x64 host, e.g. x64 to ia32.

Branch with generated binaries: https://github.com/dennisameling/libsignal-client-node/tree/cross-compilation-compat/build

```
export npm_config_arch=ia32
yarn install
yarn tsc
```

Tested on:
- [x] Windows ia32
- [x] Windows x64
- [x] Windows arm64
- [x] MacOS arm64 (Apple Silicon)
- [x] MacOS x64
- [x] Linux x64

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -207,10 +207,10 @@ jobs:
     - run: yarn tsc
 
     - run: yarn lint
-      if: matrix.arch == 'ubuntu-latest'
+      if: matrix.os == 'ubuntu-latest'
 
     - run: yarn format -c
-      if: matrix.arch == 'ubuntu-latest'
+      if: matrix.os == 'ubuntu-latest'
 
     - name: Run yarn test
       uses: GabrielBB/xvfb-action@v1.4
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -172,7 +172,11 @@ jobs:
   node:
     name: Node
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
 
     needs: changes
 
@@ -186,8 +190,14 @@ jobs:
       with:
         profile: minimal
 
-    - run: nvm install && nvm use
-      shell: bash -l {0}
+    - name: Get Node version from .nvmrc
+      id: get-nvm-version
+      shell: bash
+      run: echo "::set-output name=node-version::$(cat .nvmrc)"
+
+    - uses: actions/setup-node@v2
+      with:
+        node-version: ${{ steps.get-nvm-version.outputs.node-version }}
 
     - name: Verify that the Node bindings are up to date
       run: rust/bridge/node/bin/gen_ts_decl.py --verify
@@ -197,8 +207,10 @@ jobs:
     - run: yarn tsc
 
     - run: yarn lint
+      if: matrix.arch == 'ubuntu-latest'
 
     - run: yarn format -c
+      if: matrix.arch == 'ubuntu-latest'
 
     - name: Run yarn test
       uses: GabrielBB/xvfb-action@v1.4
--- a/node/build_node_bridge.py
+++ b/node/build_node_bridge.py
@@ -31,6 +31,10 @@ def main(args=None):
                       help='specify Node OS name')
     parser.add_option('--cargo-build-dir', default='target', metavar='PATH',
                       help='specify cargo build dir (default %default)')
+    parser.add_option('--cargo-target', default=None,
+                      help='specify cargo target')
+    parser.add_option('--node-arch', default=None,
+                      help='specify node arch (x64, ia32, arm64)')
 
     (options, args) = parser.parse_args(args)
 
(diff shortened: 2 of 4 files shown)


node/index.ts
```diff
@@ -7,7 +7,8 @@ import * as os from 'os';
 import bindings = require('bindings'); // eslint-disable-line @typescript-eslint/no-require-imports
 import * as SignalClient from './libsignal_client';
 
-const SC = bindings('libsignal_client_' + os.platform()) as typeof SignalClient;
+const arch = process.env.npm_config_arch || process.arch
```
jrose-signal, line 10: Under what conditions will these be different?

node/build_node_bridge.py
```diff
@@ -49,9 +53,23 @@ def main(args=None):
...
+        return 1
+    if cargo_target.startswith('..\\'):
+        cargo_target = cargo_target[3:]
+
+    node_arch = options.node_arch
+    if node_arch is None:
+        print('ERROR: --node_arch is required')
+        return 1
+    if node_arch.startswith('..\\'):
+        node_arch = node_arch[3:]
```
jrose-signal, line 68: @jack-signal, do you remember what this `..\` check is for? It doesn't seem like any of these would come up in arguments.

.github/workflows/build_and_test.yml
```diff
@@ -201,6 +209,7 @@ jobs:
     - run: yarn format -c
 
     - name: Run yarn test
+      if: matrix.arch == 'x64'
```
jrose-signal, line 212: Do we need this test? (Are you expecting to put combinations in the matrix that won't run?)

jrose-signal (commented):
Thank you! This seems pretty straightforward to me.

node/build_node_bridge.py
jack-signal: It's certainly not supposed to. But it was showing up when invoked via `gyp` on Windows. I think the way `gyp` (or possibly Python) are parsing command line arguments on Windows is wrong, but fixing that was more than I wanted to deal with.

.github/workflows/build_and_test.yml
dennisameling: I was first planning to add a matrix of multiple OSes (ubuntu-latest, windows-latest, macos-latest) and archs (x64, arm64, ia32), but then we can only run tests on x64 since GH Actions is x64-only. It might be good to at least have the Node build task on the different OSes and archs (it would cross-compile to arm64 and ia32 then). Is that OK for you? Then I'm happy to update the workflow accordingly 😊

node/build_node_bridge.py
dennisameling: Yeah I also had it show up on Windows, very weird indeed... I think it's Python's parsing of command line arguments on Windows though, rather than Gyp, IIRC

node/index.ts
dennisameling: Good question. For the build step `npm_config_arch` is definitely needed to support cross-compilation from a x64 host. But since the code here will only run on the target platform (I assume), `process.env.npm_config_arch` should be superfluous here. I'll remove it and only use `process.arch` instead. Thanks!

.github/workflows/build_and_test.yml
jrose-signal: Let's at least add all three OSs Desktop supports today. Cross-compiling without testing adds a little value in that we could be doing something terribly broken on ia32 (less likely for arm64), but I don't think it's enough to be worth it. It's not like we're interfacing with C types when it comes to the Node bridge.

package.json
```diff
@@ -14,7 +14,7 @@
     "clean": "rimraf node/dist build",
     "test": "electron-mocha --recursive node/dist/test",
     "lint": "cd node && eslint . --ext .js,.jsx,.ts,.tsx",
-    "format": "p() { prettier ${@:- --write} package.json '*.js' 'node/**/*.{css,js,json,md,scss,ts,tsx}'; }; p"
+    "format": "bash -c \"p() { prettier ${@:- --write} package.json '*.js' 'node/**/*.{css,js,json,md,scss,ts,tsx}'; }; p\""
```
jrose-signal, line 17: I think we can just skip the `format` and `lint` steps on Windows, or even "not on Linux".

.github/workflows/build_and_test.yml
```diff
@@ -186,8 +190,14 @@ jobs:
       with:
         profile: minimal
 
-    - run: nvm install && nvm use
-      shell: bash -l {0}
+    - name: Get Node version from .nvmrc
+      id: get-nvm-version
+      shell: bash
+      run: echo "::set-output name=node-version::$(cat .nvmrc)"
```
dennisameling, line 196: This is needed because the Windows image doesn't have nvm installed

dennisameling:
Alright, this should be good to go now @jrose-signal. Thanks for the feedback! 🚀

jrose-signal:
Thanks for the contribution, and for working through all the follow-up!

PR #200, opened 2021-02-17
Move the burden of SessionCipher locking to the client.

greyson-signal:
The lock inside of `SessionCipher` is causing a lot of pain due to the threading model of the Android app. Let's just move the entire responsibility of locking to the client.


java/java/src/main/java/org/whispersystems/libsignal/state/SessionLock.java
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (C) 2014-2021 Open Whisper Systems
+ *
+ * Licensed according to the LICENSE file in this repository.
+ */
```
jack-signal, line 5: New headers should be

```
//
// Copyright {\d{4}(-\d{4})?} Signal Messenger, LLC.
//
```

jack-signal (approved):

ahmadalighavami:
sudo ./init-certificate.sh

PR #201, opened 2021-02-18
TS store APIs

jack-signal:
~Not complete -still missing~

- [x] SealedSender_DecryptMessage
- [x] SealedSender_DecryptToUsmc
- [x] SealedSender_EncryptMessage
- [x] SessionCipher_DecryptPreKeySignalMessage

And needs more TS tests.

### DIFF
--- a/node/index.ts
+++ b/node/index.ts
@@ -882,18 +882,6 @@ export class SenderKeyDistributionMessage {
     return new SenderKeyDistributionMessage(handle);
   }
 
-  static async process(
-    name: SenderKeyName,
-    message: SenderKeyDistributionMessage,
-    store: SenderKeyStore
-  ): Promise<void> {
-    await SC.SenderKeyDistributionMessage_Process(
-      name._unsafeGetNativeHandle(),
-      message._unsafeGetNativeHandle(),
-      store
-    );
-  }
-
   static new(
     keyId: number,
     iteration: number,
@@ -931,6 +919,18 @@ export class SenderKeyDistributionMessage {
   id(): number {
     return SC.SenderKeyDistributionMessage_GetId(this.nativeHandle);
   }
+}
+
+export async function processSenderKeyDistributionMessage(
+  name: SenderKeyName,
+  message: SenderKeyDistributionMessage,
+  store: SenderKeyStore
+): Promise<void> {
+  await SC.SenderKeyDistributionMessage_Process(
+    name._unsafeGetNativeHandle(),
+    message._unsafeGetNativeHandle(),
+    store
+  );
 }
 
 export class SenderKeyMessage {
@@ -1180,7 +1180,7 @@ export abstract class SenderKeyStore implements SignalClient.SenderKeyStore {
   abstract getSenderKey(name: SenderKeyName): Promise<SenderKeyRecord | null>;
 }
 
-export async function GroupCipher_Encrypt(
+export async function groupEncrypt(
   name: SenderKeyName,
   store: SenderKeyStore,
   message: Buffer
@@ -1188,7 +1188,7 @@ export async function GroupCipher_Encrypt(
   return SC.GroupCipher_Encrypt(name._unsafeGetNativeHandle(), store, message);
 }
 
-export async function GroupCipher_Decrypt(
+export async function groupDecrypt(
   name: SenderKeyName,
   store: SenderKeyStore,
   message: Buffer
@@ -1248,7 +1248,7 @@ export class CiphertextMessage {
   }
 }
 
-export function ProcessPreKeyBundle(
+export function processPreKeyBundle(
   bundle: PreKeyBundle,
   address: ProtocolAddress,
   sessionStore: SessionStore,
@@ -1262,7 +1262,7 @@ export function ProcessPreKeyBundle(
   );
 }
 
-export async function EncryptMessage(
+export async function signalEncrypt(
   message: Buffer,
   address: ProtocolAddress,
   sessionStore: SessionStore,
@@ -1278,7 +1278,7 @@ export async function EncryptMessage(
   );
 }
 
-export function DecryptSignalMessage(
+export function signalDecrypt(
   message: SignalMessage,
   address: ProtocolAddress,
   sessionStore: SessionStore,
@@ -1292,7 +1292,7 @@ export function DecryptSignalMessage(
   );
 }
 
-export function DecryptPreKeySignalMessage(
+export function signalDecryptPreKey(
   message: PreKeySignalMessage,
   address: ProtocolAddress,
   sessionStore: SessionStore,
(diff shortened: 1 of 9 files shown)


node/index.ts
```diff
@@ -11,7 +11,19 @@ const SC = bindings(
   'libsignal_client_' + os.platform() + '_' + process.arch
 ) as typeof SignalClient;
 
-export const { initLogger, LogLevel, CiphertextMessageType } = SC;
+export const { initLogger, LogLevel } = SC;
+
+export const enum CiphertextMessageType {
```
jrose-signal, line 16: Why move this out here? (especially since it's still in SC, but that's probably an oversight)

node/test/PublicAPITest.ts
```diff
@@ -17,23 +17,135 @@ SignalClient.initLogger(
...
+  }
+
+  async saveIdentity(
+    name: SignalClient.ProtocolAddress,
+    record: SignalClient.PublicKey
+  ): Promise<boolean> {
+    const idx = name.name() + '::' + name.deviceId();
+    const seen = this.idKeys.has(idx);
+    if (!seen) {
+      this.idKeys.set(idx, record);
```
jrose-signal, line 82: This isn't quite correct: it's supposed to return `true` if the identity *changed,* but it's supposed to replace it regardless.

rust/bridge/shared/src/lib.rs
```diff
@@ -744,6 +745,13 @@ bridge_get!(SessionRecord::local_registration_id -> u32);
 bridge_get!(SessionRecord::remote_registration_id -> u32);
 bridge_get!(SessionRecord::has_sender_chain as HasSenderChain -> bool, ffi = false, node = false);
 
+bridge_get!(SealedSenderDecryptionResult::sender_uuid -> String, ffi = false, jni = false);
+bridge_get!(SealedSenderDecryptionResult::sender_e164 -> Option<String>, ffi = false, jni = false);
```
jrose-signal, line 749: You should be able to use `&str` now (to avoid a copy)…

rust/protocol/src/sealed_sender.rs
```diff
@@ -677,6 +677,24 @@ pub struct SealedSenderDecryptionResult {
     pub message: Vec<u8>,
 }
 
+impl SealedSenderDecryptionResult {
+    pub fn sender_uuid(&self) -> Result<&str> {
+        Ok(self.sender_uuid.as_ref())
+    }
+
+    pub fn sender_e164(&self) -> Result<Option<String>> {
+        Ok(self.sender_e164.clone())
```
jrose-signal, line 686: …returning `as_deref()` here.

jrose-signal (commented):
Looks pretty good, only a few comments.

node/index.ts
jack-signal: Was a hack. Initially tried

export const { CiphertextMessageType } = SC;

to re-export from SC but that gives

```
node/index.ts:1248:11 - error TS2749: 'CiphertextMessageType' refers to a value, but is being used as a type here.

1248   type(): CiphertextMessageType {
               ~~~~~~~~~~~~~~~~~~~~~

```

jack-signal: But anyway we don't use CiphertextMessageType within the SC layer, just return the integer value, so IMO best move is just remove it from SC. Thoughts?

rust/bridge/shared/src/lib.rs
jack-signal: A little WTF here, I changed to &str, works great locally, but CI fails

```
error[E0277]: the trait bound `Option<&str>: ffi::convert::ResultTypeInfo` is not satisfied
   --> rust/bridge/shared/src/support/mod.rs:94:13
    |
94  |             #[bridge_fn($($param = $val),*)]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ffi::convert::ResultTypeInfo` is not implemented for `Option<&str>`
    | 
   ::: rust/bridge/shared/src/lib.rs:549:1
    |
549 | bridge_get!(SenderCertificate::sender_e164 -> Option<&str>);
    | ------------------------------------------------------------ in this macro invocation
    |
```

node/index.ts
jrose-signal: [Const enums](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) aren't values, it's true; maybe it'd be `export type CiphertextMessageType = SC.CiphertextMessageType` or something. But at the same time, making this strongly typed is a bit funny regardless, since as you say we mostly care about the integer value, which could be unknown. (The main reason I had it in the .d.ts was so that it could be declared next to the Rust enum with the same constants.)

I'm inclined to say *this* method should *also* return `number` rather than CiphertextMessageType, and not throw on unknown values. That would sidestep the problem too, as long as we can expose the named constants properly.

rust/bridge/shared/src/lib.rs
jrose-signal: Oops, guess I never added that for result types in FFI. It shouldn't be too hard to add, but you don't have to do that now.

node/index.ts
jack-signal: OK changed to return number here, and the app can still compare against the enum values to figure out what they are looking at

rust/protocol/src/identity_key.rs
```diff
@@ -67,6 +67,11 @@ impl IdentityKeyPair {
         }
     }
 
+    pub fn from_private_key(private_key: curve::PrivateKey) -> Result<Self> {
```
jrose-signal, line 70: Should this be a TryFrom implementation?

node/index.ts
```diff
@@ -1003,3 +1195,168 @@ export async function GroupCipher_Decrypt(
...
+  serialize(): Buffer {
+    return SC.CiphertextMessage_Serialize(this.nativeHandle);
+  }
+
+  type(): number {
+    return SC.CiphertextMessage_Type(this.nativeHandle);
+  }
+}
+
+export function ProcessPreKeyBundle(
```
jrose-signal, line 1251: These top-level functions should be lowerCamelCase, yeah? We can use the Swift names for them.

rust/protocol/src/identity_key.rs
jack-signal: It could just be `From` since it can't fail

rust/bridge/node/src/lib.rs
```diff
@@ -40,6 +40,561 @@ fn js_error_to_rust(func: &'static str, err: String) -> SignalProtocolError {
     SignalProtocolError::ApplicationCallbackError(func, Box::new(CallbackError::new(err)))
 }
 
+struct NodePreKeyStore<'a> {
```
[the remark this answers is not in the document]
jack-signal: I had added them in the course of writing SenderKeyStore( and they got copied to all the others). I recall I needed it but don't remember why now. I'll check if removing works

rust/protocol/src/identity_key.rs
jack-signal: There is some other cleanups that should happen in the curve code - I will defer this to a future PR (today) since it doesn't really have anything to do with the TS API

jrose-signal (approved):

PR #203, opened 2021-02-18
Don't stuff long logs into error/exception messages

jack-signal:
Instead use the log integration

(We just had no log integration at the point that the decryption failure logging was needed)

### DIFF
--- a/rust/protocol/src/session_cipher.rs
+++ b/rust/protocol/src/session_cipher.rs
@@ -1,5 +1,5 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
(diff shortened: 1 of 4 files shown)


rust/protocol/src/session_cipher.rs
```diff
@@ -425,8 +425,12 @@ fn decrypt_message_with_record<R: Rng + CryptoRng>(
                 previous_state_count(),
             );
         }
-        Err(SignalProtocolError::MessageDecryptionFailed(
-            create_decryption_failure_log(remote_address, &errs, record, ciphertext)?,
+        log::error!(
+            "{}",
+            create_decryption_failure_log(remote_address, &errs, record, ciphertext)?
```
jrose-signal, line 430: Can we go even further and have `create_decryption_failure_log` just call `log::error!` multiple times? Or is it worth having it all on one line?

jack-signal: I considered that. But the list of possible chains is IMO one "log", not something to be considered individually. OTOH they all will get dumped consecutively into the file anyway so either way is probably fine? IDK, no strong feelings here from me. And I don't think Android cares either way as long as the bits end up in the debug log somehow.

jrose-signal (approved):

PR #204, opened 2021-02-19
CocoaPods: Build the Rust library as part of installing the CocoaPod

jrose-signal:
Reverts #163 in favor of an alternate approach to using cocoapods-binary that allows the Swift code to be built on the app side, avoiding some LLDB/Swift integration issues.


Imperiopolis-Signal (approved):

jack-signal (approved):

charlesmchen-signal (approved):
LGTM seems fine but I don't know much about rust/cargo/etc.

PR #205, opened 2021-02-19
Merge poksho into libsignal-cliennt

jack-signal:


jrose-signal (approved):
This is coming wholesale from the original repo, so I didn't review the code changes, only the addition to the workspace. We should work out the duplicate dependencies but otherwise this is great to see!

jack-signal:
We need to update our fork of curve25519-dalek first in order to get consistent hash trait impls https://github.com/signalapp/curve25519-dalek/pull/1

PR #215, opened 2021-02-20
Fix or mark-as-allowed Clippy lints in tests

jrose-signal:
And tweak the GitHub Clippy enforcement so this doesn't regress.

I do feel better about marking Clippy lints as allowed *in tests* because sometimes the test is trying to test something that you wouldn't write in normal code, or can safely ignore something you wouldn't want to ignore in normal code. I do think it still makes sense to decide that on a test-by-test basis, though.


jack-signal (approved):
Concur re allowing Clippy warnings in tests but only if explicitly allowed

PR #218, opened 2021-02-22
Bridge: split protocol and aes_gcm_siv bindings into separate files

jrose-signal:
Apart from helping avoid the top-level file growing infinitely, this also allows using a dedicated Result type in each file instead of writing `SignalProtocolError` or `aes_gcm_siv::Error` over and over.


rust/bridge/shared/src/ffi/convert.rs
```diff
@@ -225,13 +225,13 @@ macro_rules! ffi_bridge_handle {
         impl ffi::SimpleArgTypeInfo for &$typ {
             type ArgType = *const $typ;
             #[allow(clippy::not_unsafe_ptr_arg_deref)]
-            fn convert_from(foreign: *const $typ) -> Result<Self, ffi::SignalFfiError> {
+            fn convert_from(foreign: *const $typ) -> std::result::Result<Self, ffi::SignalFfiError> {
```
jack-signal, line 228: Maybe add a `SignalFfiResult` type to reduce clutter?

rust/bridge/shared/src/jni/convert.rs
```diff
@@ -318,7 +318,7 @@ macro_rules! jni_bridge_handle {
             fn convert_from(
                 _env: &jni::JNIEnv,
                 foreign: Self::ArgType,
-            ) -> Result<Self, jni::SignalJniError> {
+            ) -> std::result::Result<Self, jni::SignalJniError> {
```
jack-signal, line 321: Similarly `SignalJniResult`

jack-signal (approved):

jack-signal (approved):

PR #222, opened 2021-02-24
Update dependencies

### LINKED ISSUE #221: Bump jni version to 0.19.0 to support building on FreeBSD
Hello, would it be possible to update the jni crate to at least 0.19.0? Version 0.19.0 contains the following patch, which enables jni to be compiled on FreeBSD: https://github.com/jni-rs/jni-rs/commit/54062fca75e22f88f82de058fbaafc77a537356d.

Thanks!

jack-signal:
This is `cargo update` along with manually upgrading jni to 0.19 and prost to 0.7

This isn't a complete cargo update as it avoids updating polyval as there is a duplicate dependency issue there in `cpuid-bool` between the newest version of `polyval` crate and the current latest version of the `sha2` crate.

The bytes dependency in protocol that was causing duplicate dependencies with prost was a red herring - actually `bytes` was not used with the protocol crate!

Supplants #164 update prost
Fixes #221 update jni


jack-signal:
Created a PR with RustCrypto to update `cpuid-bool` so we can also bump the `polyval` crate https://github.com/RustCrypto/hashes/pull/240

jrose-signal:
Oh, I misread the `jni` dependency on `combine` (that we needed for the dependency on `bytes` v1). Sorry, @svmhdvn!

Cargo.lock
```diff
@@ -1098,11 +1057,23 @@ version = "0.7.3"
...
+ "getrandom 0.1.16",
+ "libc",
+ "rand_chacha 0.2.2",
+ "rand_core 0.5.1",
+ "rand_hc 0.2.0",
+]
+
+[[package]]
+name = "rand"
+version = "0.8.3"
```
jrose-signal, line 1069: We should update our use of `rand` as well if we can, even though the second copy is just a build dependency.

jack-signal: Unfortunately that pulls in a new version of `rand_core` which would require some updating of curve25519-dalek to avoid yet another duplicated dep.

svmhdvn:
Looks like this PR takes care of the `prost` depedency right? Feel free to close #164 then!

jrose-signal (approved):

PR #226, opened 2021-02-25
Always clamp private keys on deserialize

jack-signal:
We clamp on generate but some clients will generate non-clamped keys expecting the operation will clamp, and as of now clients who deserialize such keys will produce incorrect results.


jrose-signal (approved):
"some clients" including some tests, whoops. I added that test, feel free to change it to do one more roundtrip or something.

PR #227, opened 2021-02-25
SenderKey: Rename 'keyId' to 'chainId', 'groupId' to 'distributionId'

jrose-signal:
Clarifies the use of "ID" in SenderKey-related APIs. I've left deprecated entry points for Java but not for Swift and TypeScript (which are not in use yet).

- SenderKeyMessage::key_id -> chain_id (avoids double "key" in name)
- SenderKeyDistributionMessage::id -> chain_id (to match SKM)
- SenderKeyName::group_id -> id (it's not the global group ID)


ehrenkret-signal:
SenderKeyName::id would tend to imply that's the sole identifier for a SenderKey... It's really somewhat more like group_id or distribution_id because you need to join it with the sender_id and the sender_device_id to get to a unique identifier.

jrose-signal:
We can go with `distribution_id`. The impetus for this change was to avoid `group_id` since clients have access to the *actual* ID of a group but shouldn't be using it here.

jack-signal (approved):

PR #228, opened 2021-02-25
Node: accept both `null` and `undefined` for optional arguments

jrose-signal:
Return values will continue to use `null` rather than `undefined` for the time being.


jrose-signal:
Hm, this doesn't actually mean much unless we change our .d.ts, which would then lead to the question of which one to use. I guess we should switch everything to `undefined` or leave it as is.

PR #229, opened 2021-02-26
Bridge: Limit bridge_get_bytearray so it can use bridge_fn_buffer

jrose-signal:
By disallowing arbitrary expressions, `bridge_get_bytearray` (and `bridge_get_optional_bytearray`) ends up a lot more like `bridge_get`, and can reuse the infrastructure in `bridge_fn_buffer` instead of needing a separate per-bridge implementation.


jrose-signal:
I considered having `bridge_get` behave like `bridge_get[_optional]_bytearray` when the return type was `[u8]` (not a type you can normally return), but that seemed like too much magic.

jack-signal (approved):

PR #230, opened 2021-03-02
Log panics in all three bridges

jrose-signal:
I went back and forth on including backtraces since it pulls in additional dependencies, but ultimately decided it was worth it. It's rough enough that any panics we have won't appear in crash logs (due to the bridge mechanisms for each app environment).


jack-signal (approved):
Only comment is `log_panics::init` is always called even if the logger has already been initialized. It seems like `std::panic::set_hook`, which is what `::init` uses, can be called more than once safely but there doesn't seem to be a reason to do so.

jrose-signal:
`init_logger` isn't meant to be called more than once either, since it leaks a `Box<dyn Logger>` each time.

jrose-signal:
But I can move it to the success path only just in case.

PR #235, opened 2021-03-05
Sealed sender: allow any known message type as the payload

jrose-signal:
`sealed_sender_decrypt` won't decrypt additional message kinds, but they can be decrypted to UnidentifiedSenderMessageContent and dealt with that way. Test this by adding `sealed_sender_encrypt_from_usmc` and shoving a SenderKeyMessage in an UnidentifiedSenderMessage.


jack-signal (approved):

jrose-signal:
Incorporated into #269.

PR #236, opened 2021-03-05
Device Sender Utilities - Generate Key + Self-Signed Certificate

jack-signal:
This still needs some basic Java+Swift tests which I'll add on Monday.

@cody-signal No need to review this whole thing I just wanted to run the Java API in `DeviceTransfer.java` by you

@jrose-signal I could use a hint on the Swift binding. I'm hitting a type error in generateDeviceTransferCertificate and I can't make it happy, which is why that fn is commented out.


swift/Sources/SignalClient/DeviceTransfer.swift
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright 2020 Signal Messenger, LLC
```
jrose-signal, line 2: 2021

swift/Sources/SignalClient/DeviceTransfer.swift
```diff
@@ -0,0 +1,34 @@
...
+        try invokeFnReturningArray {
+            signal_device_transfer_generate_private_key($0, $1)
+        }
+    }
+}
+
+/*
+public func generateDeviceTransferCertificate<Bytes: ContiguousBytes>(privateKey: Bytes,
+                                                                      name: String,
+                                                                      daysTilExpire: UInt32) -> [UInt32] {
```
jrose-signal, line 20: Return type should be [UInt8]. (That's probably the only problem.)

jrose-signal: Argument type should be Int too—it's conventional in Swift to use Int everywhere unless you need a specific range, even if a number's required to be non-negative. Converting to UInt32 will do a range check and trap.

jrose-signal (commented):
Haven't looked at the actual implementation yet, but the wrappers all look fine. That said, what do you think of a more typesafe design?

```swift
public struct DeviceTransferKey {
  public let bytes: [UInt8]
  public static func generate() -> Self {
    let bytes = ...
    return Self(bytes: bytes)
  }
  public func generateCertificate(name: String, daysTilExpire: Int) -> [UInt8] {
    // …
  }
}
```

Doesn't have to have stronger Rust types backing it, just makes the interface layer a little nicer.

swift/Sources/SignalClient/DeviceTransfer.swift
jack-signal: :joy: Thanks just completely missed this, and the error message wasn't giving any hints.

jack-signal:
I like your proposed binding approach, that prevents someone from giving a mangled key to the cert gen function

cody-signal (commented):
Java API looks good to me. Any chance you could send me a key and cert so I can test them for compatibility? Mainly to sanity check Java will read/parse the binary that these will generate.

cody-signal (approved):
@jack-signal provided key/cert and verified they work with the java ssl code. Looks good from my side.

jrose-signal (commented):
Looks good, other than…

java/java/src/main/java/org/signal/libsignal/devicetransfer/DeviceTransferKey.java
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright 2021 Signal Messenger, LLC.
+// SPDX-License-Identifier: AGPL-3.0-only
+//
+
+package org.signal.libsignal.devicetransfer;
+
+import org.signal.client.internal.Native;
+
+class DeviceTransferKey {
```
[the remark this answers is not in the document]
cody-signal: If this is the intended interface for Android to call, would also need to make a public accessor for the private key.

jrose-signal (approved):

PR #237, opened 2021-03-08
Add a test of hasCurrentState

jack-signal:
(which fails at the moment)


rust/bridge/shared/src/node/storage.rs
```diff
@@ -7,7 +7,7 @@ use super::*;
 
 use async_trait::async_trait;
 use signal_neon_futures::*;
-use std::sync::Arc;
+use std::{cell::RefCell, sync::Arc};
```
jrose-signal, line 10: Oops, I should figure out how to turn this off, or stop taking VS Code's suggestions.

jrose-signal (approved):

PR #240, opened 2021-03-09
Bump Java version to 0.3.4

jack-signal:
@jrose-signal Bumping from 0.2.3->0.4.0 to indicate new support for DeviceTransferKey, with the idea that next Swift build also goes to 0.4.0 (from current 0.3.1) for same reason. Seems ok?

### DIFF
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -1,5 +1,5 @@
 subprojects {
-    ext.version_number     = "0.2.3"
+    ext.version_number     = "0.3.4"
     ext.group_info         = "org.whispersystems"
 
     if (JavaVersion.current().isJava8Compatible()) {
--- a/rust/bridge/jni/Cargo.toml
+++ b/rust/bridge/jni/Cargo.toml
@@ -5,7 +5,7 @@
 
 [package]
 name = "libsignal-jni"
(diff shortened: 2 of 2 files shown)


jrose-signal:
I was thinking there's no breaking change, only new features, so we'd stay on the 0.3 series (per [Cargo's modified SemVer](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements), also [used by yarn](https://classic.yarnpkg.com/en/docs/dependency-versions#toc-caret-ranges)). But since the app teams always manually pick a new version of libsignal-client, it's not like we really have to follow that exactly.

jack-signal:
There is a breaking change between here and 0.2.3 because this picks up the UUID non-optionality changes.

jack-signal:
Oh but Swift already has that channge and is on 0.3, so bumping to 0.3.0 here?

jack-signal:
Using 0.3.1 to match Swift

jack-signal:
OK after further discussion, using 0.3.4 here so that next Node release can go 0.3.3->0.3.4 and Swift can go 0.3.1->0.3.4 and we can maybe have consistent versioning

PR #241, opened 2021-03-10
By default, skip Rust tests that take over 10s in debug builds

jrose-signal:
We'll still run them in CI, but we don't need to run them locally.


jack-signal (approved):
Not sure what is up with the CI failure

PR #243, opened 2021-03-11
Bridge: add ignored Context args to methods taking stores for JNI/Node

jrose-signal:
Previously we defined one entry point for FFI (Swift) that took an extra "context" parameter (to pass through iOS's database transactions), and one for JNI+Node that did not (no context needed currently). But this is all in our glue layer, which doesn't need to be a perfect reflection of the outside interface. Remove that duplication by accepting a Context parameter for both JNI and Node that, for now, must be null.


jack-signal (approved):
Nice simplification

PR #244, opened 2021-03-12
Node: use 'Native' to refer to the loaded Rust library

jrose-signal:
Specifically, use `NativeImpl` to refer to the run-time bindings (instead of `SC`), and `Native` to refer to the TypeScript module for those bindings (instead of `SignalClient`). This makes compile-time diagnostics clearer, since the overall Node package is named `signal-client`.


jrose-signal:
Lint issue fixed in #245.

jack-signal (approved):

PR #246, opened 2021-03-12
Add hashes and HMAC for Java, move AES-GCM-SIV to signal-crypto

jack-signal:
I'm still working on GCM (and there is CTR, CBC, etc yet to happen) but this is self-contained and has all of the bridge-related changes already. I'll do ciphers in a followup pr.


Cargo.toml
```diff
@@ -10,6 +11,7 @@ members = [
 ]
 default-members = [
     "rust/aes-gcm-siv",
+    "rust/crypto",
```
jrose-signal, line 14: Long-term, should `aes-gcm-siv` move into `crypto`? Or is it sufficiently different?

jrose-signal: Alternately, should `crypto` have a more specific name, even if it's exposed generically in Java for backwards-compatibility?

java/java/src/main/java/org/signal/libsignal/crypto/CryptographicHash.java
```diff
@@ -0,0 +1,34 @@
...
+  public CryptographicHash(String algo) {
+    this.handle = Native.CryptographicHash_New(algo);
+  }
+
+  @Override
+  protected void finalize() {
+    Native.CryptographicHash_Destroy(this.handle);
+  }
+
+  public void update(byte[] input, int offset, int len) {
```
jrose-signal, line 22: Makes me appreciate slices all over again in Rust and Swift.

java/tests/src/test/java/org/signal/libsignal/crypto/CryptographicHashTests.java
```diff
@@ -0,0 +1,57 @@
...
+     CryptographicHash hash = new CryptographicHash(algo);
+
+     byte[] input = Hex.fromStringCondensed(hexInput);
+
+     hash.update(input);
+     byte[] digestAllInOne = hash.finish();
+
+     assertEquals(Hex.toHexString(digestAllInOne), hexExpectedOutput);
+
+     if(input.length > 1) {
```
jrose-signal, line 23: Why this check?

rust/crypto/src/hash.rs
```diff
@@ -0,0 +1,80 @@
...
+    HmacSha1(Hmac<Sha1>),
+}
+
+impl CryptographicMac {
+    pub fn new(algo: &str, key: &[u8]) -> Result<Self> {
+        match algo {
+            "HMACSha1" | "HmacSha1" => Ok(Self::HmacSha1(
+                Hmac::<Sha1>::new_varkey(key).expect("HMAC accepts any key length"),
+            )),
+            "HMACSha256" | "HmacSha256" => Ok(Self::HmacSha256(
```
jrose-signal, line 24: I was going to suggest that the string-based generic entry points be implemented in Java, since they're not a design we'd use for any other languages even if we want to expose the operations, but it does come out really compact in Rust with the enum. The alternate approach would probably be to use a trait instead, and have the bridge functions return `Box<dyn CryptographicMac>`. Thoughts?

java/tests/src/test/java/org/signal/libsignal/crypto/CryptographicHashTests.java
jack-signal: I guess not strictly necessary, but without it the test function would fail for an empty input.

rust/crypto/src/hash.rs
jack-signal: I'd rather avoid the extra heap allocation if we can, given how small most of these objects are.

java/tests/src/test/java/org/signal/libsignal/crypto/CryptographicHashTests.java
jrose-signal: :facepalm: Right, my brain saw the condition and flagged it even after I'd looked at what was being tested. My bad.

Cargo.toml
jack-signal: I've moved `aes-gcm-siv` code into `rust/crypto`

rust/crypto/src/lib.rs
```diff
@@ -1,12 +1,23 @@
 //
-// Copyright 2021 Signal Messenger, LLC.
+// Copyright 2020 Signal Messenger, LLC.
```
jrose-signal: Oops!

jrose-signal (approved):

PR #247, opened 2021-03-12
Include distribution ID inside SenderKey[Distribution]Message

jrose-signal:
The distribution ID is used to identify which key a particular sender is using to encrypt their SenderKeyMessage, so it has to be known as part of decryption. The previous design had the distribution ID stored alongside each message (perhaps on the "envelope" structure that's received from the server), but that's harder to keep track of, and it would only be present for certain message kinds anyway.

Includes #227. Also removes the SenderKeyName abstraction, which is no longer pulling its weight in the new design.


rust/bridge/shared/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 
 #![allow(clippy::missing_safety_doc)]
 #![deny(clippy::unwrap_used)]
+#![feature(min_const_generics)]
```
jrose-signal, line 8: This is in stable in the latest rustc, so it can go away when we bump the compiler requirement.

jrose-signal:
@ehrenkret-signal I tagged you to verify the structural changes, which in practice is probably wire.proto and the changes to group_cipher.rs.

rust/protocol/src/uuid.rs
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright 2021 Signal Messenger, LLC.
+// SPDX-License-Identifier: AGPL-3.0-only
+//
+
+use std::array::TryFromSliceError;
+use std::convert::TryFrom;
+
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
+pub struct Uuid([u8; 16]);
```
jrose-signal, line 10: This is very similar to the [`uuid` crate](https://docs.rs/uuid/0.8.2/uuid/), but I wasn't sure if we wanted to pull in another dependency.

java/java/src/main/java/org/whispersystems/libsignal/groups/GroupSessionBuilder.java
```diff
@@ -33,14 +36,14 @@ public GroupSessionBuilder(SenderKeyStore senderKeyStore) {
...
-   * Construct a group session for receiving messages from senderKeyName.
+   * Construct a group session for receiving messages from sender.
    *
-   * @param senderKeyName The (groupId, senderId, deviceId) tuple associated with the SenderKeyDistributionMessage.
+   * @param sender The address of the device that sent the message.
    * @param senderKeyDistributionMessage A received SenderKeyDistributionMessage.
    */
-  public void process(SenderKeyName senderKeyName, SenderKeyDistributionMessage senderKeyDistributionMessage) {
+  public void process(SignalProtocolAddress sender, SenderKeyDistributionMessage senderKeyDistributionMessage) {
     synchronized (GroupCipher.LOCK) {
```
jack-signal, line 45: Unrelated to this PR, but should we remove this lock?

rust/bridge/shared/src/jni/convert.rs
```diff
@@ -188,6 +189,36 @@ impl<'a> SimpleArgTypeInfo<'a> for Option<String> {
...
+        };
+        let lsb = match call_method_checked(env, foreign, "getLeastSignificantBits", "()J", &[])? {
+            JValue::Long(value) => value,
+            anything_else => {
+                return Err(SignalJniError::UnexpectedJniResultType(
+                    "getLeastSignificantBits",
+                    anything_else.type_name(),
+                ))
+            }
+        };
```
jack-signal, line 213: Lots of code duplication here feels like it could be an inner fn or a loop

rust/protocol/src/uuid.rs
jack-signal: This seems simple enough to just use inline without adding a dep, IMO

jack-signal (approved):

java/java/src/main/java/org/whispersystems/libsignal/groups/GroupSessionBuilder.java
jrose-signal: Probably, yes. cc @greyson-signal

rust/bridge/shared/src/jni/convert.rs
jrose-signal: Yeah, `call_method_checked` can probably take care of this! \*goes off to implement\*

jrose-signal: Okay, how do you feel about the latest commit?

jack-signal: Looks great, checking the return type removes a lot of noise.

java/java/src/main/java/org/whispersystems/libsignal/groups/GroupSessionBuilder.java
greyson-signal: Yeah, please no extra locks, we can wrap it in our own locking structure. Just annotate the class as not thread-safe, similar to what I did in `SessionCipher`.

PR #251, opened 2021-03-16
Add AES-256-GCM implementation

jack-signal:
Along with sub-components AES-256-CTR

We can't use https://github.com/RustCrypto/AEADs/tree/master/aes-gcm/src or most other existing libraries because we need to be able do incremental encryption/decryption. [Also RustCrypto still doesn't support the relevant ARMv8 instructions so would be quite slow]

https://tools.ietf.org/html/rfc8452 explains the relationship between GHASH (used in GCM) and POLYVAL (used in GCM-SIV), which probably helps review of that code

The best reference to GCM overall is probably https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf


rust/crypto/src/aes_gcm.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub fn verify_tag(self, tag: &[u8]) -> Result<()> {
+        if tag.len() != TAG_SIZE {
+            return Err(Error::InvalidTag);
+        }
+
+        let computed_tag = self.ghash.finalize()?;
+
+        let tag_ok = tag.ct_eq(&computed_tag);
+
+        if tag_ok.unwrap_u8() == 0 {
```
jrose-signal, line 184: [`tag_ok.into()`](https://doc.dalek.rs/subtle/struct.Choice.html#method.unwrap_u8)?

rust/crypto/src/ghash.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::{Error, Result};
+use std::convert::TryInto;
+
+pub struct Ghash {
+    polyval: Polyval,
+}
+
+fn reverse_bytes_16(c: &[u8]) -> [u8; 16] {
+    assert!(c.len() == 16);
+    u128::from_be_bytes(c.try_into().expect("Verified size")).to_le_bytes()
```
jrose-signal, line 16: Ugh, I'd consider it a Rust optimizer bug that this is faster than the generic version below. Oh well.

rust/crypto/src/ghash.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    pub fn finalize(self) -> Result<[u8; 16]> {
+        Ok(reverse_bytes_16(&self.polyval.finalize()?))
+    }
+}
+
+#[test]
+fn ghash_test() -> Result<()> {
+    // RFC 8452 Appdendix A
```
jrose-signal, line 65: Typo: "Appendix"

jrose-signal (commented):
Non-algorithm comments (i.e. comments I can make without having the RFC open at the same time).

rust/crypto/src/aes_gcm.rs
```diff
@@ -0,0 +1,271 @@
...
+            return Err(Error::InvalidTag);
+        }
+
+        Ok(())
+    }
+}
+
+pub struct Aes256Gcm {
```
jrose-signal, line 192: I don't really understand what this extra type is for. I see that it has setup work, but is that work actually reused in practice? Or will it always be Aes256GcmEncryption / Aes256GcmDecryption that get reused?

rust/crypto/src/ghash.rs
jack-signal: Yeah I was kind of sad but the difference is stark https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=683509d4f2f8f0a78388512626e58ff0

rust/crypto/src/aes_gcm.rs
jack-signal: `Aes256Gcm{Encryption,Decryption}` are maintaining per-message state and are not reused. `Aes256Gcm` is caching the expanded AES key plus the encryption of all-zeros which is used across all GCM messages encrypted with that key.

rust/crypto/src/ghash.rs
jrose-signal: I was going to file a bug report and then thought about how you'd load the bytes *without* memcpy and, ugh, I guess it's a Hard Problem. *However,* you could still use the same *implementation* (with `reverse()`) rather than using byte swapping, and the compiler optimizes it successfully. I think that'd be easier to understand at a quick glance.

rust/crypto/src/aes_gcm.rs
jrose-signal: Right, but are multiple messages ever encrypted with the same key in practice? If not, it doesn't seem like an API we need to provide right now.

jack-signal: Ah, I see what you mean. You're right probably we don't actually need this. Will modify

rust/crypto/src/aes_ctr.rs
```diff
@@ -0,0 +1,77 @@
...
+                for b in 0..PAR_BLOCKS {
+                    assert_eq!(counter.to_be_bytes().len(), 4);
+                    self.ctr[AES_BLOCK_SIZE * b + 12..AES_BLOCK_SIZE * (b + 1)]
+                        .copy_from_slice(&counter.to_be_bytes());
+                    counter = counter.wrapping_add(1);
+                }
+
+                self.pad.copy_from_slice(&self.ctr);
+                self.aes256.encrypt(&mut self.pad[..])?;
+                self.pad_offset = 0;
```
jrose-signal, line 68: It took me a bit to understand that this logic was the same as the setup. Is the setup performance important enough to not factor it out?

rust/crypto/src/aes_gcm.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        let mut final_block = [0u8; 16];
+        final_block[..8].copy_from_slice(&(8 * self.ad_len).to_be_bytes());
+        final_block[8..].copy_from_slice(&(8 * self.msg_len).to_be_bytes());
+
+        self.ghash.update(&final_block)?;
+        let mut hash = self.ghash.finalize()?;
+
+        for (i, b) in hash.iter_mut().enumerate() {
+            *b ^= self.nonce[i];
```
jrose-signal, line 91: It took me a while to realize that this nonce wasn't the same as the original nonce; would `nonce_after_ctr` work instead or would that just cause more confusion / feel non-idiomatic?

jrose-signal (commented):
Algorithm comments! Of course you're running a whole test suite so I didn't expect to find any problems.

rust/crypto/src/aes_ctr.rs
jack-signal: Here we can avoid having to repeatedly copy the nonce (since those fields are unchanged) which saves a bunch of `memcpy`s.

rust/crypto/tests/aes_ctr.rs
```diff
@@ -0,0 +1,58 @@
...
+        let mut buf = vec![0u8; output.len()];
+
+        let mut processed = 0;
+        while processed != buf.len() {
+            let remaining = buf.len() - processed;
+            let this_time = std::cmp::max(1, std::cmp::min(remaining, i));
+            assert!(this_time > 0);
+            aes_ctr.process(&mut buf[processed..processed + this_time])?;
+            processed += this_time;
+        }
```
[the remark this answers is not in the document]
jack-signal: It really is. That's much cleaner.

I originally intended on doing randomized splitting but laziness led to just splitting by `i`-len chunks and if I'm doing that might as well use `chunks_mut`

rust/crypto/src/aes_gcm.rs
jack-signal: Something like `ghash_pad` is probably better as  `pad` is a somewhat common notation for this value in the general context of Carter-Wegman MACs (of which GCM's authentication function is an example).

rust/crypto/src/aes_ctr.rs
jack-signal: However pulling the counter increment logic out to a fn doesn't seem to impact perf much at all (at least as measured for overall GCM) so I did that

rust/crypto/tests/aes_ctr.rs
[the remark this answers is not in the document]
jack-signal: I changed to randomized splitting instead as it's more likely to flush out any weird buffering corner case bugs.

rust/crypto/tests/aes_ctr.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    for _ in 0..32 {
+        // Do it again but with split inputs:
+        let mut aes_ctr = signal_crypto::Aes256Ctr32::from_key(&key, &nonce, init_ctr)?;
+        let mut buf = vec![0u8; output.len()];
+
+        let mut processed = 0;
+        while processed != buf.len() {
+            let remaining = buf.len() - processed;
+            let this_time = std::cmp::max(1, rng.gen::<usize>() % remaining);
```
jrose-signal, line 52: `rng.get_range(1..=remaining)`

jack-signal: Little more complicated than that because we are on `rand 0.7` still

rust/crypto/tests/aes_ctr.rs
```diff
@@ -49,7 +49,7 @@ fn aes_ctr_long_test() -> Result<(), signal_crypto::Error> {
         let mut processed = 0;
         while processed != buf.len() {
             let remaining = buf.len() - processed;
-            let this_time = std::cmp::max(1, rng.gen::<usize>() % remaining);
+            let this_time = if remaining > 1 { rng.gen_range(1, remaining) } else { remaining };
```
jrose-signal: 👍 I didn't realize the range-based `gen_range` was from a newer `rand` than what we're using.

jrose-signal (approved):

PR #255, opened 2021-03-17
Fix bridge_handle! logic for disabling for certain targets

jack-signal:
It only matched exactly `$typ as false` so for example `$typ as false, mut = true` would cause the type to still be emitted.

There is probably a better way to resolve this, but this hack fixes my immediate issue

### DIFF
--- a/rust/bridge/shared/src/ffi/convert.rs
+++ b/rust/bridge/shared/src/ffi/convert.rs
@@ -333,6 +333,7 @@ impl crate::support::Env for Env {
 /// Implementation of [`bridge_handle`](crate::support::bridge_handle) for FFI.
 macro_rules! ffi_bridge_handle {
     ( $typ:ty as false ) => {};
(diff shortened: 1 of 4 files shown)


rust/bridge/shared/src/ffi/convert.rs
```diff
@@ -333,6 +333,7 @@ impl crate::support::Env for Env {
 /// Implementation of [`bridge_handle`](crate::support::bridge_handle) for FFI.
 macro_rules! ffi_bridge_handle {
     ( $typ:ty as false ) => {};
+    ( $typ:ty as false, clone = true ) => {};
```
jrose-signal, line 336: The `$(, clone = $_:tt)?` from the bottom case is probably the most general approach, or even `$(, $($_:tt)*)` ("any remaining tokens after a comma").

jrose-signal (approved):
Oooops, thanks for catching this.

PR #256, opened 2021-03-17
Bridge: add support for strongly-typed Node errors

jrose-signal:
- After loading the Rust library into Node, set an 'Errors' property with the relevant error types on the module object.

- Whenever a `bridge_fn` produces an error, pass it to a new `SignalNodeError::throw` API along with the `this` object, which is assumed to have to be the object with the 'Errors' property.

This is a little less tidy than how we do Java exceptions, but it comes from not having access to the error classes by some kind of absolute name. Alternate approaches considered include:

- Use an initialized-once global. Downside: would not work if you ever had more than one Node engine live in a process, or quit and restarted it. Upside: available everywhere.

- Store the errors on the global object under some long, complicated key (like "org.signal.libsignal-client.Errors"). Downside: pollutes the global object.

- Generate the classes using Neon instead of writing them in TypeScript. Downsides: inconvenient, difficult to maintain, harder to use *from* TypeScript.

### DIFF
--- a/node/test/PublicAPITest.ts
+++ b/node/test/PublicAPITest.ts
@@ -857,6 +857,127 @@ describe('SignalClient', () => {
       }
     });
 
+    it('rejects self-sent messages', async () => {
+      const sharedKeys = new InMemoryIdentityKeyStore();
+
+      const aSess = new InMemorySessionStore();
+      const bSess = new InMemorySessionStore();
+
+      const bPreK = new InMemoryPreKeyStore();
+      const bSPreK = new InMemorySignedPreKeyStore();
+
+      const bPreKey = SignalClient.PrivateKey.generate();
+      const bSPreKey = SignalClient.PrivateKey.generate();
+
+      const sharedIdentityKey = await sharedKeys.getIdentityKey();
+
+      const aE164 = '+14151111111';
+
+      const sharedDeviceId = 1;
+
+      const sharedUuid = '9d0652a3-dcc3-4d11-975f-74d61598733f';
+
+      const trustRoot = SignalClient.PrivateKey.generate();
+      const serverKey = SignalClient.PrivateKey.generate();
+
+      const serverCert = SignalClient.ServerCertificate.new(
+        1,
+        serverKey.getPublicKey(),
+        trustRoot
+      );
+
+      const expires = 1605722925;
+      const senderCert = SignalClient.SenderCertificate.new(
+        sharedUuid,
+        aE164,
+        sharedDeviceId,
+        sharedIdentityKey.getPublicKey(),
+        expires,
+        serverCert,
+        serverKey
+      );
+
+      const sharedRegistrationId = await sharedKeys.getLocalRegistrationId();
+      const bPreKeyId = 31337;
+      const bSignedPreKeyId = 22;
+
+      const bSignedPreKeySig = sharedIdentityKey.sign(
+        bSPreKey.getPublicKey().serialize()
+      );
+
+      const bPreKeyBundle = SignalClient.PreKeyBundle.new(
+        sharedRegistrationId,
+        sharedDeviceId,
+        bPreKeyId,
+        bPreKey.getPublicKey(),
+        bSignedPreKeyId,
+        bSPreKey.getPublicKey(),
+        bSignedPreKeySig,
+        sharedIdentityKey.getPublicKey()
+      );
+
+      const bPreKeyRecord = SignalClient.PreKeyRecord.new(
+        bPreKeyId,
+        bPreKey.getPublicKey(),
+        bPreKey
+      );
+      bPreK.savePreKey(bPreKeyId, bPreKeyRecord);
+
+      const bSPreKeyRecord = SignalClient.SignedPreKeyRecord.new(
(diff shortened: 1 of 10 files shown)


jrose-signal:
@scottnonnenberg-signal @indutny-signal Any better ideas?

indutny-signal:
I think this is a fine solution.

An alternative could be to follow [Node.js footsteps](https://github.com/nodejs/node/blob/3b3b02f8bbf99e2a746eb264ef4a0f864346a1a5/src/api/exceptions.cc#L89-L149) and just set `.errno` property on the error. That being said, I think your solution looks nicer than this from the user API perspective.

@scottnonnenberg-signal has used signal-client much much more extensively than me, though. He's opinion on this matter should be of more value.

indutny-signal:
To elaborate a bit more. What's potentially good about using `.errno` or `.code` field is that it is more extensible and you could just use TypeScript enums for the values. Let's say that users will have to handle some other errors in a special way. Creating a class for each kind of error might be an overkill, but using the `.code` field would be just as convenient.

jrose-signal:
> To elaborate a bit more. What's potentially good about using `.errno` or `.code` field is that it is more extensible and you could just use TypeScript enums for the values. Let's say that users will have to handle some other errors in a special way. Creating a class for each kind of error might be an overkill, but using the `.code` field would be just as convenient.

It feels weird to me to do that in cases where there will be extra info (in this case the `addr` field of UntrustedIdentityError is exposed), but I suppose it's possible. Let's see what Scott thinks.

jack-signal (approved):
Change itself lgtm. Class per error does seem excessive, but we do it for Java (and it could probably be generated). I'm not sure if the exact error matters 99% of the time though - I think the logic is more like, you tried an operation, it failed, you want to log what happened, but beyond that the exact nature of the error doesn't affect handling.

jrose-signal:
> Class per error does seem excessive, but we do it for Java (and it could probably be generated). I'm not sure if the exact error matters 99% of the time though - I think the logic is more like, you tried an operation, it failed, you want to log what happened, but beyond that the exact nature of the error doesn't affect handling.

Yeah, I wasn't planning to do a class for *every* error, only the ones that might be interesting to handle specifically. These are the two that have come up in prior discussions. (I think we can cut down our Swift errors too.)

scottnonnenberg-signal:
Separate error classes are honestly a bit of a pain in the world of Javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#conditional_catch-blocks

However, we can make either one work:`errno` property or separate classes.

Also, returning `null` from the sealed sender decrypt was a bit cryptic, so it's good to get something more informative. Though I do recall that self-sends are somewhat frequent, so it's not ideal that we're throwing an error for that case. Errors are for exceptional cases, right?

jrose-signal:
> Also, returning `null` from the sealed sender decrypt was a bit cryptic, so it's good to get something more informative. Though I do recall that self-sends are somewhat frequent, so it's not ideal that we're throwing an error for that case. Errors are for exceptional cases, right?

Sealed sender messages aren't used for messages sent between one account's devices (on the grounds that it's not hiding anything anyway, especially given the traffic patterns). But this is specifically about sending a sealed sender message from a device to *itself,* which feels like it should never happen anyway? "It was like this when I got here."

> Separate error classes are honestly a bit of a pain in the world of Javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#conditional_catch-blocks

Ah, I didn't remember that JavaScript `catch` doesn't automatically do an `instanceof` check like Java does. Still, *some* form of this will be relevant for fixing `sealedSenderDecrypt` in failure cases, where we have both an error message and a known sender. If you could have any form for that (including "errors should be union returns rather than thrown exceptions"), what would it be?

scottnonnenberg-signal:
> If you could have any form for that (including "errors should be union returns rather than thrown exceptions"), what would it be?

I think I'd prefer `errno` given the choice.

jrose-signal:
`errno` and what else, for the additional information? (e.g. `senderUuid`)

scottnonnenberg-signal:
Yup, field names that make sense for the data provided. Nothing complicated there.

node/test/PublicAPITest.ts
```diff
@@ -857,6 +857,127 @@ describe('SignalClient', () => {
...
+
+      const aCiphertext = await SignalClient.sealedSenderEncryptMessage(
+        aPlaintext,
+        sharedAddress,
+        senderCert,
+        aSess,
+        sharedKeys
+      );
+
+      try {
```
indutny-signal, line 956: You could use `assert.rejects` for that

indutny-signal (approved):
LGTM with a suggestion, but I'll defer to @scottnonnenberg-signal for feedback.

node/test/PublicAPITest.ts
jrose-signal: I had that originally but it seemed more awkward when it came to poking at the error in multiple ways.

scottnonnenberg-signal (approved):
I look forward to getting a little bit more information from thrown errors!

PR #257, opened 2021-03-18
Format all Java source files

jack-signal:
Using https://github.com/google/google-java-format v1.9

I've been using this (inconsistently) on new files, but had not applied it to the old stuff imported from libsignal-protocol-java.

As with `cargo fmt` its certainly debatable if this actually improves the readability. But there is something to be said for consistency, and every other language used has formatting lints enabled.

If seems worth doing I'll also also look at CI integration


java/java/src/main/java/org/whispersystems/libsignal/util/ByteUtil.java
```diff
@@ -77,24 +79,22 @@ public static int highBitsToInt(byte value) {
...
     return bytes;
   }
 
   public static long byteArray5ToLong(byte[] bytes, int offset) {
-    return
-        ((bytes[offset]     & 0xffL) << 32) |
-        ((bytes[offset + 1] & 0xffL) << 24) |
-        ((bytes[offset + 2] & 0xffL) << 16) |
-        ((bytes[offset + 3] & 0xffL) << 8) |
-        ((bytes[offset + 4] & 0xffL));
```
jrose-signal, line 97: If I were going to grumble about specific formatting decisions, this one seems pretty unfortunate.

jrose-signal:
(still need to do this but will take another approach)

PR #258, opened 2021-03-19
Support offset/length params for GCM encryption in Java API

jack-signal:
As Java has no concept of slices and we need some way of doing partial updates, since this is required to implement the JCE Cipher interface, and Android uses it.

### DIFF
--- a/java/java/src/main/java/org/signal/internal/Native.java
+++ b/java/java/src/main/java/org/signal/internal/Native.java
@@ -72,16 +72,19 @@ public final class Native {
   public static native void Aes256Ctr32_Destroy(long handle);
   public static native long Aes256Ctr32_New(byte[] key, byte[] nonce, int initialCtr);
   public static native byte[] Aes256Ctr32_Process(long ctr, byte[] data);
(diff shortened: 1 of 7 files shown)


java/java/src/main/java/org/signal/internal/Native.java
```diff
@@ -72,16 +72,19 @@ private Native() {}
   public static native void Aes256Ctr32_Destroy(long handle);
   public static native long Aes256Ctr32_New(byte[] key, byte[] nonce, int initialCtr);
   public static native byte[] Aes256Ctr32_Process(long ctr, byte[] data);
+  public static native byte[] Aes256Ctr32_ProcessWithOffset(long ctr, byte[] data, int offset, int length);
```
jrose-signal, line 75: Tempted to say the glue layer should *only* have the offset/length APIs and not bother with the whole-array one.

jack-signal: Yeah that makes sense. TS might also find that convenient, if/when we ever need to expose these for Desktop.

jrose-signal (approved):

PR #263, opened 2021-03-26
Update README, add FUNDING.yml, add SECURITY.md

jrose-signal:
Improve our landing page and our GitHub metadata.


jrose-signal:
@jlund-signal already looked over this; will merge when the tests pass.

PR #264, opened 2021-03-29
Start removing nightly features from support libraries

jrose-signal:
In the hopes that they'll be more reusable that way. The remaining feature flags, `wake_trait` and `min_const_generics`, were both stabilized recently, so when we roll our compiler version forward we'll be able to eliminate them too.

Note that this is *not* going to get this repository off of nightly: the `crypto` crate still uses the `stdsimd` and `aarch64_target_feature` features on Android and iOS for access to intrinsics, and cbindgen always needs *some* nightly installed in order to support macros. But signal-neon-futures and libsignal-bridge-macros are potentially reusable elsewhere.

### DIFF
--- a/rust/bridge/node/futures/src/future.rs
+++ b/rust/bridge/node/futures/src/future.rs
@@ -18,15 +18,14 @@ use crate::util::call_method;
 mod builder;
 pub use builder::JsFutureBuilder;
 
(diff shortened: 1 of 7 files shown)


jrose-signal:
Easier to review ignoring whitespace changes: https://github.com/signalapp/libsignal-client/pull/264/files?w=1

jim-signal (approved):

PR #271, opened 2021-04-06
GitHub: add a workflow to publish a tag to NPM

jrose-signal:
And some other changes in support of this. This is an alternate release strategy to the separate "artifact repo" of [libsignal-client-node](https://github.com/signalapp/libsignal-client-node), using the same upload-to-release Action as [zkgroup](https://github.com/signalapp/zkgroup).


.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,113 @@
+name: Upload Release Artifacts
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag'
+        required: true
```
jrose-signal, line 8: Right now I have this as a manually-run workflow, but we could change it to happen automatically for tags of the form `v*`, like zkgroup does.

jim-signal (approved):

jrose-signal:
After talking to the Desktop folks, directly referencing a tgz seems like it's non-idiomatic for Node/Yarn. I'll be looking into alternate publishing options, and possibly dialing this back to just producing the three *.node libraries.

jrose-signal:
New approach uses `prebuildify` to support publishing binary artifacts directly to NPM.

peter-signal (approved):

PR #272, opened 2021-04-13
[CocoaPod] Properly quote path to libsignal_ffi.a in build script

jrose-signal:
Also, if the library somehow isn't built when used as a non-local CocoaPods dependency, don't suggest running swift/build_ffi.sh; CocoaPods will have removed it already. Re-running `pod install` is more likely to help, though in practice something's probably gone wrong in the configuration.


Imperiopolis-Signal (approved):
LGTM, thank you!

PR #273, opened 2021-04-16
signal-neon-futures: Add a Node benchmark for a single round trip

jrose-signal:
It turns out this takes less than a millisecond, at least when the event loop is empty, but the setup might still be useful in the future. Run with `cargo bench -p signal-neon-futures`.

### DIFF
--- a/rust/bridge/node/futures/tests/node-tests/bench.js
+++ b/rust/bridge/node/futures/tests/node-tests/bench.js
@@ -1,0 +1,24 @@
+//
+// Copyright 2021 Signal Messenger, LLC.
+// SPDX-License-Identifier: AGPL-3.0-only
+//
+
+/* eslint-env es2017 */
+
+const Benchmark = require('benchmark');
+
+const Native = require(process.env.SIGNAL_NEON_FUTURES_TEST_LIB);
+
+const suite = new Benchmark.Suite();
+suite
+  .add('await Native.incrementCallbackPromise(async () => 7)', {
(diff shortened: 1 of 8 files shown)


rust/bridge/node/futures/tests/node-tests/bench.js
```diff
@@ -0,0 +1,28 @@
...
+    },
+  })
+  .on('complete', function() {
+    this.forEach(test => {
+      console.log(
+        '"' + test.name + '": ' + JSON.stringify(test.times, null, 2)
+      );
+    });
+  })
+  .run();
```
indutny-signal, line 28:
```suggestion
  .on('cycle', function(event) {
    console.log(String(event.target));
  })
  .run();
```

Looks a bit nicer:

```
await Native.incrementCallbackPromise(async () => 7) x 29,782 ops/sec ±3.62% (74 runs sampled)
```

indutny-signal (approved):
LGTM with a suggestion.

rust/bridge/node/futures/tests/node-tests/bench.js
jrose-signal: Yeah, much better! Their docs are, uh, not wonderful.

indutny-signal: Yeah, I was trying to look up what `defer: true` meant and found it in their tests :joy:

PR #275, opened 2021-04-20
signal-neon-futures: Replace JsFutureState with a one-shot channel

jrose-signal:
There's no need to be rolling our own state machine here; not using an off-the-shelf channel from the `futures` crate is leftover from the design where the Rust future was synchronously awoken.

### DIFF
--- a/rust/bridge/node/futures/src/future.rs
+++ b/rust/bridge/node/futures/src/future.rs
@@ -3,59 +3,20 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+use futures::channel::oneshot;
 use neon::prelude::*;
 use std::future::Future;
 use std::marker::PhantomData;
-use std::mem;
-use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe};
+use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe, UnwindSafe};
 use std::pin::Pin;
-use std::sync::{Arc, Mutex, Weak};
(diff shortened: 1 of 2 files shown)


rust/bridge/node/futures/src/future.rs
```diff
@@ -65,100 +26,106 @@ impl<T> JsFutureState<T> {
...
+    transform: Box<
+        dyn for<'a> FnOnce(&mut FunctionContext<'a>, JsPromiseResult<'a>) -> T + 'static + Send,
+    >,
+}
+
+/// Represents a shared reference to a [FutureSettler].
+///
+/// Exactly one callback (from JavaScript) will be able to actually settle the future.
+// FIXME: `Mutex<Option<X>>` is heavy for "a value that can atomically only be consumed once".
+type FutureSettlerRef<T> = Arc<Mutex<Option<FutureSettler<T>>>>;
```
jrose-signal, line 70: I found [atomic_take](https://docs.rs/atomic-take/1.0.0/atomic_take/) which exactly what I want, but I'm not sure it's worth pulling in another crate for.

jrose-signal:
Gonna hold off on this for now; @indutny-signal has done some performance investigation and the cost of scheduling things on the Node event queue instead of executing them synchronously might be enough to push us back to some kind of secret handshake on wake ("if the future's executor is the EventQueue I'm running on, wake it synchronously" or something). That'll mean the promise fulfiller will need more information than just "send here".

PR #278, opened 2021-04-25
rust/crypto: bump libc to support getauxval on musl

svmhdvn:
See https://github.com/rust-lang/libc/commit/2eaa621a00a992bfda290457329a49c9a2847711, included in 0.2.93.


jrose-signal:
Thanks! Can you do a build too, so the Cargo.lock gets updated?

svmhdvn:
Done

PR #279, opened 2021-04-27
Make sure SignalProtocolError is Sync

gferon:
When using wrapper types like [`anyhow::Error`](https://docs.rs/anyhow/1.0.25/anyhow/struct.Error.html), it is required that all underlying error types are `Send`, `Sync`, and `'static`. I tried to compile the other crates (with all features) after this change, and it looks like it doesn't break anything.

### DIFF
--- a/rust/protocol/src/error.rs
+++ b/rust/protocol/src/error.rs
@@ -56,7 +56,10 @@ pub enum SignalProtocolError {
     InvalidMessage(&'static str),
     InternalError(&'static str),
     FfiBindingError(String),
-    ApplicationCallbackError(&'static str, Box<dyn Error + Send + UnwindSafe + 'static>),
+    ApplicationCallbackError(
+        &'static str,
+        Box<dyn Error + Send + Sync + UnwindSafe + 'static>,
+    ),
 
     InvalidSealedSenderMessage(String),
     UnknownSealedSenderVersion(u8),


jrose-signal:
Hm, I'm not sure this case ought to be `Sync`. It represents the error produced from a callback in the client application, and while that error has to be Send (because we use async callbacks in the Node bindings) requiring it to be Sync could be limiting. Then again, most errors are Sync in practice, including our current use of Java and JavaScript exceptions.

Further thoughts?

rubdos:
Something else we thought about was to make the `Callback` enum field a compile-time option. The reason being: it's the only thing holding back the whole enum to be `Sync`, and in Rust context the Callback is never used. I voted against that, because it unnecessarily complicates the build for the official Signal clients.

I wonder what kind of non-`Sync` error we could encounter, and if and how any of the compilers would complain about this.

jrose-signal:
> Something else we thought about was to make the `Callback` enum field a compile-time option. The reason being: it's the only thing holding back the whole enum to be `Sync`, and in Rust context the Callback is never used. I voted against that, because it unnecessarily complicates the build for the official Signal clients.

You *can* still end up with ApplicationCallbackErrors in a Rust context: if you had an error, say, looking up a session in the database, you'd want to abort the database operation. (Arguably that means our Store requirements should be returning `Result<_, Box<dyn Error + …>>` or `Result<_, anyhow::Error>` rather than `Result<_, SignalProtocolError>`, since SignalProtocolErrors ought to only be produced by the libsignal-protocol library. But we're fuzzy on that anyway today for validation purposes, such as rejecting a null result.)

rubdos:
Some arguments for the proposed patch:

- `rg ApplicationCallbackError -A3 -B1` returns `CallbackError`s and `ThrownException`, but you made that argument already.
- I would argue that making the Rust part as Rusty as possible is already something to persue an sich.
- Having a basic structure like an Error type `!Sync` is a bit weird, if I may say so. An error type sounds like something that should be a POD, copying it around and moving along threads is a pretty basic thing.
- I would argue that any future `!Sync` error should be made `Sync` by rendering it to a `String`.

An argument against my last point:
- I imagine there are `!Sync` errors, e.g. coming from an obscure FFI, which you could handle by having the original type (for example, an error that refers to a C++ object, on which you want to call a method to dismiss/handle/parse/... the error).

I do hope that I never encounter a `!Sync` error in my life though. @jrose-signal, can you come up with `!Sync` errors, especially `Send + !Sync`? Admittedly, I have never really known `Send` and `Sync` that well.

jrose-signal:
I guess the only `Send + !Sync` errors I could come up with are very contrived; there's no reason for an error to have interior mutability, for example. Non-`Send` is much more plausible (especially allowing non-`'static`) but we're already past that. Okay, let's take this; thanks for the discussion!

rubdos:
Thank you for having it, and thank you for having us think twice! :-)

PR #282, opened 2021-04-29
Node: Make a dedicated error code for DuplicatedMessage

jrose-signal:
This was MessageCounterError in libtextsecure; it's an "error" in that it interrupts processing of an individual message, but that message is then discarded in practice.

### DIFF
--- a/node/Errors.ts
+++ b/node/Errors.ts
@@ -5,8 +5,10 @@
 
 export enum ErrorCode {
   Generic,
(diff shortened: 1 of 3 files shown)


jrose-signal:
@scottnonnenberg-signal, you asked for this a while back (before #256 went in). The error message won't change so your workaround will continue to work, but in the future you'll be able to match explicitly as well.

jim-signal (approved):

Issue #285, opened 2021-05-01
tracking issue for some Rust API learnability improvements


cosmicexplorer:
### Goal: Add docstrings and determine appropriate visibility specifiers for the Rust API
#152 describes a lack of API documentation. I found this to be a blocker when looking to depend on this crate from my own (also AGPLv3) rust project. There has clearly been a lot of extremely thoughtful work modularizing the `jni`, `node`, and `ffi` backends (that part was awesome!), and this is demonstrated in the much higher prevalence of docstrings in the `libsignal_bridge` crate than in the `libsignal_protocol` crate (compare the compiled docsites I've listed near the bottom of this post). I would like to help fix this.

I plan to do this mostly by adding docstrings to modules, structs, etc. Docstrings can be converted into a fantastic searchable webpage for API docs with [`cargo doc`](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html). To get an idea of what I'm thinking of, this is a github pages static site of the most recent docs build I have performed: https://cosmicexplorer.github.io/libsignal-client/target/doc/libsignal_protocol/.

#### Additional orthogonal changes
However, in my perusal, I have also found a couple other types of changes I think will improve the API's learnability:
1. Add (documented!) traits for standard behavior across structs. There are currently several methods spelled slightly differently and implemented as `pub fn` methods directly on structs, e.g. https://github.com/signalapp/libsignal-client/blob/87f205e80cc21ce0b1aa1a0e81f1a291d62fba49/rust/protocol/src/protocol.rs#L121-L123
1. Prefer to return slices with a static size over arbitrary-sized `&[u8]` and `Box<[u8]>` -- this helps keep e.g. secret key data on the stack (vs `Box<[u8]>`) and avoids having to handle any lifetimes (vs `&[u8]`). See https://github.com/signalapp/libsignal-client/blob/87f205e80cc21ce0b1aa1a0e81f1a291d62fba49/rust/protocol/src/identity_key.rs#L30-L32

*This issue previously described a desire to modify module paths as well in service of generated documentation, but that was discussed and decided against here: https://github.com/signalapp/libsignal-client/pull/286#discussion_r625977154.*

### Progress
I'm currently working on this branch: https://github.com/cosmicexplorer/libsignal-client/tree/docs. At the time of writing, the diff contains 15 sequential commits, spanning a net `+2492/-1671`: https://github.com/signalapp/libsignal-client/compare/master...cosmicexplorer:docs?expand=1.

[...]

PR #286, opened 2021-05-01
Add docstrings for a few small modules!

cosmicexplorer:
### Problem
This is the first in a set of planned PRs to add inline documentation to the Rust API which can then be beautifully published with `cargo doc`. See #285.

**I am not a Signal engineer. I am making my best guesses as to the appropriate docstring content. I think I've been able to do just fine, but please push back if I'm mischaracterizing anything -- it's easy for me to just revert anything that seems possibly too bikesheddable.**

### Solution
1. Add module docs for the `libsignal_protocol` crate.
1. Add docs to the `device_transfer` crate.
1. Add docs to the `address` module.
### Result
We will have established a rapport for further improvements to the documentation and module layout as per #285!

### DIFF
--- a/rust/protocol/src/address.rs
+++ b/rust/protocol/src/address.rs
@@ -1,26 +1,67 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+#![warn(missing_docs)]
+
+//! A normalized representation of an individual Signal client instance.
+
+#[cfg(doc)]
+use crate::SignalMessage;
+
 use std::fmt;
 
+/// The type used in memory to represent a *device*, i.e. a particular Signal client instance which
+/// represents some user.
+///
+/// Used in [ProtocolAddress].
+pub type DeviceId = u32;
+
+/// Represents a unique Signal client instance as `(<user ID>, <device ID>)` pair.
 #[derive(Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
 pub struct ProtocolAddress {
     name: String,
-    device_id: u32,
+    device_id: DeviceId,
 }
 
 impl ProtocolAddress {
-    pub fn new(name: String, device_id: u32) -> Self {
+    /// Create a new address.
+    ///
+    /// - `name` defines a user's public identity, and therefore must be globally unique to that
+    /// user.
+    /// - Each Signal client instance then has its own `device_id`, which must be unique among
+    ///   all clients for that user.
+    ///
+    ///```
+    /// use libsignal_protocol::{DeviceId, ProtocolAddress};
+    ///
+    /// // This is a unique id for some user, typically a UUID.
+    /// let user_id: String = "04899A85-4C9E-44CC-8428-A02AB69335F1".to_string();
+    /// // Each client instance representing that user has a unique device id.
+    /// let device_id: DeviceId = 2_u32.into();
+    /// let address = ProtocolAddress::new(user_id.clone(), device_id);
+    ///
+    /// assert!(address.name() == &user_id);
+    /// assert!(address.device_id() == device_id);
+    ///```
+    pub fn new(name: String, device_id: DeviceId) -> Self {
         ProtocolAddress { name, device_id }
     }
 
+    /// A unique identifier for the target user. This is usually a UUID.
+    #[inline]
     pub fn name(&self) -> &str {
         &self.name
     }
 
-    pub fn device_id(&self) -> u32 {
+    /// An identifier representing a particular Signal client instance to send to.
+    ///
+    /// For example, if a user has set up Signal on both their phone and laptop, any [SignalMessage]
+    /// sent to the user will still only go to a single device. So when a user sends a message to
+    /// another user at all, they're actually sending a message to *every* device.
+    #[inline]
+    pub fn device_id(&self) -> DeviceId {
         self.device_id
     }
 }
--- a/rust/protocol/src/lib.rs
+++ b/rust/protocol/src/lib.rs
@@ -3,8 +3,24 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Rust implementation of the **[Signal Protocol]** for asynchronous
+//! forward-secret public-key cryptography.
+//!
+//! In particular, this library implements operations conforming to the following specifications:
+//! - the **[X3DH]** key agreement protocol,
+//! - the **[Double Ratchet]** *(Axolotl)* messaging protocol,
+//!
+//! [Signal Protocol]: https://signal.org/
+//! [X3DH]: https://signal.org/docs/specifications/x3dh/
+//! [Double Ratchet]: https://signal.org/docs/specifications/doubleratchet/
+
 #![warn(clippy::unwrap_used)]
 #![deny(unsafe_code)]
+
+// TODO(https://github.com/signalapp/libsignal-client/issues/285): it should be an aspiration to
+// eventually warn and then error for public members without docstrings. Also see
+// https://doc.rust-lang.org/rustdoc/what-to-include.html for background.
+// #![warn(missing_docs)]
 
 mod address;
 mod consts;
@@ -29,7 +45,7 @@ mod utils;
 use error::Result;
 
 pub use {
-    address::ProtocolAddress,
+    address::{DeviceId, ProtocolAddress},
     curve::{KeyPair, PrivateKey, PublicKey},
     error::SignalProtocolError,
     fingerprint::{DisplayableFingerprint, Fingerprint, ScannableFingerprint},
--- a/rust/device-transfer/src/lib.rs
+++ b/rust/device-transfer/src/lib.rs
@@ -3,7 +3,10 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Support logic for Signal's device-to-device transfer feature.
+
 #![deny(unsafe_code)]
+#![warn(missing_docs)]
 
 use chrono::{Datelike, Duration, Utc};
 use picky::key::PrivateKey;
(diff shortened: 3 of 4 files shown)


rust/protocol/src/address.rs
```diff
@@ -1,26 +1,57 @@
...
+#[allow(unused_imports)]
+use crate::sealed_sender::{self, UnidentifiedSenderMessageContent};
+
 use std::fmt;
 
+/// The type used in memory to represent a *device*, i.e. a particular Signal client instance which
+/// represents some user.
+///
+/// Used in [ProtocolAddress].
+pub type DeviceId = u32;
```
jrose-signal, line 18: Thank you for this; it definitely needs to propagate further!

rust/protocol/src/address.rs
```diff
@@ -1,26 +1,57 @@
...
+    /// user.
+    /// - Each Signal client instance then has its own `device_id`, which must be unique among
+    ///   all clients for that user.
+    pub fn new(name: String, device_id: DeviceId) -> Self {
         ProtocolAddress { name, device_id }
     }
 
+    /// A unique identifier for the target user.
+    ///
+    /// In the Signal app, this is currently a phone number.
```
jrose-signal, line 40: This is not true; it's usually a UUID (and that may be enforced at some point in the future). ProtocolAddress still accepts a free-form string so that it works with phone numbers, but that's mostly just to support tests. All the actual app logic should be using UUIDs.

rust/protocol/src/lib.rs
```diff
@@ -3,28 +3,41 @@
...
+pub mod proto;
+pub mod protocol;
+pub mod ratchet;
+pub mod sealed_sender;
+pub mod sender_keys;
+pub mod session;
+pub mod session_cipher;
+pub mod state;
+pub mod storage;
+pub mod utils;
```
jrose-signal, line 40: I know you mentioned that this is one of your goals in #285, but even if we do decide to publicize the module structure of the library, it won't be by making every single submodule public. (And it might not be with these names either.) We should think about the organization that makes sense for the types that really do make sense to export (and possibly talk about it too, either in #285 or on https://community.signalusers.org).

rust/protocol/src/address.rs
```diff
@@ -1,26 +1,57 @@
...
     }
 
-    pub fn device_id(&self) -> u32 {
+    /// An identifier representing a particular Signal client instance to send to.
+    ///
+    /// For example, if a user has set up Signal on both their phone and laptop, any
+    /// [UnidentifiedSenderMessageContent] sent to the user will only go to a single
+    /// device. Handling this is part of the [Sesame] algorithm.
+    ///
+    /// [Sesame]: https://signal.org/docs/specifications/sesame/#assumptions
```
jrose-signal, line 52: I misread this at first as "if a user sends a message on Signal, it goes to exactly one of the recipient's devices", as opposed to "a single Signal message goes to a particular device, so when a user sends a message they're actually sending a message to *every* device", which I think means the example is potentially more confusing than helpful. It's not really Sesame-related either—that's everything to do with "pre-keys" rather than multiple devices per user.

rust/protocol/src/address.rs
```diff
@@ -1,26 +1,57 @@
...
+
 use std::fmt;
 
+/// The type used in memory to represent a *device*, i.e. a particular Signal client instance which
+/// represents some user.
+///
+/// Used in [ProtocolAddress].
+pub type DeviceId = u32;
+
+/// The target of an [UnidentifiedSenderMessageContent].
```
jrose-signal, line 20: The core type here is SignalMessage, not UnidentifiedSenderMessageContent, but even then ProtocolAddress is used for purposes beyond that. (For example, it's used when encrypting the new multi-recipient sealed sender messages, even though the USMC of such a message is the same for each recipient.) I think it'd be better to just say that ProtocolAddress represents a unique Signal client instance as a (user ID, device ID) pair.

jrose-signal (commented):
Okay, here we go! Some initial comments for this effort, which I suppose is a monotonic improvement over no documentation. :-) Thanks for stepping up!

rust/protocol/src/address.rs
cosmicexplorer: Normally when I write Rust code like this, I actually avoid type aliases and instead use wrapper structs and `From` impls. Everything can still be `Copy`, but I think you get a form of type safety when you do this, e.g.:
```rust
#[derive(Copy, Clone)]
pub struct DeviceId(u32);
impl From<u32> for DeviceId { ... }
impl From<DeviceId> for u32 { ... }
```

Would you accept that sort of change in a followup PR (to focus this one on just doc fixes)? Normally I would expect to see a type alias used as here when using `#[repr(C)]` or other things where you want a very clear struct layout. In the implementation of this API (essentially everywhere except the `rust/bridge/` crate), I think we can make the effort to use explicit wrapper structs with explicit `From` impls.

However, if that sort of larger change is not a priority right now, this type alias is a good intermediate solution I think.

rust/protocol/src/address.rs
cosmicexplorer: AWESOME!!!!!!!!! might tweet about this

rust/protocol/src/lib.rs
cosmicexplorer: That's a great amendment to the goals I wrote in #285 and I will link this comment there and follow up in https://community.signalusers.org/! I will avoid changing any of the module structure right now except to make the `address` module `pub` -- does that sound reasonable?

rust/protocol/src/address.rs
cosmicexplorer: Thank you for this thorough feedback! I will try to correct this.

rust/protocol/src/address.rs
cosmicexplorer: This is super helpful, thanks!

rust/protocol/src/address.rs
jrose-signal: I was thinking about that too. It can't propagate all the way into all three app languages because Java and TypeScript don't have zero-cost wrapper types, but that's no reason not to do it in Rust, so sure, I'd take a follow-up for that.

rust/protocol/src/lib.rs
jrose-signal: Sorry, but I'd really rather keep the various kinds of changes separate. Right now we have a flat structure for our public API, so *any* changes indicate a new design. Therefore rather than say "okay, we definitely want this module but I'm not sure about the others", I'd rather look at some possible organizations all together (flat, today's groupings, proposed future groupings) and pick one. (And similarly I think it would make sense to audit for `pub(crate)` first, at least on types, so that we know what the set of actually-public types is.)

cosmicexplorer: Ok, that's very easy to follow! I think that `cargo doc` actually has flags to build things regardless of `pub(crate)` status, which I'm realizing is the only motivation to mess with the module structure at all. Thank you for the prompt and complete response!

rust/protocol/src/address.rs
cosmicexplorer: Awesome, thanks!!

rust/protocol/src/address.rs
cosmicexplorer: https://twitter.com/hipsterelectron/status/1389639068862930944

Be-ing:
Thanks for working on this! GitHub is showing a warning about the email you've used for these commits:
image

cosmicexplorer:
@Be-ing thanks! I have fixed the issue.

rust/protocol/src/address.rs
cosmicexplorer: Broke this out into #289!

rust/protocol/src/address.rs
```diff
@@ -1,26 +1,74 @@
...
+    ///   all clients for that user.
+    ///
+    ///```
+    /// use libsignal_protocol::{DeviceId, ProtocolAddress};
+    /// # use rand::{Rng, rngs::OsRng};
+    /// # use uuid::Uuid;
+    ///
+    /// // This is a unique id for some user, typically a UUID.
+    /// let user_id: String;
+    /// # user_id = Uuid::new_v4().to_hyphenated().to_string();
```
jrose-signal, line 43: Nitpick: given that it's a docstring, you could just hardcode a UUID instead of generating one.

rust/protocol/src/address.rs
```diff
@@ -1,26 +1,74 @@
...
+    /// assert!(address.name() == &user_id);
+    /// assert!(address.device_id() == device_id.into());
+    ///```
+    pub fn new(name: String, device_id: DeviceId) -> Self {
         ProtocolAddress { name, device_id }
     }
 
+    /// A unique identifier for the target user.
+    ///
+    /// In the Signal app, this is usually a UUID, although that may change in the future.
```
jrose-signal, line 59: This codebase is likely to get stricter on this rather than looser, either by validating that the name is a UUID or perhaps even parsing it into the cheaper `uuid::Uuid` bytes form. It's probably better not to speculate about the Signal app at all here, and just stop at "this is usually a UUID".

jrose-signal (commented):
Getting there!

rust/protocol/src/address.rs
cosmicexplorer: Done!

rust/protocol/src/address.rs
cosmicexplorer: Thanks! That's easy enough to follow.

rust/protocol/src/lib.rs
```diff
@@ -3,9 +3,27 @@
...
+//!
+//! In particular, this library implements operations conforming to the following specifications:
+//! - the **[X3DH]** key agreement protocol,
+//! - the **[Double Ratchet]** *(Axolotl)* messaging protocol,
+//! - the **[Sesame]** session agreement protocol.
+//!
+//! [Signal Protocol]: https://signal.org/
+//! [X3DH]: https://signal.org/docs/specifications/x3dh/
+//! [Double Ratchet]: https://signal.org/docs/specifications/doubleratchet/
+//! [Sesame]: https://signal.org/docs/specifications/sesame/
```
jrose-signal, line 17: I am still trying to come up with a simpler explanation of Sesame than what's in that spec, but it's about linking multiple devices to a single user, something that's not implemented in this library. We can drop it from the doc comment, at least for now.

cosmicexplorer: Thanks! That makes sense! I think I was confused at first mostly by the dizzying array of different IDs being handed around (which is addressed in #289) -- I assumed that something more complex had to be going on.

jrose-signal (approved):

cosmicexplorer:
Added the generated swift ffi changes!

Issue #152, opened 2021-01-21
documentation


Be-ing:
There currently isn't any API usage documentation. The README.md still says "Work in progress. Subject to change without notice, use outside Signal not yet recommended." so maybe it's not worthwhile to invest the effort in thorough documentation yet? Nevertheless at least a minimal getting started example for each language would be helpful.

jack-signal:
At this point the library is not sufficient for actually developing a complete Signal client - a great deal of protocol logic is still embedded in the individual clients. In the future that will hopefully change and we'll consider our options for documentation and examples.

The Java binding closes matches libsignal-protocol-java and so that libs documentation may give some hints as to usage.

Be-ing:
Thank you for clarifying. It looks like the original [libsignal-protocol-c](https://github.com/signalapp/libsignal-protocol-c) is not very actively developed anymore so it wouldn't be a great idea to start a new client using that now. But this new library isn't quite ready yet either. Am I understanding that right?

jack-signal:
I believe libsignal-protocol-c was originally written for Whatsapp. It's possible FB is still using it but I don't know. Signal is not using it in any of the clients and nobody here spends any time on it.

libsignal-protocol-java is relatively well tested, but, in recent releases the Android client has switched to using the Rust library instead. So it is unlikely that libsignal-protocol-java will see any further development. That said the low level protocol details encapsulated in the libsignal-protocol-{c,java} libraries do not change much.

If you want to build a complete Signal client, of the options available I think your best bet is this experimental, completely undocumented, work in progress, unsupported library. Unlike the C and Java libraries we are planning to expand this to cover a great deal of additional necessary client logic, and while we can't really offer much assistance to outside users, we'll at least respond to bug reports and can consider merging PRs, which is far more than you will get from the C or Java libs.

[Do note that this library is AGPLv3, while the libsignal-protocol-{c,java} libraries are GPLv3]

Be-ing:
Great, thanks for clarifying again. I personally don't have the capacity to lead development of a new client (I already have [one really complicated cross platform application](https://mixxx.org/) to maintain). Because the Electron client maintainers consider feature parity with the Android and iOS client [out of scope](https://github.com/signalapp/Signal-Desktop/issues/2383#issuecomment-743533977), there is [interest](https://source.puri.sm/Librem5/chatty/-/issues/300) in starting a new native client for Linux. It seems that using the C FFI of this new library would be the most maintainable way to implement that.

[...]

PR #140, opened 2021-01-11
Use `thiserror` for error type

Spaceface16518:
Details of implementation in commit message and commit comments.

# Library

[`thiserror`](https://lib.rs/crates/thiserror) is an error handling helper library that uses procedural macros to derive `std::error::Error` and `Display`, as well as any needed `From` impls.

# Benefits

This approach is beneficial because error descriptions and variable interpolations are located directly next to enum variant definitions, making the code more readable and understandable.

# Limitations

This cleaner code comes at the price of compile time. On my machine (`AMD Ryzen 7 4700U with Radeon Graphics (8) @ 2.000GHz`, 8GB memory), this branch takes about 4 seconds longer to compile than the master branch.

| Command | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| `master` | 21.120 ± 0.263 | 20.750 | 21.630 |
| `thiserror` | 25.156 ± 0.306 | 24.454 | 25.576 |

# Licensing

`thiserror` [uses the MIT License](https://github.com/dtolnay/thiserror/blob/master/LICENSE-MIT), so it should be compatible with this repository's license. It can also licensed under Apache, if that's easier. I'm not super familiar with licensing, so I'm not sure if this is a challenge.

PR #287, opened 2021-05-02
Add docstrings for 7 more Rust API modules

cosmicexplorer:
This is the second in a set of planned PRs to add docstrings to the Rust API, see #285.

I took care to make these changes highly orthogonal, and have split each module's changes into its own commit.

Notably, this does *not* depend on #286!


rust/protocol/src/curve.rs
```diff
@@ -15,8 +19,10 @@ use arrayref::array_ref;
 use rand::{CryptoRng, Rng};
 use subtle::ConstantTimeEq;
 
+/// TODO: describe the intent of this enum!
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub enum KeyType {
+    /// TODO: What does this case mean?
     Djb,
 }
```
jrose-signal, line 27: Currently the only type of key Signal supports is [djb](https://en.wikipedia.org/wiki/Daniel_J._Bernstein) Curve25519 keys, but in theory we could move to another key type that supports signatures and Diffie-Helman agreements in the future (probably another [elliptic curve](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography), but that's not inherently necessary).

jrose-signal: …I'd be happy to rename "DJB" to "Curve25519" throughout, though, because they really are used synonymously here.

jrose-signal: Note that this makes it incorrect to expose `PUBLIC_KEY_LENGTH` and `PRIVATE_KEY_LENGTH`; they can be `CURVE25519_PUBLIC_KEY_LENGTH` and `CURVE25519_PRIVATE_KEY_LENGTH` where applicable (or `curve25519::PUBLIC_KEY_LENGTH` and `curve25519::PRIVATE_KEY_LENGTH`), but that's not everywhere in these APIs.

rust/protocol/src/curve.rs
```diff
@@ -256,6 +269,7 @@ impl PrivateKey {
         }
     }
 
+    /// Calculate a new key agreed between this private key and the public key `their_key`.
```
jrose-signal, line 272: The result of this operation is usually known as a "shared secret" rather than a "key".

rust/protocol/src/curve.rs
```diff
@@ -299,13 +315,15 @@ impl KeyPair {
...
 
+    /// Instantiate an identity from a known public/private key pair.
     pub fn new(public_key: PublicKey, private_key: PrivateKey) -> Self {
         Self {
             public_key,
             private_key,
         }
     }
 
+    /// Instantiate an identity from byte strings for public and private keys.
```
jrose-signal, line 326: I'd specifically use the word "serialized" here rather than just "byte strings", since public keys are "serialized" with the leading "type" byte and that's what's expected here.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,59 @@
...
+///
+/// [IdentityKeyStore::is_trusted_identity] uses this to ensure the identity provided is configured
+/// for the appropriate role.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum Direction {
     Sending,
     Receiving,
 }
 
+/// Interface defining the TOFU identity store, which may be in-memory, on-disk, etc.
```
jrose-signal, line 35: It doesn't *have* to be TOFU, even though Signal's use of it is.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,59 @@
...
 #[async_trait(?Send)]
 pub trait IdentityKeyStore {
+    /// Return the single specific identity each store is assumed to represent, with private key.
```
jrose-signal, line 38: "the store" rather than "each store"

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,59 @@
...
     async fn get_identity_key_pair(&self, ctx: Context) -> Result<IdentityKeyPair>;
 
+    /// Return a [u32] specific to this store instance.
+    ///
+    /// This numerical id should be randomly generated each time the store is instantiated.
```
jrose-signal, line 43: This is not correct; the local registration ID is a per-device identifier (what you've called "client instance" elsewhere) and should not change run to run.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,59 @@
...
     async fn get_local_registration_id(&self, ctx: Context) -> Result<u32>;
 
+    /// Record an identity into the store. The identity is then considered "trusted".
+    ///
+    /// ah TODO: what does `Ok(false)` represent here?
```
jrose-signal, line 48: Good question, and definitely points to the need for docs. The return value represents whether an existing identity was replaced; if it is new or hasn't changed, the return is `Ok(false)`.

rust/protocol/src/storage/traits.rs
```diff
@@ -62,6 +85,7 @@ pub trait PreKeyStore {
     async fn remove_pre_key(&mut self, prekey_id: PreKeyId, ctx: Context) -> Result<()>;
 }
 
+/// Interface for storing signed pre-keys downloaded from a server. (TODO: ???)
```
jrose-signal, line 88: I'll come back with an explanation of pre-keys and signed pre-keys later; for now feel free to leave this generic.

rust/protocol/src/storage/traits.rs
```diff
@@ -78,6 +102,7 @@ pub trait SignedPreKeyStore {
     ) -> Result<()>;
 }
 
+/// Interface for storing a session associated with the chat with a particular separate user.
```
jrose-signal, line 105: I'd be careful about "user" vs. "device" here (or "client instance" if you prefer). "Users" are pretty much not interesting in libsignal-protocol; everything is device-to-device.

rust/protocol/src/storage/traits.rs
```diff
@@ -94,8 +119,11 @@ pub trait SessionStore {
     ) -> Result<()>;
 }
 
+/// Interface for storing a sender key for other users, allowing multiple keys per user.
```
jrose-signal, line 122: Nitpick: "sender keys for other users"

rust/protocol/src/identity_key.rs
```diff
@@ -51,20 +55,23 @@ impl From<PublicKey> for IdentityKey {
...
 impl IdentityKeyPair {
+    /// Create a new key pair from a public `identity_key` and a private `private_key`.
     pub fn new(identity_key: IdentityKey, private_key: PrivateKey) -> Self {
         Self {
             identity_key,
             private_key,
         }
     }
 
+    /// Generate an unguessable new identity from randomness in `csprng`.
```
jrose-signal, line 74: "unguessable" is probably taking things too far. How about "cryptographically random"?

jrose-signal (commented):
Many comments, apologies for the onslaught. :-)

jrose-signal:
cc @rubdos and @be-ing, also interested in using libsignal-client as a Rust library directly.

rubdos:
cc @gferon and @boxdot, thanks for the heads up @jrose-signal.

@cosmicexplorer You may be interested in the libraries and work at github.com/whisperfish. Feel free to join our Matrix channel to coordinate some other work; we are planning to submit some PRs into here wrt. some crypto code. Cfr also #152.

jrose-signal:
> @jrose-signal would you appreciate a review from me and Gabriel on #286 and this one? I'm not sure whether we can catch any more than you already did, but I'd rather ask whether it'd be appreciated or not.

I think it's reasonable for multiple perspectives, but I don't want to stress out @cosmicexplorer either. @cosmicexplorer, what do you think?

rust/protocol/src/crypto.rs
```diff
@@ -13,25 +15,34 @@ use hmac::{Hmac, Mac, NewMac};
...
+/// The required length of the key provided to [aes_256_ctr_encrypt].
+pub const AES_INPUT_SIZE: usize = 32;
+
+/// The size of the generated nonce used in [aes_256_ctr_encrypt].
+pub const AES_NONCE_SIZE: usize = 16;
 
-    let zero_nonce = [0u8; 16];
+/// Encrypt `ptext` using `key` with AES-256 in CTR mode.
+///
+/// `key` is expanded to 256 bits.
```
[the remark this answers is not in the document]
cosmicexplorer: Yes! Totally right!

rust/protocol/src/crypto.rs
```diff
@@ -13,25 +15,34 @@ use hmac::{Hmac, Mac, NewMac};
...
     let mut ctext = ptext.to_vec();
     cipher.apply_keystream(&mut ctext);
-    Ok(ctext)
+    ctext
 }
 
-pub fn aes_256_ctr_decrypt(ctext: &[u8], key: &[u8]) -> Result<Vec<u8>> {
+/// Decrypt `ptext` using `key` with AES-256 in CTR mode.
+///
+/// `key` is expanded to 256 bits. Note that this method just delegates to [aes_256_ctr_encrypt].
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you!

rust/protocol/src/crypto.rs
```diff
@@ -61,45 +75,62 @@ pub fn aes_256_cbc_decrypt(ctext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
...
 }
 
+/// Length of the MAC key used for [aes256_ctr_hmacsha256_encrypt] and
+/// [aes256_ctr_hmacsha256_decrypt].
+pub const MAC_KEY_LENGTH: usize = 10;
+
+/// Concatenate the results of [aes_256_ctr_encrypt] with `cipher_key` and [hmac_sha256] with
+/// `mac_key` over the message `msg`.
+///
+/// Only the first [MAC_KEY_LENGTH] bytes of the MAC key are used.
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you!

rust/protocol/src/crypto.rs
```diff
@@ -61,45 +75,62 @@ pub fn aes_256_cbc_decrypt(ctext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
...
-    result.extend_from_slice(&mac[..10]);
-    Ok(result)
+    result.extend_from_slice(&mac[..MAC_KEY_LENGTH]);
+    result
 }
 
+/// Validate that the MAC code from `mac_key` agrees with the ciphertext `ctext`, then decrypt with
+/// `cipher_key` using [aes_256_ctr_decrypt].
+///
+/// `ctext` is assumed to contain exactly [MAC_KEY_LENGTH] key bytes at the end.
```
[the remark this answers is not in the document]
cosmicexplorer: I completely agree about the use of "assumed"! I have also taken your advice for the rest of this docstring!

rust/protocol/src/crypto.rs
```diff
@@ -13,25 +15,34 @@ use hmac::{Hmac, Mac, NewMac};
...
+/// Decrypt `ptext` using `key` with AES-256 in CTR mode.
+///
+/// `key` is expanded to 256 bits. Note that this method just delegates to [aes_256_ctr_encrypt].
+pub fn aes_256_ctr_decrypt(ctext: &[u8], key: &[u8; AES_INPUT_SIZE]) -> Vec<u8> {
     aes_256_ctr_encrypt(ctext, key)
 }
 
+/// Encrypt `ptext` using `key` and `iv` with AES-256 in CBC mode.
+///
+/// `key` is expanded to 256 bits.
```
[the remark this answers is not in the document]
cosmicexplorer: I didn't realize that the `32` number was related to `AES-256`! I have clarified this in all the docstrings -- thank you for making this more clear.

rust/protocol/src/consts.rs
```diff
@@ -1,10 +1,47 @@
...
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-pub const MAX_FORWARD_JUMPS: usize = 25_000;
-pub const MAX_MESSAGE_KEYS: usize = 2000;
-pub const MAX_RECEIVER_CHAINS: usize = 5;
-pub const ARCHIVED_STATES_MAX_LENGTH: usize = 40;
-pub const MAX_SENDER_KEY_STATES: usize = 5;
+//! Magic numbers.
```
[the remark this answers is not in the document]
cosmicexplorer: Ok! I have marked both of `crate::consts::{types,limits}` as `pub(crate)` in 2ff52da. Does that address your concern? I think that the tiny little module structure here is useful for documentation of these values. If you're not seeing it, we can totally do away with that as well though!

cosmicexplorer:
I'd be wholly unstressed for people to give this a more thorough review!

rust/protocol/src/identity_key.rs
```diff
@@ -11,12 +13,14 @@ use std::convert::TryFrom;
...
 use prost::Message;
 
+/// Wrapper for [PublicKey].
 #[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Copy)]
 pub struct IdentityKey {
     public_key: PublicKey,
 }
 
 impl IdentityKey {
+    /// Create a new instance.
```
Be-ing, line 23: This doesn't add any information that isn't obvious from the function signature. I suggest to either remove this and similar comments or add more information.

rust/protocol/src/consts.rs
```diff
@@ -1,10 +1,56 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
```
Be-ing, line 2: I'm not sure if someone from outside Signal Messenger, LLC should be changing this, but yes it should be updated.

rust/protocol/src/crypto.rs
```diff
@@ -136,7 +167,7 @@ mod test {
             .expect("valid hex");
         let ptext = [0u8; 35];
 
-        let ctext = super::aes_256_ctr_encrypt(&ptext, &key)?;
+        let ctext = super::aes_256_ctr_encrypt(&ptext, array_ref![&key, 0, AES_INPUT_SIZE]);
```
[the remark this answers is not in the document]
cosmicexplorer: I didn't know `.try_into()` would work here!

rust/protocol/src/crypto.rs
```diff
@@ -61,45 +75,62 @@ pub fn aes_256_cbc_decrypt(ctext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
         .map_err(|_| SignalProtocolError::InvalidCiphertext)?)
 }
 
-pub fn hmac_sha256(key: &[u8], input: &[u8]) -> Result<[u8; 32]> {
+/// The statically-known size of the output of [hmac_sha256].
+pub const HMAC_OUTPUT_SIZE: usize = 32;
+
+/// Calculate the HMAC-SHA256 code over `input` using `key`.
+pub fn hmac_sha256(key: &[u8], input: &[u8]) -> [u8; HMAC_OUTPUT_SIZE] {
```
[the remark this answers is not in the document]
cosmicexplorer: That sounds great! I've made them all use `Result` for now. I do think it is very possible to perform that separate pass later and it will probably be useful at that point.

rust/protocol/src/consts.rs
jrose-signal: I think because of the [CLA](https://signal.org/cla/) it will become correct upon merge, but good thought.

Be-ing:
Thanks for working on this! And yes please join us in the Whisperfish Matrix room at #whisperfish:rubdos.be. I'm not sure what your goal is with using this library, but we're working towards a GUI client written in Rust and QML with the [qmetaobject-rs crate](https://github.com/woboq/qmetaobject-rs/). Currently Whisperfish uses Jolla's proprietary QML libraries so it only works on Sailfish OS, but I just got a PinePhone so I'll be working on [adapting it](https://gitlab.com/groups/whisperfish/-/epics/5) to also work with KDE's Kirigami QML libraries without Sailfish OS. At the moment Whisperfish only supports text messages and GroupsV2 but we intend to make it feature complete. I presume you're probably writing a Signal client, in which case you'll probably be interested in the [libsignal-service-rs](https://github.com/whisperfish/libsignal-service-rs) library which now uses this libsignal-client library. @gferon is also working on using libsignal-client-rs for CLI clients and bots with [Presage](https://gitlab.com/whisperfish/presage).

rust/protocol/src/sealed_sender.rs
```diff
@@ -722,9 +724,9 @@ pub async fn sealed_sender_encrypt_from_usmc<R: Rng + CryptoRng>(
 
     let static_key_ctext = crypto::aes256_ctr_hmacsha256_encrypt(
         &our_identity.public_key().serialize(),
-        &eph_keys.cipher_key()?,
+        array_ref![&eph_keys.cipher_key()?, 0, 32],
```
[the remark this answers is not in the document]
cosmicexplorer: Done! This is much nicer!

I've added `.expect()` calls instead of using `array_ref![]` throughout this change.

rust/protocol/src/curve.rs
cosmicexplorer: I have now simply avoided exposing any of those constants at all outside of `curve.rs` as nothing at this phase is relying on them!

rust/protocol/src/curve.rs
cosmicexplorer: Yes! I learned that the hard way by messing around last night! I'll make note of this.

rust/protocol/src/curve.rs
cosmicexplorer: Fixed!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Thanks so much for this clear feedback! I will make this change!

Question:
> (what you've called "client instance" elsewhere)

Is that referring to the changes in #286? I have a small followup PR from your comments at https://github.com/signalapp/libsignal-client/pull/286/files#r625973774 which makes `DeviceId` into its own struct and I will make sure to stitch this into there as well.

rust/protocol/src/storage/traits.rs
cosmicexplorer: Fixed!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Fixed!

rust/protocol/src/storage/traits.rs
cosmicexplorer: The reason the terminology was incorrect was because I was confused about what was being stored! Thanks, will reflect this.

cosmicexplorer: > "Users" are pretty much not interesting in libsignal-protocol; everything is device-to-device.

This sounds great!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Thanks! Done!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Thanks! Fixed!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Fixed!

rust/protocol/src/identity_key.rs
cosmicexplorer: :D probably!!! I wanted to use a strong adjective so that API consumers would be aware that this is where the randomness is injected (even though the trait bound on `R: CryptoRng + Rng` is probably enough to signal that already). I have made this change!

rust/protocol/src/identity_key.rs
cosmicexplorer: Thanks! I have removed all such comments, especially those in `inmem.rs`.

cosmicexplorer:
@Be-ing I'm working on what should become a CLI tool very analogous to `gpg` that performs forward-secret encrypted messaging and document signing! I had been working on a Rust wrapper to libsignal-protocol-c a while ago before to my amazement the whole thing was rewritten in Rust! I had been expecting to have to write a lot more stuff myself too so diverted my excitement towards documenting what was already done here instead.

rust/protocol/src/consts.rs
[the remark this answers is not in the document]
jrose-signal: I still tend to prefer keeping types and constants close to their primary use site, but clearly the code isn't doing that today. I'll come back to this.

rust/protocol/src/curve.rs
jrose-signal: There are definitely other places in the larger Signal codebase that want to know (or worse, make assumptions) but we'll deal with that as they come. :sob:

rust/protocol/src/storage/traits.rs
jrose-signal: Somewhat yeah. Within Signal "a device" usually refers to a particular installation that's been registered with a particular user and assigned a device ID (thus being uniquely identified by a ProtocolAddress), but the term you picked, "client instance", might be more accessible.

As I commented in your follow-up PR, not all of these are device IDs. :-)

rust/protocol/src/storage/traits.rs
jrose-signal: [marking this unresolved so I'm less likely to forget about it]

rust/protocol/src/storage.rs
```diff
@@ -1,10 +1,15 @@
...
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-mod inmem;
-mod traits;
+// TODO
+// #![warn(missing_docs)]
+
+//! Interfaces in [traits] and reference implementations in [inmem] for various mutable stores.
```
rubdos, line 9: :arrow_up_down:

rust/protocol/src/storage/inmem.rs
```diff
@@ -25,6 +31,10 @@ pub struct InMemIdentityKeyStore {
 }
 
 impl InMemIdentityKeyStore {
+    /// Create a new instance.
+    ///
+    /// `key_pair` corresponds to [traits::IdentityKeyStore::get_identity_key_pair], and `id`
+    /// corresponds to [traits::IdentityKeyStore::get_local_registration_id].
     pub fn new(key_pair: IdentityKeyPair, id: u32) -> Self {
```
rubdos, line 38: You may even want to rename the argument `id` to `registration_id` for clarity.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,70 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Traits defining several stores used throughout the Signal Protocol.
+
+// TODO
+// #![warn(missing_docs)]
```
rubdos, line 9: :arrow_up_down:

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,70 @@
...
+    address::ProtocolAddress,
+    error::Result,
+    sender_keys::SenderKeyRecord,
+    state::{PreKeyId, PreKeyRecord, SessionRecord, SignedPreKeyId, SignedPreKeyRecord},
+    IdentityKey, IdentityKeyPair,
 };
 
+/// Handle to FFI-provided context object.
+///
+/// This object is not manipulated in rust, but is instead passed back to each FFI
```
rubdos, line 24:
```suggestion
/// This object is not manipulated in Rust, but is instead passed back to each FFI
```

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,70 @@
...
+    ///
+    /// This local registration id is the same per-device identifier used in [ProtocolAddress] and
+    /// should not change run over run.
     async fn get_local_registration_id(&self, ctx: Context) -> Result<u32>;
 
+    /// Record an identity into the store. The identity is then considered "trusted".
+    ///
+    /// The return value represents whether an existing identity was replaced (`Ok(true)`). If it is
+    /// new or hasn't changed, the return value is `Ok(false)`.
+    // TODO: create a separate enum for this result instead of using a `bool`!
```
rubdos, line 59: Yes please. If you don't do this in this PR (and I don't think you should), maybe create an issue for this. I had to search the Java implementation to know what the bool should be :-)

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,70 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Traits defining several stores used throughout the Signal Protocol.
```
rubdos, line 6: I would add a word about the usage of `async-trait`, including the fact that it's `async_trait(?Send)`. Maybe some example/dummy implementation would be really great here.

rust/protocol/src/crypto.rs
```diff
@@ -61,36 +73,53 @@ pub fn aes_256_cbc_decrypt(ctext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
...
-    let mut result = Vec::with_capacity(ctext.len() + 10);
+    let mut result = Vec::with_capacity(ctext.len() + MAC_KEY_LENGTH);
     result.extend_from_slice(&ctext);
-    result.extend_from_slice(&mac[..10]);
+    result.extend_from_slice(&mac[..MAC_KEY_LENGTH]);
     Ok(result)
 }
 
+/// Validate the MAC `mac_key` against the ciphertext `ctext`, then decrypt `ctext` using
+/// [aes_256_ctr_decrypt] with `cipher_key`.
```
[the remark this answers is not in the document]
jrose-signal: I think those terms are less well-known, so if we wanted to use them we'd still have to define them (though possibly at the module level instead of here). Given that these are currently fairly minimal docs I think I'd go for the more wordy version that uses more common terms. @cosmicexplorer, do you have an opinion one way or the other?

rust/protocol/src/curve.rs
```diff
@@ -15,9 +22,20 @@ use arrayref::array_ref;
...
 use subtle::ConstantTimeEq;
 
+/// Encapsulate the variant of key being used.
+///
+/// Currently the only type of key Signal supports is "djb"-type [Self::Curve25519] keys, but in
+/// theory we could move to another key type that supports both *signatures* and *Diffie-Helman
+/// agreements* in the future. This would probably need to be another [elliptic curve], but that's
+/// not inherently necessary.
+///
+/// [elliptic curve]: https://en.wikipedia.org/wiki/Elliptic-curve_cryptography
```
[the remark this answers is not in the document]
jrose-signal: Ah yeah, that's my fault for not separating my normative comments from my descriptive comments. I agree with @rubdos that describing possible what future implementations might be based on is not really the point of the API.

rust/protocol/src/error.rs
```diff
@@ -1,16 +1,23 @@
...
 use std::error::Error;
 use std::fmt;
 use std::panic::UnwindSafe;
 
+/// Return type for all fallible operations in the Signal Protocol.
 pub type Result<T> = std::result::Result<T, SignalProtocolError>;
 
+/// Error states recognized by the Signal Protocol.
 #[derive(Debug)]
 pub enum SignalProtocolError {
```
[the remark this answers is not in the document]
jrose-signal: We had a contributor previously suggest using [thiserror](https://docs.rs/thiserror/1.0.24/thiserror/) to generate our error formatting in #140, and we decided not to go with it just for simplicity's sake. However, I wonder if something like [displaydoc](https://docs.rs/displaydoc/0.2.1/displaydoc/) would tilt the scales in the other direction…

rust/protocol/src/identity_key.rs
```diff
@@ -51,20 +57,23 @@ impl From<PublicKey> for IdentityKey {
...
 impl IdentityKeyPair {
+    /// Create a new key pair from a public `identity_key` and a private `private_key`.
     pub fn new(identity_key: IdentityKey, private_key: PrivateKey) -> Self {
         Self {
             identity_key,
             private_key,
         }
     }
 
+    /// Generate a cryptographically random new identity from randomness in `csprng`.
```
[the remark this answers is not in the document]
jrose-signal: This one's partially my fault too.

rust/protocol/src/protocol.rs
```diff
@@ -52,11 +54,11 @@ impl CiphertextMessage {
 
 #[derive(Debug, Clone)]
 pub struct SignalMessage {
-    message_version: u8,
+    message_version: VersionType,
```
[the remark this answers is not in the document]
jrose-signal: @cosmicexplorer has been doing more on this in #289. I think we should probably leave the type changes out of this PR just to have them not conflict, but I'll note that PreKeySignalMessage and SignalMessage *do* share a version. (I'm double-checking whether we want SenderKeyMessage and SenderKeyDistributionMessage to share that same version, or a different one.)

rust/protocol/src/storage/traits.rs
gferon: I started to look into having the store traits require `Send`, just to see how far I could go, would you mind if I open an issue so that we can discuss it in more details there?

jrose-signal: @gferon That sounds great, thank you!

rust/protocol/src/crypto.rs
```diff
@@ -61,36 +73,53 @@ pub fn aes_256_cbc_decrypt(ctext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
...
 }
 
+/// Length of the MAC key used for [aes256_ctr_hmacsha256_encrypt] and
+/// [aes256_ctr_hmacsha256_decrypt].
+pub const MAC_KEY_LENGTH: usize = 10;
+
+/// Encrypt plaintext `msg` using [aes_256_ctr_encrypt] with `cipher_key`. Then generate a MAC from
+/// the resulting ciphertext with [hmac_sha256] using `mac_key`, and append that to the result.
+///
+/// Only the first [MAC_KEY_LENGTH] bytes of the MAC key are used.
```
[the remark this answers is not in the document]
cosmicexplorer: Aha! Yes! But in fact I meant instead
```suggestion
/// Only the first [MAC_KEY_LENGTH] bytes of the MAC are used.
```

rust/protocol/src/crypto.rs
```diff
@@ -41,6 +51,8 @@ pub fn aes_256_cbc_encrypt(ptext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
     }
 }
 
+/// Decrypt ciphertext `ctext` using key `key` and initialization value `iv` with AES-256 in
+/// CBC mode.
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you!

rust/protocol/src/crypto.rs
```diff
@@ -13,25 +17,31 @@ use hmac::{Hmac, Mac, NewMac};
...
 }
 
-pub fn aes_256_ctr_decrypt(ctext: &[u8], key: &[u8]) -> Result<Vec<u8>> {
+/// Decrypt ciphertext `ctext` using key `key` with AES-256 in CTR mode.
+pub fn aes_256_ctr_decrypt(ctext: &[u8], key: &[u8; AES_256_KEY_SIZE]) -> Result<Vec<u8>> {
     aes_256_ctr_encrypt(ctext, key)
 }
 
+/// Encrypt plaintext `ptext` using key `key` and initialization value `iv` with AES-256 in
+/// CBC mode.
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you!

rust/protocol/src/crypto.rs
```diff
@@ -1,8 +1,12 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+#![warn(missing_docs)]
+
+//! Application of cryptographic primitives, including HMAC and AES.
+
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you, noted and fixed! Thanks for the link!

rust/protocol/src/curve.rs
```diff
@@ -1,9 +1,16 @@
...
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-mod curve25519;
+// TODO
+// #![warn(missing_docs)]
+
+//! Identity creation primitives.
```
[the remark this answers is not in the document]
cosmicexplorer: :D thank you for noting!

rust/protocol/src/curve.rs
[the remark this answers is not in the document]
cosmicexplorer: Thank you for clarifying the appropriate boundaries for API docs! Will remove that entire paragraph.

rust/protocol/src/error.rs
```diff
@@ -1,16 +1,23 @@
...
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-use crate::curve::KeyType;
+// TODO: add documentation for each of the error cases!
+// #![warn(missing_docs)]
+
+//! Errors that may occur during various stages of the Signal Protocol.
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you!

rust/protocol/src/error.rs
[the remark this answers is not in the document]
cosmicexplorer: I've added documentation to all of the error cases as best I can. Please review the changes!

rust/protocol/src/identity_key.rs
```diff
@@ -1,8 +1,13 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+// TODO
+// #![warn(missing_docs)]
+
+//! Wrappers over identity primitives from [crate::curve].
```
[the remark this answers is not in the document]
cosmicexplorer: Fixed!

rust/protocol/src/identity_key.rs
```diff
@@ -51,20 +57,23 @@ impl From<PublicKey> for IdentityKey {
...
 
+/// Wrapper for [KeyPair].
 #[derive(Copy, Clone)]
 pub struct IdentityKeyPair {
     identity_key: IdentityKey,
     private_key: PrivateKey,
 }
 
 impl IdentityKeyPair {
+    /// Create a new key pair from a public `identity_key` and a private `private_key`.
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you for walking through your thinking! This sounds great!

rust/protocol/src/identity_key.rs
```diff
@@ -11,6 +16,7 @@ use std::convert::TryFrom;
 
 use prost::Message;
 
+/// Wrapper for [PublicKey].
```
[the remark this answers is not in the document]
cosmicexplorer: Thank you! Fixed!

rust/protocol/src/identity_key.rs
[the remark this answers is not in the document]
cosmicexplorer: Fixed and noted!

rust/protocol/src/storage.rs
cosmicexplorer: Done!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Done! We can actually remove the `#![warn(missing_docs)]` in this module and just set it in the parent module in `storage.rs`!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Thanks! Done!

rust/protocol/src/storage/traits.rs
cosmicexplorer: I have modified this TODO to point to the issue #295 I just made for this! :D Thank you!

rust/protocol/src/storage/inmem.rs
cosmicexplorer: Done! Thanks!

rust/protocol/src/crypto.rs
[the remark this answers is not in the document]
cosmicexplorer: I am personally completely unfamiliar with NaCL/libsodium (sadly!!) but I think it would definitely be good to use canonical terminology if it exists. I don't have the context to really make the decision here but I think having consistent terminology is very useful so I would be super open to describing "sealing" and "opening" in the module-level docs, in a future PR? I will add that to #285!

rust/protocol/src/protocol.rs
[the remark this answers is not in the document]
cosmicexplorer: I think I will revert the changes to use `VersionType` in this PR, as that should be wholly superseded by #289, where we can figure out (as @jrose-signal mentioned) whether these message versions should actually be sharing the same type.

rust/protocol/src/error.rs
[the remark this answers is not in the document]
cosmicexplorer: See #296 for a really lightweight integration of `thiserror`!

rust/bridge/node/futures/src/lib.rs
```diff
@@ -19,7 +19,7 @@
 //! [Neon]: https://neon-bindings.com/
 
 #![feature(wake_trait)]
-#![warn(missing_docs)]
+#![deny(missing_docs)]
```
jrose-signal, line 22: `warn` is deliberate, and actually I think we should standardize on it, because it lets you write code locally without docs. The libraries themselves should have zero warnings when compiled with their default toolchain.

rust/protocol/src/crypto.rs
```diff
@@ -13,25 +17,31 @@ use hmac::{Hmac, Mac, NewMac};
 use sha2::Sha256;
 use subtle::ConstantTimeEq;
 
-pub fn aes_256_ctr_encrypt(ptext: &[u8], key: &[u8]) -> Result<Vec<u8>> {
-    if key.len() != 32 {
-        return Err(SignalProtocolError::InvalidCipherCryptographicParameters(
-            32, 0,
-        ));
-    }
+const BYTE: usize = 8;
```
jrose-signal, line 20: I'd call this `BITS_PER_BYTE` if you think it's useful to specify the sizes in bits.

rust/protocol/src/crypto.rs
```diff
@@ -61,36 +73,53 @@ pub fn aes_256_cbc_decrypt(ctext: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8
         .map_err(|_| SignalProtocolError::InvalidCiphertext)?)
 }
 
-pub fn hmac_sha256(key: &[u8], input: &[u8]) -> Result<[u8; 32]> {
+/// The statically-known size of the output of [hmac_sha256].
+pub const HMAC_OUTPUT_SIZE: usize = 32;
```
jrose-signal, line 77: Arguably this is also bits-based (SHA-256 produces 256-bit hashes, which is 32 bytes.)

rust/protocol/src/crypto.rs
```diff
@@ -136,7 +166,7 @@ mod test {
             .expect("valid hex");
         let ptext = [0u8; 35];
 
-        let ctext = super::aes_256_ctr_encrypt(&ptext, &key)?;
+        let ctext = super::aes_256_ctr_encrypt(&ptext, &key.try_into().unwrap())?;
```
jrose-signal, line 169: We have a policy (enforced by Clippy, I thought?) to only use `expect` and never `unwrap`, forcing documentation of why it's safe.

rust/protocol/src/curve/curve25519.rs
```diff
@@ -1,8 +1,12 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Implementation of the [Curve25519] elliptic curve.
+//!
+//! [Curve25519]: https://en.wikipedia.org/wiki/Curve25519
```
jrose-signal, line 8: This is more "implementation of [X25519](https://en.wikipedia.org/wiki/Curve25519) and [Ed25519](https://en.wikipedia.org/wiki/EdDSA#Ed25519)". `curve25519_dalek` implements the curve operations themselves.

rust/protocol/src/error.rs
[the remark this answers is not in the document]
jrose-signal: Uh, hm. These'll need more attention, I'll come back to them.

rust/protocol/src/lib.rs
```diff
@@ -42,6 +42,7 @@ pub use {
     protocol::{
         CiphertextMessage, CiphertextMessageType, PreKeySignalMessage,
         SenderKeyDistributionMessage, SenderKeyMessage, SignalMessage,
+        CIPHERTEXT_MESSAGE_CURRENT_VERSION,
```
jrose-signal, line 45: I don't think this should be exported. What did you need it for?

rust/protocol/src/sealed_sender.rs
```diff
@@ -512,7 +515,7 @@ impl UnidentifiedSenderMessageContent {
     }
 }
 
-enum UnidentifiedSenderMessage {
+pub enum UnidentifiedSenderMessage {
```
jrose-signal, line 518: This is deliberately not public.

rust/protocol/src/sealed_sender.rs
```diff
@@ -735,7 +742,11 @@ pub async fn sealed_sender_encrypt_from_usmc<R: Rng + CryptoRng>(
 
     let message_data = crypto::aes256_ctr_hmacsha256_encrypt(
         usmc.serialized()?,
-        &static_keys.cipher_key()?,
+        &static_keys.cipher_key().and_then(|key| {
+            key.try_into().map_err(|_: TryFromSliceError| {
+                SignalProtocolError::BadKeyLength(curve::KeyType::Curve25519, key.len())
+            })
+        })?,
```
jrose-signal, line 749: This seems like a type system failure; we should never get another length of key from the intermediate step here of calculating static keys.

rust/protocol/src/storage.rs
```diff
@@ -1,10 +1,14 @@
...
 //
 
-mod inmem;
-mod traits;
+//! Interfaces in [traits] and reference implementations in [inmem] for various mutable stores.
+
+#![deny(missing_docs)]
+
+pub mod inmem;
+pub mod traits;
```
jrose-signal, line 11: No module structure yet, please!

rust/protocol/src/storage/inmem.rs
```diff
@@ -287,6 +304,11 @@ pub struct InMemSignalProtocolStore {
 }
 
 impl InMemSignalProtocolStore {
+    /// Create an object with the minimal implementation of [traits::ProtocolStore], representing
+    /// the given identity `key_pair` and with the separate random seed `registration_id`.
+    ///
+    /// *TODO: make `registration_id` into a more type-safe wrapper struct! See
+    /// <https://github.com/signalapp/libsignal-client/pull/289>.*
```
jrose-signal, line 311: I'd rather not refer to specific PRs in the code. Even Issues we'd been staying away from but I see how that's useful for ongoing efforts.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,73 @@
...
+    error::Result,
+    sender_keys::SenderKeyRecord,
+    state::{PreKeyId, PreKeyRecord, SessionRecord, SignedPreKeyId, SignedPreKeyRecord},
+    IdentityKey, IdentityKeyPair,
 };
 
+/// Handle to FFI-provided context object.
+///
+/// This object is not manipulated in Rust, but is instead passed back to each FFI
+/// method invocation. This argument should just be [None] for all clients of the Rust-only API.
```
jrose-signal, line 22: This isn't inherently true, but it's good enough while we talk about what to do with Context in relation to whether the async traits should require Send.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,73 @@
...
+    /// We are in the context of sending a message.
     Sending,
+    /// We are in the context of receiving a message.
     Receiving,
 }
 
+/// Interface defining the identity store, which may be in-memory, on-disk, etc.
+///
+/// Signal clients usually use the identity store in a [TOFU] manner, but this is not at
+/// all guaranteed.
```
jrose-signal, line 42: I'd say it's not *required.* The trait has a "save" API and an "is trusted" requirement, but the implementing type could choose some strategy other than TOFU.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,73 @@
...
+/// [TOFU]: https://en.wikipedia.org/wiki/Trust_on_first_use
 #[async_trait(?Send)]
 pub trait IdentityKeyStore {
+    /// Return the single specific identity the store is assumed to represent, with private key.
     async fn get_identity_key_pair(&self, ctx: Context) -> Result<IdentityKeyPair>;
 
+    /// Return a [u32] specific to this store instance.
+    ///
+    /// This local registration id is the same per-device identifier used in [ProtocolAddress] and
+    /// should not change run over run.
```
jrose-signal, line 53: Looks leftover from when the DeviceID and RegistrationID were conflated. A good example of this came up recently (prompting #304): if you unregister a linked device and then register a new one, the device ID might get reused, but the registration ID will be generated fresh.

rust/protocol/src/storage/traits.rs
```diff
@@ -41,35 +76,43 @@ pub trait IdentityKeyStore {
...
     ) -> Result<Option<IdentityKey>>;
 }
 
+/// Interface for storing pre-keys downloaded from a server.
 #[async_trait(?Send)]
 pub trait PreKeyStore {
+    /// Look up the pre-key from this store corresponding to `prekey_id`.
     async fn get_pre_key(&self, prekey_id: PreKeyId, ctx: Context) -> Result<PreKeyRecord>;
 
+    /// Set the entry for `prekey_id` to the value of `record` in this store.
```
jrose-signal, line 93: Nitpick: I think the "from this store" and "in this store" are redundant in most of these accessors.

rust/protocol/src/storage/traits.rs
```diff
@@ -94,8 +146,10 @@ pub trait SessionStore {
     ) -> Result<()>;
 }
 
+/// Interface for storing sender keys for other users, allowing multiple keys per user.
```
jrose-signal, line 149: Let's say "sender key records", since they're more than just keys. (They're sessions of their own, just not Double Ratchet sessions.)

rust/protocol/src/storage/inmem.rs
```diff
@@ -1,34 +1,41 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Implementations for stores defined in [super::traits].
```
jrose-signal, line 6: I think it's worth mentioning that they're "in-memory" or "ephemeral" or "for testing".

rust/protocol/src/protocol.rs
```diff
@@ -15,6 +15,8 @@ use sha2::Sha256;
 use subtle::ConstantTimeEq;
 use uuid::Uuid;
 
+/// Each [SignalMessage] has this message version mixed into it when created. Prefer to reference
+/// this value over use of the literal version number.
 pub const CIPHERTEXT_MESSAGE_CURRENT_VERSION: u8 = 3;
 pub const SENDERKEY_MESSAGE_CURRENT_VERSION: u8 = 3;
```
jrose-signal, line 21: I guess this new one needs a doc comment too.

rust/protocol/src/identity_key.rs
```diff
@@ -1,36 +1,53 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Wrappers over identity primitives from [crate::curve].
```
jrose-signal, line 6: I think part of the point is that the keys in `crate::curve` are sometimes used for non-identity purposes; these wrappers distinguish the ones that are used as identities.

rust/protocol/src/curve.rs
```diff
@@ -1,55 +1,80 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2021 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-mod curve25519;
+//! Identity creation primitives.
```
jrose-signal, line 6: Similarly (see identity_key.rs), these are keys but not necessarily identity keys.

rust/protocol/src/curve.rs
```diff
@@ -1,55 +1,80 @@
...
+/// Interface for structs that perform operations parameterized by values of [KeyType].
+pub trait Keyed {
+    /// Return the variant of key this object employs.
+    fn key_type(&self) -> KeyType;
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 enum PublicKeyData {
-    DjbPublicKey([u8; 32]),
+    Curve25519PublicKey([u8; PUBLIC_KEY_LENGTH]),
```
jrose-signal, line 71: This `PUBLIC_KEY_LENGTH` would probably look better *with* qualification, so it's obvious where we're getting it from. (If we ever had two types it would certainly be necessary.)

rust/protocol/src/curve.rs
```diff
@@ -60,56 +85,70 @@ impl PublicKey {
...
         }
         result.into_boxed_slice()
     }
 
+    /// Validate whether `signature` successfully matches `message` for this public key.
+    ///
+    /// Return `Ok(false)` if the signature fails to match or could not be read.
+    ///
+    /// *TODO: rewrite this method to consume static-sized arrays.
+    /// See <https://github.com/signalapp/libsignal-client/issues/295>.*
```
jrose-signal, line 148: Like the key lengths not being fixed in this API, the signature length can't really be fixed either. We've essentially got constants associated with particular key types, and pushing them into the type system would only be useful if we statically knew a key had a certain type…which we sometimes do, but often don't.

rust/protocol/src/curve.rs
```diff
@@ -122,15 +161,21 @@ impl PublicKey {
         }
     }
 
+    /// Return the bytes in memory associated with this public key.
+    ///
+    /// *TODO: rewrite this method and [PrivateKey::key_data] to expose a static-sized array.
+    /// See <https://github.com/signalapp/libsignal-client/issues/295>.*
```
jrose-signal, line 167: Ditto.

rust/protocol/src/curve.rs
jrose-signal: Also while you're here we probably don't need to repeat "PublicKey" in the case name.

rust/protocol/src/curve.rs
```diff
@@ -197,73 +241,96 @@ impl fmt::Debug for PublicKey {
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 enum PrivateKeyData {
-    DjbPrivateKey([u8; 32]),
+    Curve25519PrivateKey([u8; PRIVATE_KEY_LENGTH]),
```
jrose-signal, line 244: Same here as for PublicKeyData

rust/protocol/src/curve.rs
```diff
@@ -280,41 +347,118 @@ impl TryFrom<&[u8]> for PrivateKey {
...
+/// use libsignal_protocol::{KeyPair, PrivateKey, PublicKey};
+/// use rand::rngs::OsRng;
+/// use std::convert::TryInto;
+///
+/// let alice = KeyPair::generate(&mut OsRng);
+/// assert!(alice == alice.clone());
+/// let bob = KeyPair::generate(&mut OsRng);
+/// assert!(alice != bob);
+///
+/// // The first byte of `alice_zero` is 0, less than `bob_one` with 1, lexicographically:
```
jrose-signal, line 404: I don't think this is a good example. For one thing, the first and last bytes of Curve25519 keys have restrictions (the ones enforced in deserialization). But more importantly, there doesn't need to be a guarantee of *how* the keys are ordered, only that they can be.

rust/protocol/src/curve.rs
```diff
@@ -280,41 +347,118 @@ impl TryFrom<&[u8]> for PrivateKey {
...
+/// let mut alice_bytes: [u8; 32] = alice.public_key.public_key_bytes()?.try_into().unwrap();
+/// alice_bytes[0] = 0;
+/// let alice_zero = PublicKey::from_djb_public_key_bytes(alice_bytes.as_ref())?;
```
jrose-signal, line 407: Did this miss the name change?

rust/protocol/src/curve.rs
jrose-signal: …also, none of this has to do with KeyPair. Maybe it's module-level docs?

rust/protocol/src/curve.rs
```diff
@@ -280,41 +347,118 @@ impl TryFrom<&[u8]> for PrivateKey {
...
+/// // The same sort of thing can be done with private keys:
+/// let mut alice_private_bytes: [u8; 32] = alice.private_key.serialize().try_into().unwrap();
+/// alice_private_bytes[0] = !alice_private_bytes[0];
+/// let alice_flipped = PrivateKey::deserialize(alice_private_bytes.as_ref())?;
+/// assert!(alice_flipped != alice.private_key);
+///
+/// # Ok(())
+/// # }
+///```
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
```
jrose-signal, line 425: I don't think this needs to be Ord, actually. Eq is reasonable.

jrose-signal (commented):
Many more comments (this may drag out a while since we both have other things to work on!)

rust/protocol/src/crypto.rs
cosmicexplorer: So that's true, but it appears that clippy will not enforce that for code that's only visible during testing. If I apply the following change, clippy correctly yells at me:
```diff
diff --git a/rust/protocol/src/crypto.rs b/rust/protocol/src/crypto.rs
index 89288eae3..d0e52062f 100644
--- a/rust/protocol/src/crypto.rs
+++ b/rust/protocol/src/crypto.rs
@@ -126,6 +126,11 @@ pub fn aes256_ctr_hmacsha256_decrypt(
     aes_256_ctr_decrypt(&ctext[..ptext_len], cipher_key)
 }

+pub fn asdf() {
+    let e: ::std::result::Result<(), String> = Err("asdf".to_string());
+    e.unwrap();
+}
+
 #[cfg(test)]
 mod test {
     use super::Result;
```

I will make the change to use `.expect()` here. We could probably figure out how to make clippy check test code too.

rust/protocol/src/crypto.rs
cosmicexplorer: I like specifying them in bits so that they can be used to index into a slice directly without confusion about whether you should multiply by 8 yourself. I think the nicer solution to this would be a library that uses types for units to avoid ever getting this wrong.

rust/bridge/node/futures/src/lib.rs
cosmicexplorer: That sounds great, thanks! Done!

rust/protocol/src/crypto.rs
cosmicexplorer: Oh, yes! I have modified this to be `256 / BITS_PER_BYTE` as well.

rust/protocol/src/curve/curve25519.rs
cosmicexplorer: Done! I also added a ton of really slick doctests to this module!

rust/protocol/src/lib.rs
cosmicexplorer: I have removed this export!

rust/protocol/src/sealed_sender.rs
cosmicexplorer: Got it! Reverted!

rust/protocol/src/identity_key.rs
cosmicexplorer: Made this more clear! Let me know if the result sounds good!

rust/protocol/src/curve.rs
cosmicexplorer: Tried to fix this! Let me know if the result works!

rust/protocol/src/curve.rs
cosmicexplorer: Got it! Fixed!

rust/protocol/src/curve.rs
cosmicexplorer: Done!

rust/protocol/src/curve.rs
cosmicexplorer: Done and done!

rust/protocol/src/utils.rs
```diff
@@ -31,26 +31,39 @@ fn ct_select(mask: u8, a: u8, b: u8) -> u8 {
...
+///    Ordering enum.
+///     - *This seems unavoidable.*
+///
+/// #### Possible Improvements
+/// - It would be possible to instead have this function SHA-256 hash both
+///   inputs, then compare the resulting hashes in the usual non-constant
+///   time way.
+///     - *We avoid this approach at the moment since it is not clear if applications will rely on
+///       public key ordering being defined in some particular way or not.*
+pub fn constant_time_cmp(x: &[u8], y: &[u8]) -> Ordering {
```
cosmicexplorer, line 66: We are unable to `use libsignal_protocol::utils_unstable::constant_time_cmp;` in the doctest unless we can *declare* this method as `pub`. Currently, outside of `utils_unstable`, this method `constant_time_cmp` is still not exposed to consumers of this crate. See discussion of `*_unstable` modules in `lib.rs`.

rust/protocol/src/storage/inmem.rs
cosmicexplorer: Done!

rust/protocol/src/storage/inmem.rs
cosmicexplorer: Ok, that's totally fine! In other codebases I was used to doing this -- it is not necessary and I will remove this and the issue itself can be the way we stay on top of those things.

rust/protocol/src/storage/traits.rs
cosmicexplorer: That simplifies it a lot, thanks!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Thank you for that issue link! I have tried to transcribe that succintly!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Absolutely right!

rust/protocol/src/storage/traits.rs
cosmicexplorer: This was a super helpful remark! Thanks!

rust/protocol/src/sealed_sender.rs
cosmicexplorer: Yes! I have refactored these methods to return static-sized arrays in all possible cases, which allows us to avoid re-checking these lengths as often.

rust/protocol/src/storage.rs
cosmicexplorer: Removed!

rust/protocol/src/curve.rs
cosmicexplorer: I have removed this example, removed the `Ord` impl, and moved a smaller version of this doctest to the module-level docs!

rust/protocol/src/curve.rs
cosmicexplorer: Yes it did! Fixed!

rust/protocol/src/curve/curve25519.rs
```diff
@@ -36,6 +65,17 @@ impl KeyPair {
         }
     }
 
+    /// Do a [DH] key exchange to produce a slice that can be reproduced by another keypair.
```
jrose-signal, line 68: Maybe "the complementary keys"? It's a bit weird here because `self` is a KeyPair but only the PrivateKey part is used.

rust/protocol/src/curve/curve25519.rs
```diff
@@ -95,6 +142,17 @@ impl KeyPair {
         result
     }
 
+    /// Verify a signature from [Self::calculate_signature] against another key pair's public key.
```
jrose-signal, line 145: There's no "another" here, since the function doesn't take `self`.

rust/protocol/src/curve/curve25519.rs
```diff
@@ -133,16 +191,24 @@ impl KeyPair {
         bool::from(cap_r_check.as_bytes().ct_eq(&cap_r))
     }
 
+    /// Return a public key that another party can use to validate against [Self::verify_signature].
```
jrose-signal, line 194: …or calculate a key agreement. I think this is just "Returns the public key in the pair."

rust/protocol/src/curve/curve25519.rs
```diff
@@ -133,16 +191,24 @@ impl KeyPair {
...
     pub fn public_key(&self) -> &[u8; PUBLIC_KEY_LENGTH] {
         &self.public_key
     }
 
+    /// Return a private key that can be converted into a public key with [derive_public_key].
```
jrose-signal, line 199: Similarly, the purpose of a private key is usually not to derive a public key.

rust/protocol/src/curve/curve25519.rs
```diff
@@ -133,16 +191,24 @@ impl KeyPair {
...
 }
 
-pub fn derive_public_key(private_key: &[u8; 32]) -> [u8; 32] {
+/// Produce the result of [KeyPair::public_key], when given the output of [KeyPair::private_key].
+///```
+/// # use libsignal_protocol::curve_unstable::curve25519::*;
+/// let kp = KeyPair::new(&mut rand::thread_rng());
+/// assert!(kp.public_key() == &derive_public_key(kp.private_key()));
+///```
+pub fn derive_public_key(private_key: &[u8; PRIVATE_KEY_LENGTH]) -> [u8; PUBLIC_KEY_LENGTH] {
```
jrose-signal, line 211: "Produce the public key corresponding to the given private key". KeyPair is not the fundamental API here, even if "a private/public key pair" is the underlying abstraction.

rust/protocol/src/error.rs
```diff
@@ -1,68 +1,182 @@
...
+    protocol::{SenderKeyMessage, SignalMessage, CIPHERTEXT_MESSAGE_CURRENT_VERSION},
+    ratchet::{ChainKey, RootKey},
+    sealed_sender::{sealed_sender_decrypt, sealed_sender_multi_recipient_encrypt},
+    sender_keys::SenderKeyRecord,
+    session_cipher::{message_decrypt, message_encrypt},
+    state::{PreKeyId, PreKeyRecord, SignedPreKeyId, SignedPreKeyRecord},
+    storage::{IdentityKeyStore, PreKeyStore, SenderKeyStore, SessionStore, SignedPreKeyStore},
+};
+#[cfg(doc)]
+use uuid::Uuid;
```
jrose-signal, line 26: Let's put these below the regular imports so that someone scanning imports doesn't get caught on these.

rust/protocol/src/identity_key.rs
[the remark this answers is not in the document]
jrose-signal: Er, conversely I'd like to remove the reference to IdentityKeyStore, since that's not necessarily the only place such a key is used. The other part makes sense to me, though.

rust/protocol/src/identity_key.rs
```diff
@@ -1,46 +1,63 @@
...
+    /// Return a public key representing the public identity.
     #[inline]
     pub fn public_key(&self) -> &PublicKey {
         &self.public_key
     }
 
+    /// Return an owned byte slice which can be deserialized with [Self::decode].
     #[inline]
-    pub fn serialize(&self) -> Box<[u8]> {
+    pub fn serialize(&self) -> [u8; 1 + PUBLIC_KEY_LENGTH] {
```
jrose-signal, line 46: Here's a fixed-sized array that needs to be variable.

rust/protocol/src/identity_key.rs
```diff
@@ -51,20 +68,29 @@ impl From<PublicKey> for IdentityKey {
...
+///
+/// Also see [KeyPair].
+#[derive(Debug, Copy, Clone)]
 pub struct IdentityKeyPair {
     identity_key: IdentityKey,
     private_key: PrivateKey,
 }
 
 impl IdentityKeyPair {
+    /// Create a key pair from a public `identity_key` and a private `private_key`.
```
jrose-signal, line 85: Probably worth commenting (and for that matter, we should be `debug_assert!`ing) that the public key in `identity_key` must match the public key derivable from `private_key`.

rust/protocol/src/protocol.rs
```diff
@@ -15,7 +22,10 @@ use sha2::Sha256;
 use subtle::ConstantTimeEq;
 use uuid::Uuid;
 
+/// Each type of message in the Signal Protocol has a message version mixed into it when
+/// created. Prefer to reference this value over use of the literal version number.
 pub const CIPHERTEXT_MESSAGE_CURRENT_VERSION: u8 = 3;
+/// Each [SenderKeyRecord] has a message version field as well.
 pub const SENDERKEY_MESSAGE_CURRENT_VERSION: u8 = 3;
```
jrose-signal, line 29: These comments should each stand alone, though you could add a "see also". If we're adding doc comments, they should look good in the generated rustdoc output as well as in the source, and that might not put them next to each other.

rust/protocol/src/protocol.rs
```diff
@@ -79,7 +89,7 @@ impl SignalMessage {
         receiver_identity_key: &IdentityKey,
     ) -> Result<Self> {
         let message = proto::wire::SignalMessage {
-            ratchet_key: Some(sender_ratchet_key.serialize().into_vec()),
+            ratchet_key: Some(sender_ratchet_key.serialize().to_vec()),
```
jrose-signal, line 92: Don't forget to take these back out when you put the variable-length PublicKey/PrivateKey API back.

rust/protocol/src/protocol.rs
```diff
@@ -403,29 +413,63 @@ impl SenderKeyMessage {
...
-    pub fn verify_signature(&self, signature_key: &PublicKey) -> Result<bool> {
-        let valid = signature_key.verify_signature(
-            &self.serialized[..self.serialized.len() - Self::SIGNATURE_LEN],
-            &self.serialized[self.serialized.len() - Self::SIGNATURE_LEN..],
-        )?;
+    /// Return a tuple containing:
+    /// 1. the serialized message *without* the signature at the end.
+    /// 2. the complete serialized message *with* signature.
+    /// 3. the signature as a fixed-size array.
+    fn split_serialized_signature(
```
jrose-signal, line 442: It's bad enough that we're keeping both the serialized and unserialized forms of most of these messages; we definitely should not be copying them *again* just to have fixed-sized arrays. I think this won't matter when you put back the variable-length APIs, but in general this would be a good place for `array_ref!` or similar to reference bytes out of the buffer rather than copying it more than it already is.

rust/protocol/src/protocol.rs
```diff
@@ -797,14 +858,15 @@ impl TryFrom<&[u8]> for DecryptionErrorMessage {
...
             .timestamp
             .ok_or(SignalProtocolError::InvalidProtobufEncoding)?;
-        let ratchet_key = proto_structure
+        let ratchet_key: Option<PublicKey> = proto_structure
             .ratchet_key
-            .map(|k| PublicKey::deserialize(&k))
-            .transpose()?;
+            .map(|k| PublicKey::deserialize_result(&k))
+            .transpose()?
+            .clone();
```
jrose-signal, line 865: `clone()`?

jrose-signal (commented):
Only some comments here (I'll try to get more this week, which has more downtime). I think there are two meta-comments, though:

- It's not correct to make the PublicKey and PrivateKey APIs traffic in fixed-size buffers, as already discussed.
- This ended up being too many changes for one PR. I'm sorry to make more work for you, but can you break them up into smaller logical chunks? It's not just adding docstrings at this point.

jrose-signal:
Looks like there are still comments from our last round that haven't been incorporated. I do think it makes sense to keep peeling this one off into separate PRs though!

cosmicexplorer:
Going to turn this into a draft PR and peel off sections as discussed!

Issue #30, opened 2020-11-04
java: Code formatting


jack-signal:
Adopt a code formatter for Java, use it, and enforce it in CI. We want something usable from the command line, no IDE required. One possible option https://github.com/google/google-java-format

PR #288, opened 2021-05-04
Java: remove unused imports

### LINKED ISSUE #29: java: Remove unused imports
The Java code has many unused imports as remainders from the libsignal-protocol-java import. Find a tool that can detect them.

hiqua:
Closes #29.

Just used IntelliJ IDEA "optimize imports" feature.

### DIFF
--- a/java/android/src/androidTest/java/org/whispersystems/libsignal/CurveTest.java
+++ b/java/android/src/androidTest/java/org/whispersystems/libsignal/CurveTest.java
@@ -1,7 +1,6 @@
 package org.whispersystems.libsignal;
 
 import junit.framework.TestCase;
-
 import org.whispersystems.libsignal.ecc.Curve;
 import org.whispersystems.libsignal.ecc.ECKeyPair;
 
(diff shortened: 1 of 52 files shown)


jrose-signal:
Thanks! I'm a bit confused by how it's decided to sort the imports, though: it's not alphabetical (at the top level) and it sometimes uses groups and sometimes doesn't. Are there configuration options we could turn on/off to reduce the amount of reordering?

hiqua:
This is with these settings (defaults I think):

Of course these could be changed.

2021-05-04-213149_1524x543_scrot

jrose-signal:
Ah, great! Going off of what we currently have, I think the preferred groups are something like

- android
- junit
- org.signal.client.internal.Native specifically
- org.signal, org.whispersystems
- java, javax

though it's not super consistent. I'd simplify that to

- org.signal.client.internal.Native specifically
- org.signal, org.whispersystems
- all other packages
- java, javax (it's weird to me that IntelliJ uses the opposite order)

What we'd really like, though, is to add a formatter to Gradle so that these *stay* in order—it doesn't even exactly matter which order, if it's not easily configurable. Do you have any leads on that?

hiqua:
I think @jack-signal already hinted at this there: https://github.com/signalapp/libsignal-client/issues/30

What if you just use the google-java-formatter for all your java code (including the imports)? Then it would be easy to add it to the CI.

PR #289, opened 2021-05-04
require explicit conversions to and from address::DeviceId

cosmicexplorer:
### Problem
See https://github.com/signalapp/libsignal-client/pull/286/files#r625970592 from #286 for motivation. We currently use `u32` to mean different things in different places, including in the `.device_id()` method of `address::ProtocolAddress`. We would prefer instead to require an explicit `.into()` from an intermediate struct that looks like:
```rust
#[derive(Copy, Clone)]
pub struct DeviceId(u32);
impl From<u32> for DeviceId { ... }
impl From<DeviceId> for u32 { ... }
```

So note that this is *not* part of the documentation push in #285, but rather a separate type of refactoring change.

### Solution
(all of the below also have docstrings)
1. Create `address::DeviceId` wrapping `u32`.
1. Modify the existing definition of `storage::prekey::PreKeyId` wrapping `u32`.
1. Modify the existing definition of `storage::signed_prekey::SignedPreKeyId` wrapping `u32`.
1. Apply `.into()` calls to convert to and from `u32` as necessary to make tests pass.

### Result
We don't accidentally mistake a device id for a chain counter!

### DIFF
--- a/rust/bridge/shared/src/ffi/storage.rs
+++ b/rust/bridge/shared/src/ffi/storage.rs
@@ -197,12 +197,12 @@ pub struct FfiPreKeyStoreStruct {
 impl PreKeyStore for &FfiPreKeyStoreStruct {
     async fn get_pre_key(
         &self,
-        prekey_id: u32,
+        prekey_id: PreKeyId,
         ctx: Context,
     ) -> Result<PreKeyRecord, SignalProtocolError> {
         let ctx = ctx.unwrap_or(std::ptr::null_mut());
         let mut record = std::ptr::null_mut();
-        let result = (self.load_pre_key)(self.ctx, &mut record, prekey_id, ctx);
+        let result = (self.load_pre_key)(self.ctx, &mut record, prekey_id.into(), ctx);
 
         if let Some(error) = CallbackError::check(result) {
             return Err(SignalProtocolError::ApplicationCallbackError(
@@ -221,12 +221,12 @@ impl PreKeyStore for &FfiPreKeyStoreStruct {
 
     async fn save_pre_key(
         &mut self,
-        prekey_id: u32,
+        prekey_id: PreKeyId,
         record: &PreKeyRecord,
         ctx: Context,
     ) -> Result<(), SignalProtocolError> {
         let ctx = ctx.unwrap_or(std::ptr::null_mut());
-        let result = (self.store_pre_key)(self.ctx, prekey_id, &*record, ctx);
+        let result = (self.store_pre_key)(self.ctx, prekey_id.into(), &*record, ctx);
 
         if let Some(error) = CallbackError::check(result) {
             return Err(SignalProtocolError::ApplicationCallbackError(
@@ -240,11 +240,11 @@ impl PreKeyStore for &FfiPreKeyStoreStruct {
 
     async fn remove_pre_key(
         &mut self,
-        prekey_id: u32,
+        prekey_id: PreKeyId,
         ctx: Context,
     ) -> Result<(), SignalProtocolError> {
         let ctx = ctx.unwrap_or(std::ptr::null_mut());
-        let result = (self.remove_pre_key)(self.ctx, prekey_id, ctx);
+        let result = (self.remove_pre_key)(self.ctx, prekey_id.into(), ctx);
 
         if let Some(error) = CallbackError::check(result) {
             return Err(SignalProtocolError::ApplicationCallbackError(
@@ -282,12 +282,12 @@ pub struct FfiSignedPreKeyStoreStruct {
 impl SignedPreKeyStore for &FfiSignedPreKeyStoreStruct {
     async fn get_signed_pre_key(
         &self,
-        prekey_id: u32,
+        prekey_id: SignedPreKeyId,
         ctx: Context,
     ) -> Result<SignedPreKeyRecord, SignalProtocolError> {
         let ctx = ctx.unwrap_or(std::ptr::null_mut());
         let mut record = std::ptr::null_mut();
-        let result = (self.load_signed_pre_key)(self.ctx, &mut record, prekey_id, ctx);
+        let result = (self.load_signed_pre_key)(self.ctx, &mut record, prekey_id.into(), ctx);
 
         if let Some(error) = CallbackError::check(result) {
             return Err(SignalProtocolError::ApplicationCallbackError(
@@ -307,12 +307,12 @@ impl SignedPreKeyStore for &FfiSignedPreKeyStoreStruct {
 
     async fn save_signed_pre_key(
         &mut self,
-        prekey_id: u32,
+        prekey_id: SignedPreKeyId,
         record: &SignedPreKeyRecord,
         ctx: Context,
     ) -> Result<(), SignalProtocolError> {
         let ctx = ctx.unwrap_or(std::ptr::null_mut());
-        let result = (self.store_signed_pre_key)(self.ctx, prekey_id, &*record, ctx);
+        let result = (self.store_signed_pre_key)(self.ctx, prekey_id.into(), &*record, ctx);
 
         if let Some(error) = CallbackError::check(result) {
             return Err(SignalProtocolError::ApplicationCallbackError(
--- a/rust/protocol/src/address.rs
+++ b/rust/protocol/src/address.rs
@@ -1,5 +1,5 @@
 //
-// Copyright 2020-2021 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
@@ -16,7 +16,26 @@ use std::fmt;
 /// represents some user.
 ///
 /// Used in [ProtocolAddress].
-pub type DeviceId = u32;
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct DeviceId(u32);
+
+impl From<u32> for DeviceId {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+
+impl From<DeviceId> for u32 {
+    fn from(value: DeviceId) -> Self {
+        value.0
+    }
+}
+
+impl fmt::Display for DeviceId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
 
 /// Represents a unique Signal client instance as `(<user ID>, <device ID>)` pair.
 #[derive(Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
--- a/rust/protocol/src/protocol.rs
+++ b/rust/protocol/src/protocol.rs
@@ -4,6 +4,7 @@
 //
 
 use crate::proto;
+use crate::state::{PreKeyId, SignedPreKeyId};
 use crate::{IdentityKey, PrivateKey, PublicKey, Result, SignalProtocolError};
 
 use std::convert::TryFrom;
@@ -236,8 +237,8 @@ impl TryFrom<&[u8]> for SignalMessage {
 pub struct PreKeySignalMessage {
     message_version: u8,
     registration_id: u32,
-    pre_key_id: Option<u32>,
-    signed_pre_key_id: u32,
+    pre_key_id: Option<PreKeyId>,
+    signed_pre_key_id: SignedPreKeyId,
     base_key: PublicKey,
     identity_key: IdentityKey,
     message: SignalMessage,
@@ -248,16 +249,16 @@ impl PreKeySignalMessage {
     pub fn new(
         message_version: u8,
         registration_id: u32,
-        pre_key_id: Option<u32>,
-        signed_pre_key_id: u32,
+        pre_key_id: Option<PreKeyId>,
+        signed_pre_key_id: SignedPreKeyId,
         base_key: PublicKey,
         identity_key: IdentityKey,
         message: SignalMessage,
     ) -> Result<Self> {
         let proto_message = proto::wire::PreKeySignalMessage {
             registration_id: Some(registration_id),
-            pre_key_id,
-            signed_pre_key_id: Some(signed_pre_key_id),
+            pre_key_id: pre_key_id.map(|id| id.into()),
+            signed_pre_key_id: Some(signed_pre_key_id.into()),
             base_key: Some(base_key.serialize().into_vec()),
             identity_key: Some(identity_key.serialize().into_vec()),
             message: Some(Vec::from(message.as_ref())),
@@ -291,12 +292,12 @@ impl PreKeySignalMessage {
     }
 
     #[inline]
-    pub fn pre_key_id(&self) -> Option<u32> {
+    pub fn pre_key_id(&self) -> Option<PreKeyId> {
         self.pre_key_id
     }
 
     #[inline]
-    pub fn signed_pre_key_id(&self) -> u32 {
+    pub fn signed_pre_key_id(&self) -> SignedPreKeyId {
         self.signed_pre_key_id
     }
 
@@ -368,8 +369,8 @@ impl TryFrom<&[u8]> for PreKeySignalMessage {
         Ok(PreKeySignalMessage {
             message_version,
             registration_id: proto_structure.registration_id.unwrap_or(0),
-            pre_key_id: proto_structure.pre_key_id,
-            signed_pre_key_id,
+            pre_key_id: proto_structure.pre_key_id.map(|id| id.into()),
+            signed_pre_key_id: signed_pre_key_id.into(),
             base_key,
             identity_key: IdentityKey::try_from(identity_key.as_ref())?,
             message: SignalMessage::try_from(message.as_ref())?,
@@ -906,7 +907,7 @@ mod tests {
             3,
             365,
             None,
-            97,
+            97.into(),
             base_key_pair.public_key,
             identity_key_pair.public_key.into(),
             message,
@@ -1016,7 +1017,7 @@ mod tests {
             3,
             365,
             None,
-            97,
+            97.into(),
             base_key_pair.public_key,
             identity_key_pair.public_key.into(),
             message,
--- a/rust/protocol/src/session_cipher.rs
+++ b/rust/protocol/src/session_cipher.rs
@@ -1,5 +1,5 @@
 //
-// Copyright 2020-2021 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
--- a/rust/protocol/src/state/prekey.rs
+++ b/rust/protocol/src/state/prekey.rs
@@ -1,13 +1,36 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
 use crate::proto::storage::PreKeyRecordStructure;
 use crate::{KeyPair, PrivateKey, PublicKey, Result, SignalProtocolError};
+
 use prost::Message;
 
-pub type PreKeyId = u32;
+use std::fmt;
+
+/// A unique identifier selecting among this client's known pre-keys.
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub struct PreKeyId(u32);
+
+impl From<u32> for PreKeyId {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+
+impl From<PreKeyId> for u32 {
+    fn from(value: PreKeyId) -> Self {
+        value.0
+    }
+}
+
+impl fmt::Display for PreKeyId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
 
 #[derive(Debug, Clone)]
 pub struct PreKeyRecord {
@@ -20,7 +43,7 @@ impl PreKeyRecord {
         let private_key = key.private_key.serialize().to_vec();
         Self {
             pre_key: PreKeyRecordStructure {
-                id,
+                id: id.into(),
                 public_key,
                 private_key,
             },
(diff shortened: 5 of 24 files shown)


rust/bridge/shared/src/ffi/storage.rs
```diff
@@ -77,7 +77,10 @@ impl IdentityKeyStore for &FfiIdentityKeyStoreStruct {
         Ok(IdentityKeyPair::new(IdentityKey::new(pub_key), *priv_key))
     }
 
-    async fn get_local_registration_id(&self, ctx: Context) -> Result<u32, SignalProtocolError> {
+    async fn get_local_registration_id(
+        &self,
+        ctx: Context,
+    ) -> Result<DeviceId, SignalProtocolError> {
```
jrose-signal, line 83: This is *not* a device ID; it's a locally-generated random number used for pre-keys / constructing the initial session.

rust/bridge/shared/src/ffi/storage.rs
```diff
@@ -194,12 +197,12 @@ pub struct FfiPreKeyStoreStruct {
 impl PreKeyStore for &FfiPreKeyStoreStruct {
     async fn get_pre_key(
         &self,
-        prekey_id: u32,
+        prekey_id: DeviceId,
```
jrose-signal, line 200: This is *also* not a device ID; it's a unique identifier selecting among this device's pre-keys.

rust/bridge/shared/src/ffi/storage.rs
```diff
@@ -279,12 +282,12 @@ pub struct FfiSignedPreKeyStoreStruct {
 impl SignedPreKeyStore for &FfiSignedPreKeyStoreStruct {
     async fn get_signed_pre_key(
         &self,
-        prekey_id: u32,
+        prekey_id: DeviceId,
```
jrose-signal, line 285: Same thing, with a different set of identifiers for signed pre-keys.

rust/protocol/src/address.rs
```diff
@@ -1,10 +1,38 @@
...
+impl From<u32> for DeviceId {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+
+impl From<DeviceId> for u32 {
+    fn from(value: DeviceId) -> Self {
+        let DeviceId(value) = value;
+        value
```
jrose-signal, line 25: I'd tend to go with `value.0`; is there a reason to prefer the pattern-matching syntax?

jrose-signal (commented):
I think you've made a very strong case for why strong types are a good idea. ;-)

rust/bridge/shared/src/ffi/storage.rs
cosmicexplorer: Thanks! I have used that phrasing in the documentation for the new `SessionSeed` struct.

rust/bridge/shared/src/ffi/storage.rs
cosmicexplorer: Fixed!

rust/protocol/src/address.rs
cosmicexplorer: Not at all! Fixed!

rust/bridge/shared/src/ffi/storage.rs
cosmicexplorer: Noted in the documentation for `PreKeyId`!

cosmicexplorer:
@jrose-signal I've split up this change into three logically separate commits to make it easier to review.

rust/bridge/shared/src/ffi/storage.rs
```diff
@@ -77,7 +77,10 @@ impl IdentityKeyStore for &FfiIdentityKeyStoreStruct {
         Ok(IdentityKeyPair::new(IdentityKey::new(pub_key), *priv_key))
     }
 
-    async fn get_local_registration_id(&self, ctx: Context) -> Result<u32, SignalProtocolError> {
+    async fn get_local_registration_id(
+        &self,
+        ctx: Context,
+    ) -> Result<SessionSeed, SignalProtocolError> {
```
jrose-signal, line 83: I'm not sure about the name "SessionSeed". That's how it's used, sure, but it already has a name: "RegistrationId".

rust/protocol/src/kdf.rs
```diff
@@ -16,17 +18,18 @@ pub struct HKDF {
...
                 iteration_start_offset: 0,
             }),
-            3 => Ok(HKDF {
+            MessageVersion::Version3 => Ok(HKDF {
                 iteration_start_offset: 1,
             }),
-            _ => Err(SignalProtocolError::UnrecognizedMessageVersion(
-                message_version,
-            )),
         }
```
jrose-signal, line 33: I think there's more going on here than just the message version at this point, since we also use KDF for other functionality besides just the session state. In addition, we don't actually support any message versions but 3 in this library (and the clients don't use HKDF for "other functionality" with any version but 3), so generalizing over message versions isn't buying us anything at the moment. Finally, we wouldn't want adding a new message version to change the derived keys for that "other functionality", so I'd consider a no-args constructor an anti-pattern here.

All of this suggests a redesign, either to have a proper enum here that isn't strictly about message versions, or to only support version 3 here and drop the generality. Both of those should wait for another PR.

rust/protocol/src/protocol.rs
```diff
@@ -17,6 +20,44 @@ use uuid::Uuid;
...
 pub const CIPHERTEXT_MESSAGE_CURRENT_VERSION: u8 = 3;
 
+/// A [u8] describing the version of the message chain format to use when starting a chain.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, num_enum::TryFromPrimitive, num_enum::IntoPrimitive)]
+#[repr(u8)]
+pub enum MessageVersion {
+    Version2 = 2,
+    /// **\[CURRENT\]**.
+    Version3 = CIPHERTEXT_MESSAGE_CURRENT_VERSION,
+}
```
jrose-signal, line 30: I think this would go the other way around: the constant should have enum type now. (Also, I'd just spell these as "V2" and "V3"; the enum name already has "Version" spelled out.)

rust/protocol/src/protocol.rs
```diff
@@ -17,6 +20,44 @@ use uuid::Uuid;
...
+
+impl Default for MessageVersion {
+    fn default() -> Self {
+        Self::Version3
+    }
+}
```
jrose-signal, line 36: This should not have a default, since it's something that ends up in a wire format.

rust/protocol/src/protocol.rs
```diff
@@ -358,10 +401,10 @@ impl TryFrom<&[u8]> for PreKeySignalMessage {
         let base_key = PublicKey::deserialize(base_key.as_ref())?;
 
         Ok(PreKeySignalMessage {
-            message_version,
-            registration_id: proto_structure.registration_id.unwrap_or(0),
-            pre_key_id: proto_structure.pre_key_id,
-            signed_pre_key_id,
+            message_version: message_version.try_into()?,
+            registration_id: (proto_structure.registration_id.unwrap_or(0) as u32).into(),
```
jrose-signal, line 405: Why `as u32` here? Shouldn't `proto_structure.registration_id` already have the right type?

rust/protocol/src/protocol.rs
```diff
@@ -358,10 +401,10 @@ impl TryFrom<&[u8]> for PreKeySignalMessage {
         let base_key = PublicKey::deserialize(base_key.as_ref())?;
 
         Ok(PreKeySignalMessage {
-            message_version,
-            registration_id: proto_structure.registration_id.unwrap_or(0),
-            pre_key_id: proto_structure.pre_key_id,
-            signed_pre_key_id,
+            message_version: message_version.try_into()?,
```
jrose-signal, line 404: Hmm, we've already checked the version at this point, so this should probably use `expect()` instead of propagating the error.

rust/protocol/src/state/session.rs
```diff
@@ -390,8 +392,9 @@ impl SessionState {
         signed_pre_key_id: SignedPreKeyId,
         base_key: &PublicKey,
     ) -> Result<()> {
+        let signed_pre_key_id: u32 = signed_pre_key_id.into();
         let pending = session_structure::PendingPreKey {
-            pre_key_id: pre_key_id.unwrap_or(0),
+            pre_key_id: pre_key_id.unwrap_or(0.into()).into(),
```
jrose-signal, line 397: Maybe `pre_key_id.map(|id| id.into()).unwrap_or(0)`? I know it's both equivalent and a little longer, but it does make it clear that "0" is a placeholder rather than a real pre-key with ID #0.

rust/protocol/tests/session.rs
```diff
@@ -37,10 +37,10 @@ fn test_basic_prekey_v3() -> Result<(), SignalProtocolError> {
...
         let bob_pre_key_bundle = PreKeyBundle::new(
-            bob_store.get_local_registration_id(None).await?,
-            1,                                               // device id
-            Some((pre_key_id, bob_pre_key_pair.public_key)), // pre key
-            signed_pre_key_id,                               // signed pre key id
+            bob_store.get_local_registration_id(None).await?.into(),
+            1,                                                      // device id
+            Some((pre_key_id.into(), bob_pre_key_pair.public_key)), // pre key
+            signed_pre_key_id.into(),                               // signed pre key id
             bob_signed_pre_key_pair.public_key,
```
jrose-signal, line 44: Why not use strong types here too?

jrose-signal (commented):
Looks like the split up commits aren't quite split up right anyway, so I just reviewed this all together.

jrose-signal:
I double-checked and the message versions for SenderKeyMessage and SenderKeyDistributionMessage should be the same, as should SignalMessage and PreKeySignalMessage, but they are two different message families and so they should not be the same as each other. I'll put in a PR to record the SK(D)M version in the stored state like we do for 1-1 sessions.

jrose-signal:
(That was #293.)

rust/bridge/shared/src/ffi/storage.rs
cosmicexplorer: Fixed!

rust/protocol/src/protocol.rs
cosmicexplorer: You're right -- great catch. This can just use `.into()` instead.

rust/protocol/src/protocol.rs
cosmicexplorer: It turns out that it does!

rust/protocol/src/state/session.rs
cosmicexplorer: That makes much more sense!

rust/protocol/tests/session.rs
cosmicexplorer: Great idea!

rust/protocol/src/protocol.rs
cosmicexplorer: Removed!

rust/protocol/src/protocol.rs
cosmicexplorer: Renamed to `V2` and `V3`, and converted the constant to be enum type!

rust/protocol/tests/session.rs
cosmicexplorer: Done!

rust/protocol/src/kdf.rs
cosmicexplorer: Ok, I've interpreted this as:
1. remove the no-arg constructor `HKDF::new()` (done).
1. split off a separate PR to remove the `MessageVersion` enum.

Is that correct?

cosmicexplorer:
Thanks so much for fixing the `SenderKeyState` dissonance in #293! In another branch I haven't pushed I had been wondering what to do with that.

rust/protocol/src/error.rs
```diff
@@ -1,10 +1,11 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020, 2021 Signal Messenger, LLC.
```
jrose-signal, line 2: Nitpick: `2020-2021`

rust/protocol/src/protocol.rs
```diff
@@ -4,19 +4,63 @@
...
 use prost::Message;
 use rand::{CryptoRng, Rng};
 use sha2::Sha256;
 use subtle::ConstantTimeEq;
 use uuid::Uuid;
 
-pub const CIPHERTEXT_MESSAGE_CURRENT_VERSION: u8 = 3;
-pub const SENDERKEY_MESSAGE_CURRENT_VERSION: u8 = 3;
+pub const CIPHERTEXT_MESSAGE_CURRENT_VERSION: MessageVersion = MessageVersion::V3;
+pub const SENDERKEY_MESSAGE_CURRENT_VERSION: MessageVersion = MessageVersion::V3;
```
jrose-signal, line 21: These should (sadly) not be the same type, as they already don't represent the same kind of message or session (even if they share an HKDF). Honestly given that message version 2 isn't supported I'm not sure what to think about `HKDF::new`.

rust/protocol/src/state/signed_prekey.rs
```diff
@@ -1,28 +1,63 @@
...
+
+impl From<SignedPreKeyId> for u32 {
+    fn from(value: SignedPreKeyId) -> Self {
+        value.0
+    }
+}
+
+/// A timestamp assigned when creating a [SignedPreKeyRecord].
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct SignedPreKeyTimestamp(pub u64);
```
jrose-signal, line 28: Let's hold off on this one. There's a general consistent "timestamp" type throughout all these APIs which is "milliseconds since 1970", and we can do them all together.

rust/protocol/src/protocol.rs
```diff
@@ -231,10 +277,10 @@ impl TryFrom<&[u8]> for SignalMessage {
 
 #[derive(Debug, Clone)]
 pub struct PreKeySignalMessage {
-    message_version: u8,
-    registration_id: u32,
-    pre_key_id: Option<u32>,
-    signed_pre_key_id: u32,
+    message_version: MessageVersion,
+    registration_id: DeviceId,
```
jrose-signal, line 281: RegistrationId

rust/protocol/src/storage/traits.rs
```diff
@@ -20,11 +20,27 @@ pub enum Direction {
     Receiving,
 }
 
+/// A locally-generated random number used to construct the initial value of a message chain.
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct RegistrationId(pub u32);
```
jrose-signal, line 25: These should not be `Ord`, as they are generally randomly generated.

For bonus points later we should actually make this a u16, because that's what the server expects, but that'll mess up all the bridges, so let's leave it for this PR.

jrose-signal (commented):
This is looking pretty good, but let's leave out some of the more complicated ones: MessageVersion (and the HKDF changes), and SignedPreKeyTimestamp

rust/protocol/src/state/signed_prekey.rs
cosmicexplorer: Ok, done!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Fixed! Before reading this comment I was actually able to address the bridge part anyway by producing `unsafe_*` extractor functions in 89a79bb.

rust/protocol/src/protocol.rs
cosmicexplorer: I have reverted this change and isolated it into https://github.com/cosmicexplorer/libsignal-client/commit/a29e6d3b0ed1f4dd5ce5c4944b90accab8c3b1a9 which I will push in a followup PR!

cosmicexplorer:
Ok, I have responded to all of your comments and rebased the result into 4 logically independent commits for easier review!

cosmicexplorer:
Was having lots of difficulty making the bridge tests pass and would probably like your assistance on those changes independently, so have reverted the `RegistrationId` changes and stuffed them into https://github.com/cosmicexplorer/libsignal-client/compare/type-safety-over-ambiguously-named-fields...TYPE-SAFETY-V2?expand=1 for a followup PR.

rust/protocol/src/address.rs
```diff
@@ -16,7 +16,26 @@ use std::fmt;
 /// represents some user.
 ///
 /// Used in [ProtocolAddress].
-pub type DeviceId = u32;
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct DeviceId(pub u32);
```
jrose-signal, line 20: We talked about this in one of our other repos, and came to the conclusion that it was better to make the field non-`pub`, so that call sites uniformly use `into()` to go back to a u32. (As you already have here.)

rust/protocol/src/state/prekey.rs
```diff
@@ -7,7 +7,21 @@ use crate::proto::storage::PreKeyRecordStructure;
 use crate::{KeyPair, PrivateKey, PublicKey, Result};
 use prost::Message;
 
-pub type PreKeyId = u32;
+/// A unique identifier selecting among this client's known pre-keys.
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct PreKeyId(pub u32);
```
jrose-signal, line 12: Same as DeviceId about `pub`, and also I think this one doesn't need Ord/PartialOrd.

rust/protocol/src/state/session.rs
jrose-signal: Looks like this missed the rebase.

rust/protocol/src/state/signed_prekey.rs
```diff
@@ -7,7 +7,21 @@ use crate::proto::storage::SignedPreKeyRecordStructure;
 use crate::{KeyPair, PrivateKey, PublicKey, Result};
 use prost::Message;
 
-pub type SignedPreKeyId = u32;
+/// A unique identifier selecting among this client's known signed pre-keys.
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct SignedPreKeyId(pub u32);
```
jrose-signal, line 12: Ditto PreKeyId

rust/protocol/tests/groups.rs
```diff
@@ -86,10 +86,10 @@ fn group_using_context_arg() -> Result<(), SignalProtocolError> {
     async {
         let mut csprng = OsRng;
 
-        let sender_address = ProtocolAddress::new("+14159999111".to_owned(), 1);
+        let sender_address = ProtocolAddress::new("+14159999111".to_owned(), 1.into());
         let distribution_id = Uuid::from_u128(0xd1d1d1d1_7000_11eb_b32a_33b8a8a487a6);
 
-        let x = Box::new(1);
+        let x: Box<DeviceId> = Box::new(1.into());
```
jrose-signal, line 92: This isn't a device ID, just a placeholder to test context.

jrose-signal (commented):
This one's pretty straightforward, so I picked it up first. Only a few comments.

rust/protocol/tests/groups.rs
cosmicexplorer: Thanks--reverted!

rust/protocol/src/state/signed_prekey.rs
cosmicexplorer: Done!

rust/protocol/src/state/prekey.rs
cosmicexplorer: Done!

rust/protocol/src/address.rs
cosmicexplorer: Done!

rust/protocol/src/state/session.rs
jrose-signal: Still shows up in at least one other spot!

rust/protocol/src/session_cipher.rs
```diff
@@ -62,7 +62,7 @@ pub async fn message_encrypt(
             remote_address,
             items
                 .pre_key_id()
-                .map_or_else(|| "<none>".to_string(), |id| id.to_string())
+                .map_or_else(|| "<none>".to_string(), |id| format!("{:?}", id))
```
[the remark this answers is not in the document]
cosmicexplorer: I've implemented `std::fmt::Display` for these structs and have converted the `format!()` call to use `"{}"`.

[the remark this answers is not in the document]
jrose-signal: …which can be spelled `to_string()` as before. :-)

rust/protocol/src/state/session.rs
```diff
@@ -401,8 +401,9 @@ impl SessionState {
         signed_pre_key_id: SignedPreKeyId,
         base_key: &PublicKey,
     ) {
+        let signed_pre_key_id: u32 = signed_pre_key_id.into();
         let pending = session_structure::PendingPreKey {
-            pre_key_id: pre_key_id.unwrap_or(0),
+            pre_key_id: pre_key_id.unwrap_or_else(|| 0.into()).into(),
```
jrose-signal, line 406: Here's another one where `.map(PreKeyId::into).unwrap_or(0)` would read a little better.

cosmicexplorer: Fixed!

cosmicexplorer:
Ok, one of the CI runs is failing reliably, and I can't reproduce it locally. I'm running `cargo +stable test --workspace --verbose -- --include-ignored`, but that command succeeds for me. It doesn't look like this test is failing on `main`, so I'm wondering what I'm doing wrong. The beginning of the errors in https://github.com/signalapp/libsignal/runs/7399386113?check_suite_focus=true look like:
```rust
 ---- src/support/mod.rs - support::bridge_get (line 162) stdout ----
error: cannot find macro `bridge_get` in this scope
  --> src/support/mod.rs:172:1
   |
12 | bridge_get!(Foo::bar as GetBar -> &str, ffi = "foo_get_bar", jni = "Foo_GetBar", node = "Foo_GetBar");
   | ^^^^^^^^^^
```

cosmicexplorer:
Hm, it looks like that exact failure is happening for all the other PRs I have open right now as well. It seems like the CI might be wonky.

jrose-signal:
Cherry-picked as ec3c2d32bcf027ed1766e5bdc91cdd3f16a8e5ba, will be in the next release!

PR #296, opened 2021-05-09
add `thiserror` crate to derive `fmt::Display` for `SignalProtocolError`

cosmicexplorer:
### Problem

I looked into @jrose-signal's comment https://github.com/signalapp/libsignal-client/pull/287#discussion_r627650292 comparing the use of `thiserror` vs adding real docstrings as in #287. I believe they are totally orthogonal.

With that in mind, it looks like [`thiserror`](https://docs.rs/thiserror/1.0.24/thiserror/) might be able to just save some of our boilerplate in `error.rs`.

### Solution
1. import `thiserror::Error` and add `#[derive(Error)]` to `SignalProtocolError`.
    - deriving `Error` neatly required the introduction of a small workaround `CallbackErrorWrapper`.
1. Add `#[error(...)]` annotations to every case of `SignalProtocolError` to auto-derive `fmt::Display`.

### Result
Our `fmt::Display` implementation for `SignalProtocolError` requires less code because we can rely on a macro to derive it! And because we derive `Error` as well, we can now propagate error backtraces and display them with a crate like [`eyre`](https://docs.rs/eyre/0.6.5/eyre/) in the future!


simonsan:
https://crates.io/crates/displaydoc might come handy as well. ;)

jrose-signal:
Right, we've previously had a PR to switch to `thiserror` and decided that the boilerplate savings weren't worth the additional dependency and DSL. `displaydoc` pushes my thoughts back in the other direction because they become (sparse) doc comments as well.

cosmicexplorer:
Got it! I'll try out displaydoc! Might steal from #287!

cosmicexplorer:
Ok, redone to use `displaydoc`! Note that we still depend on `thiserror` for the `Error` derive macro with the `#[from]` and `#[source]` annotations.

rust/protocol/src/error.rs
```diff
@@ -5,186 +5,120 @@
...
+use thiserror::Error;
+
 use std::fmt;
 use std::panic::UnwindSafe;
 
 pub type Result<T> = std::result::Result<T, SignalProtocolError>;
 
+/// Wraps a boxed error struct and delegates the [std::error::Error] trait to it.
+///
+/// A [Box] wrapping an error apparently does not implement [std::error::Error] itself, which breaks
```
jrose-signal, line 18: Hm, [that implementation is in the stdlib](https://doc.rust-lang.org/std/error/trait.Error.html#impl-Error-39). Something else must be going on.

cosmicexplorer: Unfortunately, there is a distinction between `Box<E: Error>` and `Box<dyn Error>`. I have added this explanatory link to the docstring: https://stackoverflow.com/a/62452516/2518889.

cosmicexplorer: I believe we could work around this by adding a type parameter `E: Error` to the definition of `SignalProtocolError`, but that seems more annoying than wrapping in `CallbackErrorWrapper`.

rust/protocol/src/error.rs
```diff
@@ -1,198 +1,131 @@
...
+///
+/// Unlike a `Box<E: Error>` wrapping a concrete error type, a `Box<dyn Error>` wrapping
+/// a type-erased trait object does *not* implement [std::error::Error] itself (see [this
+/// stackoverflow answer] for details). This breaks the ability to nicely use the `#[source]`
+/// annotation from the [thiserror::Error] derive macro. In order to avoid manually impling
+/// [std::error::Error::source] for each case of [SignalProtocolError], we create this wrapper
+/// struct to delegate to the underlying `dyn Error` in order to nicely derive [thiserror::Error]
+/// for [SignalProtocolError::ApplicationCallbackError].
+///
+/// [this stackoverflow answer]: https://stackoverflow.com/a/62452516/2518889
```
jrose-signal, line 26: I don't think the resulting code is clearer than just implementing `source` manually (and therefore Error). In the choice between making one place a lot more complicated and many places a little more complicated, I'd usually err on the side of containing the complexity. ([If only we didn't need UnwindSafe here.](https://github.com/dtolnay/thiserror/pull/41))

PR #297, opened 2021-05-10
[Experiment]: Required Send in store traits implementation with a feature flag

gferon:
See https://github.com/signalapp/libsignal-client/issues/298

### DIFF
--- a/rust/bridge/ffi/src/lib.rs
+++ b/rust/bridge/ffi/src/lib.rs
@@ -134,7 +134,6 @@ pub unsafe extern "C" fn signal_sealed_session_cipher_decrypt(
             &mut session_store,
             &mut prekey_store,
             &mut signed_prekey_store,
-            Some(ctx),
         ))?;
 
         write_optional_cstr_to(sender_e164, Ok(decrypted.sender_e164))?;
--- a/rust/bridge/shared/src/ffi/storage.rs
+++ b/rust/bridge/shared/src/ffi/storage.rs
@@ -42,6 +42,7 @@ pub enum FfiDirection {
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct FfiIdentityKeyStoreStruct {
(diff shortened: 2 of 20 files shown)


lushing:
I would say that the solution presented here feels pretty reasonable. @jrose-signal do you have any particular use case in mind for keeping the context around as a parameter to all these async APIs? Maybe this way deviates from the norms a bit, but it seems like a more pragmatic choice.

lushing:
Hi @jrose-signal, if you have a particular preference among the potential solutions you presented, do tell, as I am more than willing to help implement whichever solution that can be included upstream.

jrose-signal:
I feel like iOS's use case of passing a transaction through is a reasonable one, and that we're more likely to want to expose that kind of thing to Java and TypeScript in the future than to have them forever be different. Of course, Rust supports zero-cost wrappers, but I still don't think that leads to a good experience: you'd have to write a wrapper type to carry the extra information and forward the trait methods, and then construct it at the call sites:

```rust
message_encrypt(
  message,
  destination,
  SessionStoreWrapper::new(sessionStore, transaction),
  IdentityKeyStoreWrapper::new(identityStore, transaction))
```

as opposed to

```rust
message_encrypt(
  message,
  destination,
  sessionStore,
  identityStore,
  transaction as StoreContext)
```

So I think I like the idea of Context becoming a marker for `Any + Send`, and the *specific* Context used by FFI to be a wrapper type around a pointer with `unsafe impl Send`. This is okay by this description from [the Rustonomicon](https://doc.rust-lang.org/nomicon/send-and-sync.html), not to mention how they'll be used in practice:

> However raw pointers are, strictly speaking, marked as thread-unsafe as more of a *lint.* Doing anything useful with a raw pointer requires dereferencing it, which is already unsafe. In that sense, one could argue that it would be "fine" for them to be marked as thread safe.

PR #304, opened 2021-05-17
Add registration IDs to the Sealed Sender v2 upload (encrypt) format

jrose-signal:
Registration IDs are used to detect if a device ID has been reused, since the new device will (with high probability) use a different randomly-generated registration ID from the old one. The server should be able to validate this for SSv2 like it does for SSv1, though the handling of this for SSv1 is in the various apps.

Replaces #302, which I clicked "merge" on by mistake.


jrose-signal:
Closing in favor of a quick stopgap to match (private) changes in the server logic, followed by a better client API.

PR #308, opened 2021-05-19
Node: Switch to our fork of Neon with faster EventQueue operations

jrose-signal:
See [signalapp/neon](/signalapp/neon) and neon-bindings/neon#739.

### DIFF
--- a/rust/bridge/node/Cargo.toml
+++ b/rust/bridge/node/Cargo.toml
@@ -18,7 +18,7 @@ crate-type = ["cdylib"]
 libsignal-protocol = { path = "../../protocol" }
 libsignal-bridge = { path = "../shared", features = ["node"] }
 signal-neon-futures = { path = "futures" }
-neon = { version = "0.8", default-features = false, features = ["napi-4", "event-queue-api"] }
+neon = { version = "0.8", default-features = false, features = ["napi-6", "event-queue-api"] }
 rand = "0.7.3"
 log = "0.4"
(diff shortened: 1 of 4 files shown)


rust/bridge/node/Cargo.toml
```diff
@@ -18,7 +18,7 @@ crate-type = ["cdylib"]
 libsignal-protocol = { path = "../../protocol" }
 libsignal-bridge = { path = "../shared", features = ["node"] }
 signal-neon-futures = { path = "futures" }
-neon = { version = "0.8", default-features = false, features = ["napi-4", "event-queue-api"] }
+neon = { version = "0.8", default-features = false, features = ["napi-6", "event-queue-api"] }
```
indutny-signal, line 21: Should it point to git as well?

indutny-signal (approved):
I'm not entirely familiar with the crate resolution system, but shouldn't all neon references point to git? Also, there are bunch of other `Cargo.toml` that have `neon`. Should they be updated?

Given my lack of expertise this LGTM, though, provided that it builds.

jrose-signal:
The workspace-level Cargo.toml serves as an override for everything in the workspace (though not for someone outside the workspace depending on one of these crates). The same mechanism is used for our forked curve25519-dalek.

As for features, they're additive, so this effectively says "you can build signal-neon-futures (and libsignal-bridge) not depending on napi-4, but when you actually build the libsignal-node that gets shipped you should have napi-6 enabled".

PR #309, opened 2021-05-20
Get registration IDs from sessions for Sealed Sender v2

jrose-signal:
The app-visible change is that `sealedSenderMultiRecipientEncrypt` now takes a SessionStore as well. Sessions will be looked up in bulk using a new SessionStore API, `loadExistingSessions` or `getExistingSessions` (just slightly different names across the different app languages). The registration ID is then loaded from each session and included in the resulting SSv2 payload.

The implementation is a bit of a divergence from some other APIs in libsignal-client in that the "look up in bulk" step is performed in the Java, Swift, or TypeScript layer, with the resulting sessions passed down to Rust. Why? Because otherwise we'd pass a list of addresses into Rust, which would have to turn them back into a Java, Swift, or TypeScript array to call the SessionStore method. This would be (1) a bunch of extra work to implement, and (2) a waste of CPU when we already *have* a list of addresses in the correct format: the argument to `sealedSenderMultiRecipientEncrypt`.

This is an example of "the boundaries between the Rust and Java/Swift/TypeScript parts of the library don't have to be perfect; they're internal to the overall product". In this case, we've taken that a little further than usual: usually we try to make the libsignal-protocol API as convenient as possible as well, but here it had to be a bit lower-level to satisfy the needs of the app language wrappers. (Specifically, Rust callers need to fetch the list of SessionRecords themselves.)

P.S. Why doesn't v1 of sealed sender include registration IDs? Because for SSv1, libsignal-client isn't producing the entire request body to upload to the server; it's only producing the message content that will be decrypted by the recipient. With SSv2, the serialized message the recipient downloads has both shared and per-recipient data in it, which the server must assemble from the uploaded request. Because of this, SSv2's encrypt API might as well produce the entire request.

### DIFF
--- a/java/java/src/main/java/org/whispersystems/libsignal/state/impl/InMemorySessionStore.java
+++ b/java/java/src/main/java/org/whispersystems/libsignal/state/impl/InMemorySessionStore.java
@@ -5,6 +5,7 @@
  */
 package org.whispersystems.libsignal.state.impl;
 
+import org.whispersystems.libsignal.NoSessionException;
 import org.whispersystems.libsignal.SignalProtocolAddress;
 import org.whispersystems.libsignal.state.SessionRecord;
 import org.whispersystems.libsignal.state.SessionStore;
@@ -32,6 +33,23 @@ public class InMemorySessionStore implements SessionStore {
     } catch (IOException e) {
       throw new AssertionError(e);
     }
+  }
+
+  @Override
+  public synchronized List<SessionRecord> loadExistingSessions(List<SignalProtocolAddress> addresses) throws NoSessionException {
+    List<SessionRecord> resultSessions = new LinkedList<>();
+    for (SignalProtocolAddress remoteAddress : addresses) {
+      byte[] serialized = sessions.get(remoteAddress);
+      if (serialized == null) {
+        throw new NoSessionException("no session for " + remoteAddress);
+      }
+      try {
+        resultSessions.add(new SessionRecord(serialized));
+      } catch (IOException e) {
+        throw new AssertionError(e);
+      }
+    }
+    return resultSessions;
   }
 
   @Override
--- a/node/test/PublicAPITest.ts
+++ b/node/test/PublicAPITest.ts
@@ -21,7 +21,7 @@ SignalClient.initLogger(
 );
 
 class InMemorySessionStore extends SignalClient.SessionStore {
-  private state = new Map();
+  private state = new Map<string, Buffer>();
   async saveSession(
     name: SignalClient.ProtocolAddress,
     record: SignalClient.SessionRecord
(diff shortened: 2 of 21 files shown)


java/java/src/main/java/org/whispersystems/libsignal/state/impl/InMemorySessionStore.java
```diff
@@ -34,6 +35,23 @@ public synchronized SessionRecord loadSession(SignalProtocolAddress remoteAddres
     }
   }
 
+  @Override
+  public synchronized List<SessionRecord> loadExistingSessions(List<SignalProtocolAddress> addresses) throws NoSessionException {
+    List<SessionRecord> resultSessions = new LinkedList<>();
+    for (SignalProtocolAddress remoteAddress : addresses) {
+      byte[] serialized = sessions.get(remoteAddress);
+      if (serialized == null) {
+        throw new NoSessionException("no session for " + remoteAddress);
```
jim-signal, line 44: It would be considered a bad state if one of the remote clients was missing a session, and would prevent loading any of the sessions? It is all-or-nothing?

node/test/PublicAPITest.ts
```diff
@@ -33,14 +33,27 @@ class InMemorySessionStore extends SignalClient.SessionStore {
...
     }
   }
+  async getExistingSessions(
+    addresses: SignalClient.ProtocolAddress[]
+  ): Promise<SignalClient.SessionRecord[]> {
+    return addresses.map(address => {
+      const idx = address.name() + '::' + address.deviceId();
+      const serialized = this.state.get(idx);
+      if (!serialized) {
+        throw 'no such session';
```
jim-signal, line 52: Should the error be consistent with java "no session for " + remoteAddress?

jim-signal (approved):
Looks good. Just a couple informational questions.

java/java/src/main/java/org/whispersystems/libsignal/state/impl/InMemorySessionStore.java
jrose-signal: Yeah, this API is meant for the fast path of "I already have sessions with all these devices", something we can be assured of because the payload is either a SignalMessage or PreKeySignalMessage (which came from a 1:1 session), or a SenderKeyMessage (which must have previously had its key sent in a 1:1 distribution message). That is, if apps are hitting this exception, there's a bug.

node/test/PublicAPITest.ts
jrose-signal: I guess. Unlike the Java one, this in-memory implementation is only exposed to tests, but it certainly doesn't hurt to have more descriptive error messages in tests.

PR #313, opened 2021-05-20
Node: make sealedSenderDecryptToUsmc synchronous

jrose-signal:
...by requiring `IdentityKeyStore.getIdentityKey` and `getLocalRegistrationId` to be synchronous.

Because of the way libsignal-client is built, this change propagates to the other platforms' implementations, but the interface doesn't change there.

In theory we could implement this at the Rust level by having a "synchronous IdentityKeyStore trait" and the regular IdentityKeyStore trait inherit from it, but that's a lot of trouble for the single accessor `sealed_sender_decrypt_to_usmc` was using unconditionally.

### DIFF
--- a/java/java/src/main/java/org/signal/internal/Native.java
+++ b/java/java/src/main/java/org/signal/internal/Native.java
@@ -176,7 +176,7 @@ public final class Native {
 
   public static native boolean ScannableFingerprint_Compare(byte[] fprint1, byte[] fprint2);
 
-  public static native long SealedSessionCipher_DecryptToUsmc(byte[] ctext, IdentityKeyStore identityStore, Object ctx);
+  public static native long SealedSessionCipher_DecryptToUsmc(byte[] ctext, long identityPrivate, long identityPublic);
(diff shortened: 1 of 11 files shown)


droidben:
May I, out of curiosity, ask what the task 'sealedSenderDecryptToUsmc' does?

jrose-signal:
"USMC" is "UnidentifiedSenderMessageContent", the structure inside a sealed-sender message. Decrypting to a USMC lets you see the sender (and verify their certificate), but the actual message content is still encrypted with either the Double Ratchet protocol or SenderKey-based encryption. This is in contrast to decrypting all the way to message plaintext, which shouldn't be done synchronously since it may require database lookups.

https://github.com/signalapp/libsignal-client/blob/72ba4e6959d80b4f091fade5373e5d9aae966c01/rust/protocol/src/proto/sealed_sender.proto#L36-L57

jrose-signal:
This PR is still in draft status because it was just a proof-of-concept; it doesn't seem like it'd be enough of a win for Desktop to push it through at the moment.

PR #315, opened 2021-05-26
Add DecryptionErrorMessage and PlaintextContent

jrose-signal:
These will be used to help reset 1:1 and SenderKey sessions when messages fail to decrypt, rather than ending up in a state where message after message fails to decrypt. (1:1 sessions already have a form of automatic session reset in recent releases of Signal, but this provides a bit more feedback.)

### DIFF
--- a/java/tests/src/test/java/org/signal/libsignal/metadata/SealedSessionCipherTest.java
+++ b/java/tests/src/test/java/org/signal/libsignal/metadata/SealedSessionCipherTest.java
@@ -14,6 +14,7 @@ import org.whispersystems.libsignal.InvalidMessageException;
 import org.whispersystems.libsignal.LegacyMessageException;
 import org.whispersystems.libsignal.NoSessionException;
 import org.whispersystems.libsignal.SessionBuilder;
+import org.whispersystems.libsignal.SessionCipher;
 import org.whispersystems.libsignal.SignalProtocolAddress;
 import org.whispersystems.libsignal.UntrustedIdentityException;
 import org.whispersystems.libsignal.ecc.Curve;
@@ -22,6 +23,8 @@ import org.whispersystems.libsignal.ecc.ECPublicKey;
 import org.whispersystems.libsignal.groups.GroupCipher;
 import org.whispersystems.libsignal.groups.GroupSessionBuilder;
 import org.whispersystems.libsignal.protocol.CiphertextMessage;
+import org.whispersystems.libsignal.protocol.DecryptionErrorMessage;
+import org.whispersystems.libsignal.protocol.PlaintextContent;
 import org.whispersystems.libsignal.protocol.SenderKeyDistributionMessage;
 import org.whispersystems.libsignal.state.PreKeyBundle;
 import org.whispersystems.libsignal.state.PreKeyRecord;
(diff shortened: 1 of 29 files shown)


jim-signal (approved):
LGTM

java/tests/src/test/java/org/whispersystems/libsignal/SessionCipherTest.java
```diff
@@ -95,7 +95,9 @@ public void testDecryptAfterReset() throws Exception {
     CiphertextMessage message2 = aliceCipher.encrypt(alicePlaintext);
 
     SessionRecord bobSession = bobStore.loadSession(aliceAddress);
+    bobSession.currentRatchetKeyMatches(Curve.generateKeyPair().getPublicKey()); // very unlikely but possible
```
alan-signal, line 98: You appear to be missing an `assertFalse` here. I would then put the comment as the failure message:

```java
assertFalse("very unlikely but possible, just rerun", bobSession.currentRatchetKeyMatches(Curve.generateKeyPair().getPublicKey()));
```

If it ever did happen, we would see message and rerun tests.

Unless this is astronomically unlikely, in which case, forget the comment altogether.

java/tests/src/test/java/org/signal/libsignal/metadata/SealedSessionCipherTest.java
```diff
@@ -224,6 +228,44 @@ public void testProtocolException() throws UntrustedIdentityException, InvalidKe
...
+    CiphertextMessage bobMessage = bobUnsealedCipher.encrypt("reply".getBytes());
+
+    DecryptionErrorMessage errorMessage = DecryptionErrorMessage.forOriginalMessage(bobMessage.serialize(), bobMessage.getType(), 408);
+    PlaintextContent errorMessageContent = new PlaintextContent(errorMessage);
+    UnidentifiedSenderMessageContent errorMessageUsmc = new UnidentifiedSenderMessageContent(errorMessageContent, senderCertificate, UnidentifiedSenderMessageContent.CONTENT_HINT_SUPPLEMENTARY, Optional.<byte[]>absent());
+    byte[] errorMessageCiphertext = aliceCipher.encrypt(bobAddress, errorMessageUsmc);
+
+    DecryptionResult result = bobCipher.decrypt(certificateValidator, errorMessageCiphertext, 31335);
+    PlaintextContent bobErrorContent = result.getPlaintextContent().get();
+    DecryptionErrorMessage bobErrorMessage = bobErrorContent.getDecryptionErrorMessage().get();
```
alan-signal, line 262: It seems that `PlaintextContent` may represent an error and it may not? But I cannot see a test for when it does not.

alan-signal (commented):
I have questions probably better answered 1:1

java/tests/src/test/java/org/whispersystems/libsignal/SessionCipherTest.java
jrose-signal: I had done this deliberately, but it really is astronomically unlikely. I'll put in the assert.

java/tests/src/test/java/org/signal/libsignal/metadata/SealedSessionCipherTest.java
jrose-signal: Right now the only thing it ever represents is the error, but we might add more plaintext messages in the future. This parallels the existing Content proto used by the apps.

https://github.com/signalapp/Signal-Android/blob/fb1b46b67e791cde9c49e30e8adced3e997bc63e/libsignal/service/src/main/proto/SignalService.proto#L35-L42

alan-signal (approved):
LGTM, whatever the client teams come back with following our discussion

jrose-signal:
Closing in favor of #316.

PR #316, opened 2021-05-26
Add DecryptionErrorMessage and PlaintextContent (alternate)

jrose-signal:
An alternative to #315 that avoids having a separate PlaintextContent *proto* while still having a PlaintextContent *type* (to prevent encoding messages that should be encrypted as plaintext). This makes the decryption side uniform with existing message types.

* * *

These will be used to help reset 1:1 and SenderKey sessions when messages fail to decrypt, rather than ending up in a state where message after message fails to decrypt. (1:1 sessions already have a form of automatic session reset in recent releases of Signal, but this provides a bit more feedback.)

### DIFF
--- a/java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
+++ b/java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
@@ -1,0 +1,58 @@
+/**
+ * Copyright (C) 2014-2016 Open Whisper Systems
+ *
+ * Licensed according to the LICENSE file in this repository.
+ */
+package org.whispersystems.libsignal.protocol;
+
+import org.signal.client.internal.Native;
+
+import org.whispersystems.libsignal.InvalidMessageException;
+import org.whispersystems.libsignal.ecc.ECPublicKey;
+import org.whispersystems.libsignal.util.guava.Optional;
+
+public final class DecryptionErrorMessage {
+
+  final long handle;
+
+  @Override
+  protected void finalize() {
+     Native.DecryptionErrorMessage_Destroy(this.handle);
+  }
+
+  DecryptionErrorMessage(long handle) {
+    this.handle = handle;
+  }
+
+  public DecryptionErrorMessage(byte[] serialized) throws InvalidMessageException {
+    handle = Native.DecryptionErrorMessage_Deserialize(serialized);
+  }
+
+  public static DecryptionErrorMessage forOriginalMessage(byte[] originalBytes, int messageType, long timestamp) {
+    return new DecryptionErrorMessage(
+      Native.DecryptionErrorMessage_ForOriginalMessage(originalBytes, messageType, timestamp));
+  }
+
+  public byte[] serialize() {
+    return Native.DecryptionErrorMessage_GetSerialized(this.handle);
+  }
+
+  public Optional<ECPublicKey> getRatchetKey() {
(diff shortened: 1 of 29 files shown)


jrose-signal:
[Diff between old and new versions](https://github.com/signalapp/libsignal-client/compare/jrose/DecryptionErrorMessage-and-PlaintextContent..jrose/DecryptionErrorMessage-and-PlaintextContent-2)

java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.signal.client.internal.Native;
+
+import org.whispersystems.libsignal.InvalidMessageException;
+import org.whispersystems.libsignal.ecc.ECPublicKey;
+import org.whispersystems.libsignal.util.guava.Optional;
+
+public final class DecryptionErrorMessage {
+
+  final long handle;
```
alan-signal, line 16: `private`? as you have `public long nativeHandle() {` if needed.

java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public long getTimestamp() {
+    return Native.DecryptionErrorMessage_GetTimestamp(this.handle);
+  }
+
+  /// For testing only
+  public static DecryptionErrorMessage extractFromSerializedContent(byte[] serializedContentBytes) throws InvalidMessageException {
+    return new DecryptionErrorMessage(
+      Native.DecryptionErrorMessage_ExtractFromSerializedContent(serializedContentBytes));
```
alan-signal, line 60: This does seems suspicious that there's a whole native call just for testing.

alan-signal (approved):

java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
jrose-signal: Cost of not having general support for the Content proto in libsignal-client, because the apps currently all have custom handling for what kind of message is inside. In the future I'd personally like to have more of that handled in libsignal-client but it'd make this PR snowball into even more than it already is.

java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
jrose-signal: Oh, I think I meant for this to go the other way around: removing the `nativeHandle()`. If it turns out I still need it across packages I'll definitely make this `private`.

java/java/src/main/java/org/whispersystems/libsignal/protocol/DecryptionErrorMessage.java
alan-signal: Is it possible to at least move this method to the test class?

So no danger of anyone calling from Java?

jrose-signal: I feel weird putting Native calls outside of the implementation classes. It really ought to be restricted to package visibility, but unfortunately we have multiple packages and our bridge logic doesn't currently support that.

jim-signal (approved):

PR #320, opened 2021-05-28
Fix intra doc link

pierwill:


jrose-signal (approved):
Thank you! I didn't know you could use rustdoc reference syntax in a full Markdown link.

PR #323, opened 2021-06-09
Add support for M1 and Catalyst architectures via cocoapods

Imperiopolis-Signal:


rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-11-09
+nightly
```
Imperiopolis-Signal, line 1: probably should lock to a specific nightly. not sure which to target, but it needs to be newer than what we were at to get these targets

swift/build_ffi.sh
```diff
@@ -85,7 +90,12 @@ if [[ -n "${DEVELOPER_SDK_DIR:-}" ]]; then
   export LIBRARY_PATH="${DEVELOPER_SDK_DIR}/MacOSX.sdk/usr/lib:${LIBRARY_PATH:-}"
 fi
 
-echo_then_run cargo build -p libsignal-ffi ${RELEASE_BUILD:+--release} ${VERBOSE:+--verbose}
+BUILD_CMD=cargo
+if [[ -n "${USE_XARGO:-}" ]]; then
+	BUILD_CMD=xargo
```
Imperiopolis-Signal, line 95: I originally tried to use the `-Zbuild-std` approach, but it resulted in a lot of problems with some of our dependencies. Using `xargo` works around the issues.

Cargo.toml
```diff
@@ -21,3 +21,6 @@ neon = { git = 'https://github.com/signalapp/neon', tag = 'signal-2021-06-02' }
 
 [profile.dev.package.num-bigint-dig]
 opt-level = 2 # too slow otherwise!
+
+[profile.release]
+panic = "abort"
```
jrose-signal, line 26: We're deliberately recovering from panics today; since most of the types here don't expose mutating operations, it's not hard to treat a panic like any other error (albeit one that should never happen). Are the new architectures missing that part of the Rust stdlib infrastructure?

rust/crypto/src/polyval/polyval_pmul.rs
```diff
@@ -39,7 +39,7 @@ impl PolyvalPmul {
 
     #[target_feature(enable = "crypto")]
     unsafe fn process_block(&mut self, chunk: *const u8) {
-        let poly: poly64_t = cast(0xc200000000000000u64);
+        let poly: u64 = cast(0xc200000000000000u64);
```
jrose-signal, line 42: Note to self to double-check that this is correct (and remove the `cast` if so).

swift/build_ffi.sh
```diff
@@ -15,7 +15,7 @@ export CARGO_PROFILE_RELEASE_DEBUG=1 # enable line tables
 
 usage() {
   cat >&2 <<END
-Usage: $(basename "$0") [-d|-r] [-v] [--generate-ffi|--verify-ffi]
+Usage: $(basename "$0") [-d|-r] [-v] [--generate-ffi|--verify-ffi|--use-xargo]
```
jrose-signal, line 18: Is this mutually exclusive with the others? …I guess it probably is, since we're not set up to generate headers with xargo.

rust-toolchain
jrose-signal: I'll check if there's any particular nightly we should be pinning to; otherwise I'll just pick "last week" or something.

jrose-signal (commented):
Thank you for working on this, and nice work!

jrose-signal:
It may be worth pulling the toolchain update out into its own PR or at least its own commit, since it has a bunch of small knock-on effects that are just the natural result of new Clippy warnings and unstable changes.

Cargo.toml
Imperiopolis-Signal: They fail to compile without this change, yeah. I don't fully understand the ramifications

rust-toolchain
Imperiopolis-Signal: I pinned it to today's arbitrarily, but can tweak if there's a specific one

SignalClient.podspec
```diff
@@ -34,12 +34,18 @@ Pod::Spec.new do |s|
...
       'OTHER_LDFLAGS' => '$(LIBSIGNAL_FFI_LIB_IF_NEEDED)',
 
-      # Some day this will have to be updated for arm64 Macs (and the corresponding arm64 iOS simulator)
-      'CARGO_BUILD_TARGET[sdk=iphonesimulator*]' => 'x86_64-apple-ios',
+      'CARGO_BUILD_TARGET[sdk=iphonesimulator*][arch=arm64]' => 'aarch64-apple-ios-sim',
+      'CARGO_BUILD_TARGET[sdk=iphonesimulator*][arch=*]' => 'x86_64-apple-ios',
       'CARGO_BUILD_TARGET[sdk=iphoneos*]' => 'aarch64-apple-ios',
-      'CARGO_BUILD_TARGET[sdk=macosx*]' => 'x86_64-apple-darwin',
-      'ARCHS[sdk=iphonesimulator*]' => 'x86_64',
-      'ARCHS[sdk=iphoneos*]' => 'arm64',
```
jrose-signal, line 42: Looks like these were here to keep `pod lib lint` from trying to build for 32-bit platforms. CocoaPods/CocoaPods#5854 suggests we could try

```
'ARCHS[sdk=iphonesimulator*]' => '$(ARCHS_STANDARD_64_BIT)'
```

though that's from a few years ago, so it might need tweaking for Xcode 12.x.

jrose-signal:
Worth updating swift/README.md with the new requirements?

Cargo.toml
Imperiopolis-Signal: I figured this out (we have to build panic_unwind from source with xargo too)

swift/build_ffi.sh
Imperiopolis-Signal: Yeah, cbindgen doesn't work correctly with the targets that require being built from source (using xargo)

Imperiopolis-Signal:
> It may be worth pulling the toolchain update out into its own PR or at least its own commit, since it has a bunch of small knock-on effects that are just the natural result of new Clippy warnings and unstable changes.

I pulled it out into its own commit. I also tried to address all the clippy warnings in that first commit, though I'm not 100% confident in some of the changes so please take a look :)

.gitignore
```diff
@@ -23,3 +23,5 @@ java/gradle.properties
 java/local.properties
 java/android/src/main/jniLibs
 java/java/src/main/resources
+
+.DS_Store
```
jrose-signal, line 27: Heh, I just assume every macOS developer has this in their global gitignore by now. But sure.

swift/README.md
```diff
@@ -54,3 +54,12 @@ When exposing new APIs to Swift, you will need to add the `--generate-ffi` flag
...
 ...is not supported. In theory we could make this work through the use of a custom pkg-config file and requiring clients to set `PKG_CONFIG_PATH` (or install the Rust build products), but since Signal itself does not use this configuration it's considered extra maintenance burden. Development as a package is supported as a lightweight convenience (as well as a cross-platform one), but the CocoaPods build is considered the canonical one.
+
+# M1 Simulator and Catalyst
+
+Rust targets for both the M1 Simulator and Catalyst are still in tier 3 support, so we use `xargo` to build the standard library. 
+
+In order to compile for these platforms you will need to:
+* Install Xargo with `cargo install xargo`
+* Install the standard library component with `rustup component add rust-src`
+* Add the `--use-xargo` flag to your `build_ffi.sh` invocation
```
jrose-signal, line 65: This is good. Maybe a note on the last line that that's only necessary when building yourself, and not when using as a CocoaPod?

(Do we care about having the CocoaPod recover nicely if you don't have xargo installed?)

swift/build_ffi.sh
```diff
@@ -85,7 +90,18 @@ if [[ -n "${DEVELOPER_SDK_DIR:-}" ]]; then
...
-echo_then_run cargo build -p libsignal-ffi ${RELEASE_BUILD:+--release} ${VERBOSE:+--verbose}
+BUILD_CMD=cargo
+if [[ -n "${USE_XARGO:-}" ]]; then
+	if ! command -v xargo &> /dev/null; then
+    echo "error: xargo not installed" >&2
+    echo 'note: get it by running' >&2
+    printf "\n\t%s\n\n" "cargo install xargo" >&2
+    exit 1
+	fi
+	BUILD_CMD=xargo
```
jrose-signal, line 101: Looks like some tabs snuck into a spaces-indentation file.

swift/README.md
Imperiopolis-Signal: > Maybe a note on the last line that that's only necessary when building yourself, and not when using as a CocoaPod?

That's not true is it? Like if you're installing with cocoapods you still need to:
* Have rust installed
* Have the appropriate targets installed

and now..
* Have xargo installed
* Have the rust-src component added

jrose-signal: I meant just the very last bullet, about "add `--use-xargo`". The podspec already includes that.

jrose-signal (approved):

PR #324, opened 2021-06-09
node: Build node modules with python3

cryptomilk:
This fixes the build on openSUSE Tumbleweed.


jrose-signal:
We probably should have been doing that all along, thank you. (I hope there's not a problem on Windows.)

PR #326, opened 2021-06-10
Add Swift wrappers for AES operations

Imperiopolis-Signal:


swift/Sources/SignalClient/Aes256GcmEncryption.swift
```diff
@@ -0,0 +1,71 @@
...
+            try checkError(error)
+        }
+    }
+
+    public func encrypt<MutableBytes: ContiguousMutableBytes>(
+        _ bytes: inout MutableBytes,
+        _ offset: UInt32,
+        _ length: UInt32) throws {
+        try bytes.withUnsafeMutableBytes {
+            let error = signal_aes256_gcm_encryption_update(nativeHandle,
```
Imperiopolis-Signal, line 57: FWIW I don't love this API. It's weird to me that the underlying rust takes in a mutable copy of the plaintext data and then mutates it into the ciphertext

jrose-signal: The intent is to avoid copies, since you're going to throw away the plaintext after you're done. I think we can do a little better for Swift, though, thanks to pervasive copy-on-write: you could have an API that passes an argument in by value, moves it to a local var, mutates it, and returns it, and that shouldn't do any copies if it's the only reference to the plaintext. (And if it wasn't, you'd have to do a copy anyway for the `inout` version.)

The other thing Swift has is cheap slices, so I don't think we need the offset and length parameters in the public API as long as ArraySlice also conforms to the new protocol.

swift/Sources/SignalClient/ContiguousMutableBytes.swift
```diff
@@ -0,0 +1,13 @@
...
+//
+
+import Foundation
+
+public protocol ContiguousMutableBytes {
+    @inlinable mutating func withUnsafeMutableBytes<R>(_ body: (UnsafeMutableRawBufferPointer) throws -> R) rethrows -> R
+}
+
+extension Data: ContiguousMutableBytes {}
+extension Array: ContiguousMutableBytes {}
```
jrose-signal, line 13: Honestly this isn't really what we want; like ContiguousBytes we should only conform when the element is UInt8. (Unfortunately Data only has `withUnsafeMutableBytes` and `withContiguousMutableStorageIfAvailable` but not `withContiguousMutableStorageEvenIfNotCurrentlyContiguous`.)

Foundation's MutableDataProtocol is also very close to what we want, but while it gives us a mutable collection and contiguous regions we could feed piecemeal into AES-GCM, it doesn't give us both at once (mutable contiguous regions). I suspect this is an oversight that now can't be corrected for compatibility reasons. :-(

swift/Sources/SignalClient/ContiguousMutableBytes.swift
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright 2020 Signal Messenger, LLC
```
jrose-signal, line 2: 2021 (yes, we need some automated checking like Signal-iOS has)

swift/Sources/SignalClient/Aes256Ctr32.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+
+import SignalFfi
+import Foundation
+
+public class Aes256Ctr32: ClonableHandleOwner {
+    public init<KeyBytes, NonceBytes>(
+        _ key: KeyBytes,
+        _ nonce: NonceBytes,
+        _ initial_ctr: UInt32) throws
```
jrose-signal, line 13: Nitpick: even though there are Rust naming conventions throughout much of the code (especially the tests), we've tried to keep the public interface to Swift naming conventions (camelCase).

Imperiopolis-Signal: I copied the naming pattern from `Aes256GcmSiv` but happy to update :)

swift/Sources/SignalClient/ContiguousMutableBytes.swift
Imperiopolis-Signal: Given your API change suggestions, I think we can perhaps do away with this anyhow

swift/Sources/SignalClient/Aes256Ctr32.swift
jrose-signal: \me shakes fist at Jack, then at past self

jrose-signal:
Closing; superseded by a similar PR on our private repo.

PR #328, opened 2021-06-22
Use more RustCrypto

jrose-signal:
Now that RustCrypto crates support runtime-detected ARMv8 and x86_64 crypto intrinsics, we don't need our own implementations of AES to get good performance.* We still have our own wrapper for AES-256-CTR, and we have a streaming version of AES-256-GCM, but overall this makes the signal-crypto crate a lot smaller.

This is a lot of changes for one PR, but that's because our PR tests ensure that Cargo doesn't include two incompatible versions of a dependency in the output binary. This means that in order to pick up the new version of the `aead` crate, I had to remove all uses of the old version.

\* ~~blocked on RustCrypto/utils#435 getting into a release~~ done!


jrose-signal:
Nice side benefit:
```
current build is -7% larger than v0.8.1 (current: 1870912 bytes, v0.8.1: 2030936 bytes)
       0.2.2: ************** (1382592 bytes)
       0.2.3: *************** (1489088 bytes)
       0.3.4: ******************* (1949016 bytes)
       0.4.0: ******************* (1973592 bytes)
       0.5.0: ******************* (1973592 bytes)
       0.5.1: ******************* (1969496 bytes)
       0.6.0: ******************** (1998168 bytes)
       0.7.0: ******************** (2030936 bytes)
       0.8.0: ******************** (2030936 bytes)
       0.8.1: ******************** (2030936 bytes)
     current: ****************** (1870912 bytes)
```

jrose-signal:
Oh, looks like the benefit all came from [bumping the toolchain version](https://github.com/signalapp/libsignal-client/runs/2798713401), and this actually increases the size a little (1,858,608 to 1,870,912). Oh well.

java/java/src/main/java/org/signal/internal/Native.java
```diff
@@ -86,7 +86,7 @@ private Native() {}
 
   public static native byte[] Aes256GcmSiv_Decrypt(long aesGcmSiv, byte[] ctext, byte[] nonce, byte[] associatedData);
   public static native void Aes256GcmSiv_Destroy(long handle);
-  public static native byte[] Aes256GcmSiv_Encrypt(long aesGcmSiv, byte[] ptext, byte[] nonce, byte[] associatedData);
+  public static native byte[] Aes256GcmSiv_Encrypt(long aesGcmSivObj, byte[] ptext, byte[] nonce, byte[] associatedData);
```
jim-signal, line 89: Why the Obj naming here?

jim-signal (approved):

java/java/src/main/java/org/signal/internal/Native.java
jrose-signal: It comes from the Rust source, which renamed a parameter named `aes_gcm_siv` so that it wouldn't shadow the package name.

PR #330, opened 2021-07-02
GitHub: Run Rust tests for 32-bit platforms as well

jrose-signal:


jrose-signal:
Good, this correctly catches a panic I'm fixing!

jrose-signal:
Pulled into #331.

PR #331, opened 2021-07-02
Fix AES-256-GCM on 32-bit platforms

jrose-signal:
Previously our streaming AES-256-GCM implementation would panic on 32-bit platforms at the end of decryption because we copied `usize` values into a `u64`-sized slot.

To keep this from happening again, run the Rust PR tests on 32-bit Linux…except the device transfer ones, which depend on the openssl crate, which doesn't support Cargo's `--target`.

While here, improve the Java side a bit by (a) defining a constant for the AES-256-GCM tag size, and (b) removing an API that we never marked `public` but didn't pan out anyway.


.github/workflows/build_and_test.yml
```diff
@@ -89,11 +89,14 @@ jobs:
...
     - uses: actions/checkout@v2
 
+    - run: sudo apt install gcc-multilib
+
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
         components: rustfmt, clippy
+        target: i686-unknown-linux-gnu
```
jim-signal, line 99: Just checking, but this is an additional target, the default is still x86_64?

jim-signal (approved):

.github/workflows/build_and_test.yml
jrose-signal: Correct (and verified in the build log by seeing the 32-bit tests fail, oops).

PR #335, opened 2021-07-09
Upgrade prost to 0.8

rubdos:
The dreaded https://github.com/tokio-rs/prost/issues/154 (Prost Message `vec![]` + `encode` + `expect`) can now be get rid of in most places, thanks to https://github.com/tokio-rs/prost/pull/378. In some places, I've left the `.encode` call because it's doing some prepending of magic bytes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1306,9 +1306,9 @@ dependencies = [
 
 [[package]]
 name = "prost"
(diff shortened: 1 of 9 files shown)


jrose-signal (approved):
Thanks for getting that into prost, nice cleanup!

PR #337, opened 2021-07-23
Put aes' armv8 feature behind nightly feature

rubdos:
Follow-up from #284, as promised. As you said, it became a lot easier :-)

Would you like me to make the armv8 feature default in the jni/node/native ffi crates? Maybe it makes sense there. Otherwise, you'll have to supply the armv8 feature flag from the build system in Signal-Android and Signal-iOS

(off topic: re NEON and Dalek, we have something upcoming!)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -77,8 +77,40 @@ jobs:
       run: echo "::error file=.github/workflows/build_and_test.yml::File not included in any filter" && false
       if: ${{ !contains(steps.filter.outputs.*, 'true') }}
 
-  rust:
-    name: Rust
+  rust-stable:
+    name: Rust stable
+
+    runs-on: ubuntu-latest
+
+    needs: changes
+
+    if: ${{ needs.changes.outputs.rust == 'true' }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - run: sudo apt-get install gcc-multilib
+
+    - name: Install Rust
+      run: rustup +stable target add i686-unknown-linux-gnu
+
+    - name: Build
+      run: cargo +stable build --verbose
+
+    - name: Run tests
+      run: cargo +stable test --workspace --verbose -- --include-ignored
+
+    - name: Run tests (32-bit)
+      # Exclude device-transfer because OpenSSL, used for reference results,
+      # doesn't support the --target option.
+      # Exclude signal-neon-futures because those tests run Node
+      run: cargo +stable test --workspace --verbose --target i686-unknown-linux-gnu --exclude device-transfer --exclude signal-neon-futures -- --include-ignored
+
+    - name: Build benches
+      run: cargo +stable build --workspace --benches --verbose
+
+  rust-nightly:
+    name: Rust nightly
 
     runs-on: ubuntu-latest
 
--- a/rust/bridge/shared/Cargo.toml
+++ b/rust/bridge/shared/Cargo.toml
@@ -15,7 +15,7 @@ libsignal-protocol = { path = "../../protocol" }
 signal-crypto = { path = "../../crypto" }
 device-transfer = { path = "../../device-transfer" }
 libsignal-bridge-macros = { path = "macros" }
-aes-gcm-siv = { version = "0.10.1", features = ["armv8"] }
+aes-gcm-siv = "0.10.1"
 futures-util = "0.3.7"
 log = "0.4"
 paste = "1.0"
@@ -35,3 +35,4 @@ signal-neon-futures = { path = "../node/futures", optional = true }
 ffi = ["libc", "libsignal-bridge-macros/ffi"]
 jni = ["jni_crate", "libsignal-bridge-macros/jni"]
 node = ["neon", "linkme", "signal-neon-futures", "libsignal-bridge-macros/node"]
+armv8 = ["aes-gcm-siv/armv8", "libsignal-protocol/armv8", "signal-crypto/armv8"]
--- a/rust/bridge/shared/Cargo.toml
+++ b/rust/bridge/shared/Cargo.toml
@@ -32,6 +32,7 @@ linkme = { version = "0.2.4", optional = true }
 signal-neon-futures = { path = "../node/futures", optional = true }
 
 [features]
+default = ["armv8"]
 ffi = ["libc", "libsignal-bridge-macros/ffi"]
 jni = ["jni_crate", "libsignal-bridge-macros/jni"]
 node = ["neon", "linkme", "signal-neon-futures", "libsignal-bridge-macros/node"]
--- a/rust/protocol/Cargo.toml
+++ b/rust/protocol/Cargo.toml
@@ -12,8 +12,8 @@ license = "AGPL-3.0-only"
 repository = "https://github.com/signalapp/libsignal-client"
 
 [dependencies]
-aes = { version = "0.7.4", features = ["armv8", "ctr"] }
-aes-gcm-siv = { version = "0.10.1", features = ["armv8"] }
+aes = { version = "0.7.4", features = ["ctr"] }
+aes-gcm-siv = "0.10.1"
 arrayref = "0.3.6"
 async-trait = "0.1.41"
 block-modes = "0.8"
@@ -40,6 +40,7 @@ u32_backend = ["curve25519-dalek/u32_backend"]
 u64_backend = ["curve25519-dalek/u64_backend"]
 simd_backend = ["curve25519-dalek/simd_backend"]
 nightly = ["curve25519-dalek/nightly"]
+armv8 = ["aes/armv8", "aes-gcm-siv/armv8"]
 
 [dev-dependencies]
 criterion = "0.3"
--- a/rust/bridge/ffi/Cargo.toml
+++ b/rust/bridge/ffi/Cargo.toml
@@ -14,6 +14,9 @@ license = "AGPL-3.0-only"
 name = "signal_ffi"
 crate-type = ["staticlib"]
 
+[features]
+armv8 = ["libsignal-protocol/armv8", "signal-crypto/armv8"]
+
 [dependencies]
 libsignal-protocol = { path = "../../protocol" }
 device-transfer = { path = "../../device-transfer" }
--- a/rust/bridge/ffi/Cargo.toml
+++ b/rust/bridge/ffi/Cargo.toml
@@ -15,6 +15,7 @@ name = "signal_ffi"
 crate-type = ["staticlib"]
 
 [features]
+default = ["armv8"]
 armv8 = ["libsignal-protocol/armv8", "signal-crypto/armv8"]
 
 [dependencies]
--- a/rust/bridge/jni/Cargo.toml
+++ b/rust/bridge/jni/Cargo.toml
@@ -15,6 +15,7 @@ name = "signal_jni"
 crate-type = ["cdylib"]
 
 [features]
+default = ["armv8"]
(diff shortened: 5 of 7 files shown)


jrose-signal:
And with this, `cargo +stable test` passes!

Yes, please do make armv8 a default feature of the three bridge crates. That'll help IDEs running the equivalent of `cargo check` to share build products with any actual builds.

rubdos:
> Yes, please do make armv8 a default feature of the three bridge crates. That'll help IDEs running the equivalent of `cargo check` to share build products with any actual builds.

And with that, `cargo +stable test` stopped passing! Just kidding; I added it as default on `rust/bridge/*/Cargo.toml` ;-)

Maybe it's even worth it now to add `cargo +stable test --no-default-features` to CI?

jrose-signal:
> Maybe it's even worth it now to add `cargo +stable test --no-default-features` to CI?

Hm, I don't love extending test times even more here, but maybe we can gain some of it back by reordering the steps in the Rust tests so that less gets rebuilt between config changes. (I think that's mostly just putting Clippy immediately after "build".) Well, if you add it to this PR, we can see how the time changes.

Also, I suggest dropping `--no-default-features`; the way we've set up the workspace has the component crates as the default members, excluding the bridge libraries. So we can still add features that should be supported on stable, they just won't be default.

rubdos:
> Hm, I don't love extending test times even more here, but maybe we can gain some of it back by reordering the steps in the Rust tests so that less gets rebuilt between config changes. (I think that's mostly just putting Clippy immediately after "build".) Well, if you add it to this PR, we can see how the time changes.

I've added rust-stable and rust-nightly as separate jobs. Stable doesn't run clippy or `./bin/verify_crate_versions.py`, and doesn't use `-Z unstable-options`. This way, the tests should run in parallel, not sure whether that's what you would have done.

> Also, I suggest dropping `--no-default-features`; the way we've set up the workspace has the component crates as the default members, excluding the bridge libraries. So we can still add features that should be supported on stable, they just won't be default.

Ack!

rubdos:
... and with those two fix-ups, it even works. Let me know if this looks good to you, and I'll squash them or push more fix-ups.

.github/workflows/build_and_test.yml
```diff
@@ -77,8 +77,51 @@ jobs:
...
+        profile: minimal
+        toolchain: stable
+        components: rustfmt, clippy
+        target: i686-unknown-linux-gnu
+
+    - name: Check for duplicate dependencies
+      run: |
+        DUPS="$(cargo tree -d -e normal --workspace)"
+        echo "$DUPS"
+        test -z "$DUPS"
```
jrose-signal, line 106: I don't think we need to do this again; anyone who's using this as a Rust library will either want to redo the check anyway or won't care.

.github/workflows/build_and_test.yml
```diff
@@ -77,8 +77,51 @@ jobs:
...
+    - uses: actions/checkout@v2
+
+    - run: sudo apt-get install gcc-multilib
+
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        components: rustfmt, clippy
```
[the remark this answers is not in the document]
jrose-signal: Actually, do we need the install at all? The [default GitHub Ubuntu image has a recent stable](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md). `rustup target add i686-unknown-linux-gnu` might be enough. (I went back and forth on whether the 32-bit stable testing was really necessary but why not.)

.github/workflows/build_and_test.yml
```diff
@@ -77,8 +77,51 @@ jobs:
...
+      run: cargo +stable build --verbose
+
+    - name: Run tests
+      run: cargo +stable test --workspace --verbose -- --include-ignored
+
+    - name: Run tests (32-bit)
+      # Exclude device-transfer because OpenSSL, used for reference results,
+      # doesn't support the --target option.
+      # Exclude signal-neon-futures because those tests run Node
+      run: cargo +stable test --workspace --verbose --target i686-unknown-linux-gnu --exclude device-transfer --exclude signal-neon-futures -- --include-ignored
```
jrose-signal, line 118: I believe signal-neon-futures isn't in the default set anyway.

rust/bridge/shared/Cargo.toml
```diff
@@ -15,7 +15,7 @@ libsignal-protocol = { path = "../../protocol" }
 signal-crypto = { path = "../../crypto" }
 device-transfer = { path = "../../device-transfer" }
 libsignal-bridge-macros = { path = "macros" }
-aes-gcm-siv = { version = "0.10.1", features = ["armv8"] }
+aes-gcm-siv = { version = "0.10.1" }
```
jrose-signal, line 18: This can go back to the short form.

rust/protocol/Cargo.toml
```diff
@@ -12,8 +12,8 @@ license = "AGPL-3.0-only"
 repository = "https://github.com/signalapp/libsignal-client"
 
 [dependencies]
-aes = { version = "0.7.4", features = ["armv8", "ctr"] }
-aes-gcm-siv = { version = "0.10.1", features = ["armv8"] }
+aes = { version = "0.7.4", features = ["ctr"] }
+aes-gcm-siv = { version = "0.10.1" }
```
jrose-signal, line 16: Here too.

jrose-signal (commented):
> I've added rust-stable and rust-nightly as separate jobs. Stable doesn't run clippy or `./bin/verify_crate_versions.py`, and doesn't use `-Z unstable-options`. This way, the tests should run in parallel, not sure whether that's what you would have done.

It's not what I was thinking, since there are organization-level limits on the number of jobs you can run at once, but after checking internally I think it's reasonable. Thanks!

> Is it also an idea to drop the `rust-toolchain` file now? Then I probably have to swap `+stable` for `+nightly` somehow.

The `rust-toolchain` file is effectively a lockfile for us just as much as Cargo.lock: it makes the guarantee that *this* commit will build with *this* version of nightly. That's important for historical tags and especially for the Dockerized Android build, which is intended to be reproducible by anyone who wants to verify that what's in the Android app is what we said it is. And it won't get in the way of anyone using libsignal-client crates as dependencies.

.github/workflows/build_and_test.yml
rubdos: I copied that from the nightly script. Still want me to drop it?

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
rubdos: At first try, it seemed to install `nightly` because of the `rust-toolchain` file (which is why I suggested to drop it in the first place). 

I guess at least `rustfmt` and `clippy` can be dropped though.

rust/bridge/shared/Cargo.toml
rubdos: > > I've added rust-stable and rust-nightly as separate jobs. Stable doesn't run clippy or `./bin/verify_crate_versions.py`, and doesn't use `-Z unstable-options`. This way, the tests should run in parallel, not sure whether that's what you would have done.
> 
> It's not what I was thinking, since there are organization-level limits on the number of jobs you can run at once, but after checking internally I think it's reasonable. Thanks!

If you would rather have me add an additional install block and run it in the same job, I'll gladly change it!

> > Is it also an idea to drop the `rust-toolchain` file now? Then I probably have to swap `+stable` for `+nightly` somehow.
> 
> The `rust-toolchain` file is effectively a lockfile for us just as much as Cargo.lock: it makes the guarantee that _this_ commit will build with _this_ version of nightly. That's important for historical tags and especially for the Dockerized Android build, which is intended to be reproducible by anyone who wants to verify that what's in the Android app is what we said it is. And it won't get in the way of anyone using libsignal-client crates as dependencies.

Ack, sounds fair. Only "problem" with it in my eyes, is that this triggers a download of an (often out-of-date) nightly chain for developers that are external to Signal (like me), who often just want to use stable.

rubdos:
> > I've added rust-stable and rust-nightly as separate jobs. Stable doesn't run clippy or `./bin/verify_crate_versions.py`, and doesn't use `-Z unstable-options`. This way, the tests should run in parallel, not sure whether that's what you would have done.
> 
> It's not what I was thinking, since there are organization-level limits on the number of jobs you can run at once, but after checking internally I think it's reasonable. Thanks!

If you would rather have me add an additional install block and run it in the same job, I'll gladly change it!

> > Is it also an idea to drop the `rust-toolchain` file now? Then I probably have to swap `+stable` for `+nightly` somehow.
> 
> The `rust-toolchain` file is effectively a lockfile for us just as much as Cargo.lock: it makes the guarantee that _this_ commit will build with _this_ version of nightly. That's important for historical tags and especially for the Dockerized Android build, which is intended to be reproducible by anyone who wants to verify that what's in the Android app is what we said it is. And it won't get in the way of anyone using libsignal-client crates as dependencies.

Ack, sounds fair. Only "problem" with it in my eyes, is that this triggers a download of an (often out-of-date) nightly chain for developers that are external to Signal (like me), who often just want to use stable.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
jrose-signal: `rustup +stable target add i686-unknown-linux-gnu`, then.

.github/workflows/build_and_test.yml
jrose-signal: Yes please, it falls out of `--all` vs. `--workspace`.

jrose-signal:
> Ack, sounds fair. Only "problem" with it in my eyes, is that this triggers a download of an (often out-of-date) nightly chain for developers that are external to Signal (like me), who often just want to use stable.

It'll only do that when you are developing libsignal-client itself, which is correct: any changes you want to make need to work on our pinned nightly until we decide to roll forward to a new compiler.

rubdos:
> It'll only do that when you are developing libsignal-client itself, which is correct: any changes you want to make need to work on our pinned nightly until we decide to roll forward to a new compiler.

Sounds like a task for the CI in my ears, but it's only an extremely minor inconvenience, so don't bother!

jrose-signal:
! I guess I was wrong about signal-neon-futures. All right, we don't need to untangle that at the moment; going back to the way it was is fine. Sorry for the extra back-and-forth.

rubdos:
Don't worry.

Looks like `--workspace` is the new version of `--all`, and includes literally all members (instead of the default ones, as given in the `workspace.default-members` array).

.github/workflows/build_and_test.yml
```diff
@@ -77,8 +77,40 @@ jobs:
...
+      run: cargo +stable build --verbose
+
+    - name: Run tests
+      run: cargo +stable test --workspace --verbose -- --include-ignored
+
+    - name: Run tests (32-bit)
+      # Exclude device-transfer because OpenSSL, used for reference results,
+      # doesn't support the --target option.
+      # Exclude signal-neon-futures because those tests run Node
+      run: cargo +stable test --workspace --verbose --target i686-unknown-linux-gnu --exclude device-transfer -- --exclude signal-neon-features --include-ignored
```
jrose-signal, line 107: Ah, the `--` has to come after the `--exclude blah`.

rubdos: Damnit, bad copy-and-paste. Sorry for the noise, will keep an eye on CI myself now.

jrose-signal (approved):

PR #342, opened 2021-08-17
Sealed Sender v2: reuse work when sending to the same identity key

jrose-signal:
Identity keys are tied to an account, not a device, so when a multi-recipient sealed sender message is sent to every device for a particular account, a lot of work is wasted re-encoding the ephemeral message key and authentication tag for each device. This commit checks to see if the current identity key is the same as the previous one and reuses the previous encrypted message key and authentication tag if so, drastically reducing the marginal cost of another linked device (from ~200ms to ~2ms on my machine).

### DIFF
--- a/rust/protocol/src/sealed_sender.rs
+++ b/rust/protocol/src/sealed_sender.rs
@@ -10,6 +10,7 @@ use crate::{
 };
 
 use crate::crypto;
+use crate::curve;
 use crate::proto;
 use crate::session_cipher;
 use aes_gcm_siv::aead::{AeadInPlace, NewAead};
@@ -565,12 +566,19 @@ impl UnidentifiedSenderMessage {
             SEALED_SENDER_V2_VERSION => {
                 // Uses a flat representation: C || AT || E.pub || ciphertext
                 let remaining = &data[1..];
-                if remaining.len() < 32 + 16 + 32 {
+                if remaining.len()
+                    < sealed_sender_v2::MESSAGE_KEY_LEN
+                        + sealed_sender_v2::AUTH_TAG_LEN
+                        + curve::curve25519::PUBLIC_KEY_LENGTH
+                {
                     return Err(SignalProtocolError::InvalidProtobufEncoding);
                 }
-                let (encrypted_message_key, remaining) = remaining.split_at(32);
-                let (encrypted_authentication_tag, remaining) = remaining.split_at(16);
-                let (ephemeral_public, encrypted_message) = remaining.split_at(32);
+                let (encrypted_message_key, remaining) =
+                    remaining.split_at(sealed_sender_v2::MESSAGE_KEY_LEN);
+                let (encrypted_authentication_tag, remaining) =
+                    remaining.split_at(sealed_sender_v2::AUTH_TAG_LEN);
+                let (ephemeral_public, encrypted_message) =
+                    remaining.split_at(curve::curve25519::PUBLIC_KEY_LENGTH);
 
                 Ok(Self::V2 {
                     ephemeral_public: PublicKey::from_djb_public_key_bytes(ephemeral_public)?,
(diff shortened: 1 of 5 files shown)


rust/protocol/src/sealed_sender.rs
```diff
@@ -565,11 +565,13 @@ impl UnidentifiedSenderMessage {
             SEALED_SENDER_V2_VERSION => {
                 // Uses a flat representation: C || AT || E.pub || ciphertext
                 let remaining = &data[1..];
-                if remaining.len() < 32 + 16 + 32 {
+                if remaining.len() < sealed_sender_v2::M_LEN + sealed_sender_v2::AT_LEN + 32 {
```
jim-signal, line 568: nit: Should the 32 also be a const, or is it well-known? It stands out in the diff.

rust/protocol/src/sealed_sender.rs
```diff
@@ -930,27 +936,40 @@ pub async fn sealed_sender_multi_recipient_encrypt<R: Rng + CryptoRng>(
...
-        )?;
-
-        let at_i = sealed_sender_v2::compute_authentication_tag(
-            our_identity.private_key(),
-            their_identity.public_key(),
-            Direction::Sending,
-            &e_pub,
-            &c_i,
-        )?;
+        let previous_endpoint = serialized.len();
```
jim-signal, line 939: nit: Seems like previous_endpoint_len would be clearer?

rust/protocol/src/sealed_sender.rs
```diff
@@ -565,11 +565,13 @@ impl UnidentifiedSenderMessage {
...
                 // Uses a flat representation: C || AT || E.pub || ciphertext
                 let remaining = &data[1..];
-                if remaining.len() < 32 + 16 + 32 {
+                if remaining.len() < sealed_sender_v2::M_LEN + sealed_sender_v2::AT_LEN + 32 {
                     return Err(SignalProtocolError::InvalidProtobufEncoding);
                 }
-                let (encrypted_message_key, remaining) = remaining.split_at(32);
-                let (encrypted_authentication_tag, remaining) = remaining.split_at(16);
+                let (encrypted_message_key, remaining) =
+                    remaining.split_at(sealed_sender_v2::M_LEN);
```
peter-signal, line 572: Nit: MESSAGE_LEN would be a bit more readable

rust/protocol/src/sealed_sender.rs
```diff
@@ -565,11 +565,13 @@ impl UnidentifiedSenderMessage {
...
+                let (encrypted_authentication_tag, remaining) =
+                    remaining.split_at(sealed_sender_v2::AT_LEN);
```
peter-signal, line 574: Nit: AUTH_TAG_LEN would be a bit more readable

jim-signal (approved):
Looks good, just some nits.

rust/protocol/src/sealed_sender.rs
```diff
@@ -906,6 +911,7 @@ pub async fn sealed_sender_multi_recipient_encrypt<R: Rng + CryptoRng>(
         .expect("cannot fail encoding to Vec");
 
     let our_identity = identity_store.get_identity_key_pair(ctx).await?;
+    let mut previous_identity = None;
```
peter-signal, line 914: Nit: should this be previous_their_idenity?

peter-signal (approved):

rust/protocol/src/sealed_sender.rs
jrose-signal: It's not the message length, it's the seed for the keys used to encrypt the message. (I asked Trevor for more verbose names and he replied that there wasn't a standard term, and that cryptographers tend to use initials anyway.) Spelling out long constant names also makes them stand out a *lot* in this code (AUTHENTICATION_TAG_LEN).

rust/protocol/src/sealed_sender.rs
jrose-signal: :-/ The constant for it is in another module and isn't currently public, but you're right, it should be.

PR #343, opened 2021-08-18
Fix armv8 feature flag for signal-crypto

jrose-signal:
This was a rebase failure where the armv8 support in signal-crypto dependencies was not actually put behind the armv8 flag. Our CI didn't catch it because we don't do a build that's both stable *and* targeting aarch64, but for completeness we should get it right.


jrose-signal:
cc @rubdos

rubdos:
Oops, was that my rebase failure?

Patch LGTM, but haven't checked any other `Cargo.toml`s.

PR #346, opened 2021-08-20
Bulk fetch identity keys for Sealed Sender v2

indutny-signal:
Instead of fetching identity keys for each recipient one-by-one with
this change we will fetch them in bulk before calling low-level
`sealed_sender_multi_recipient_encrypt`.

This is most beneficial for the Node.js library since every asynchronous
fetch there would interrupt event-loop and make encryption slower.

cc @jrose-signal


indutny-signal:
Tons of API churn and no performance benefit. My bad!

jrose-signal:
Closing due to age; we may revisit in the future.

PR #348, opened 2021-08-30
Java: don't eagerly strip libsignal_client.so; let gradle do it for us

jrose-signal:
This is a step towards saving the debug info somewhere for the builds we actually ship on Android, but for now it's enough to have the unstripped libraries *available* for testing purposes.


jrose-signal:
Oops, I forgot I have to fix check_code_size for this…

jim-signal (approved):

PR #349, opened 2021-08-31
Sealed sender v2: add an InvalidRegistrationId exception/error

jrose-signal:
This dedicated error is thrown when a recipient has a registration ID that's out of the range used by Signal [0, 0x3FFF]. These IDs cannot be encoded in the sealed sender v2 format and are not supported, even though they don't cause any problems for 1:1 messages.

### DIFF
--- a/node/Errors.ts
+++ b/node/Errors.ts
@@ -3,17 +3,22 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+import { ProtocolAddress } from './Address';
+import * as Native from './Native';
+
 export enum ErrorCode {
   Generic,
 
   DuplicatedMessage,
   SealedSenderSelfSend,
   UntrustedIdentity,
+  InvalidRegistrationId,
 }
 
 export class SignalClientErrorBase extends Error {
   public readonly code: ErrorCode;
   public readonly operation: string;
+  readonly _addr?: string | Native.ProtocolAddress;
 
   constructor(
     message: string,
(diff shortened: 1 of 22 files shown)


swift/Sources/SignalClient/Error.swift
```diff
@@ -32,6 +32,7 @@ public enum SignalError: Error {
     case untrustedIdentity(String)
     case invalidKeyIdentifier(String)
     case sessionNotFound(String)
+    case invalidRegistrationId(address: ProtocolAddress, message: String)
```
michelle-signal, line 35: Swift API looks good for iOS 👍

jim-signal (approved):

peter-signal (approved):

greyson-signal (approved):

node/Errors.ts
```diff
@@ -41,27 +46,48 @@ export class SignalClientErrorBase extends Error {
...
-export type UntrustedIdentityError = SignalClientErrorBase & {
+export type UntrustedIdentityError = SignalClientErrorCommon & {
   code: ErrorCode.UntrustedIdentity;
   addr: string;
 };
 
+export type InvalidRegistrationIdError = SignalClientErrorCommon & {
+  code: ErrorCode.InvalidRegistrationId;
+  addr: ProtocolAddress;
+};
```
scottnonnenberg-signal, line 86: 👍

scottnonnenberg-signal (approved):

PR #351, opened 2021-09-02
protocol: Deserialize archived sessions lazily instead of eagerly

jrose-signal:
The only operation that touches archived sessions is decryption, and even then only in two cases: when checking to see if an incoming PreKey message matches an existing session, and after decrypting on the current session fails. Neither of those are unlikely, but they happen an order of magnitude less than normal decryptions (not to mention encryptions), so it's worth making the fast path faster.


jrose-signal:
Needs a dedicated benchmark; all our current benchmarks don't deal with archived sessions.

jim-signal (approved):

PR #362, opened 2021-09-23
Stop deriving curve25519 public keys when we don't need to

jrose-signal:
- Cache the computation for ephemeral keys in sealed sender v2
- Pass through the public key when we have a key pair already
- Stop deriving it and then ignoring it for low-level curve25519 operations


jrose-signal:
(reviewed privately)

PR #384, opened 2021-10-19
bridge: Simplify the jni_bridge_handle macro using a marker trait

jrose-signal:
Similar to #383, but for the JNI bridge.

### DIFF
--- a/rust/bridge/shared/src/jni/convert.rs
+++ b/rust/bridge/shared/src/jni/convert.rs
@@ -593,16 +593,14 @@ impl<T: ResultTypeInfo> ResultTypeInfo for SignalJniResult<T> {
     }
 }
 
-impl<T> ResultTypeInfo for Option<SignalJniResult<T>>
-where
-    Option<T>: ResultTypeInfo,
-{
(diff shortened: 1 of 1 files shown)


rust/bridge/shared/src/jni/convert.rs
```diff
@@ -625,98 +623,94 @@ impl crate::support::Env for &'_ JNIEnv<'_> {
...
+        }
+    }
+}
+
+impl<'a, T: BridgeHandle + 'static> SimpleArgTypeInfo<'a> for &mut T {
+    type ArgType = ObjectHandle;
+    fn convert_from(_env: &JNIEnv, foreign: Self::ArgType) -> SignalJniResult<Self> {
+        unsafe { native_handle_cast(foreign) }
+    }
+}
```
peter-signal, line 652: Why do all of the above Ts need to be 'static?

jrose-signal: Probably that should just be a requirement on BridgeHandle: if we're going to put something into a app language wrapper, we don't know how long it's going to live, and so it can't (safely) have references to anything with a local lifetime.

PR #386, opened 2021-10-20
bridge: Move as much out of the node_bridge_handle macro as possible

jrose-signal:
Unlike the FFI and JNI bridges (#383 and #384), the argument handling for wrapped Rust values in Node can't be handled with a simple blanket trait impl. Add a set of new traits and helpers to capture most of the complexity so that the macro doesn't have anything too complicated in it:

- BridgeHandle - mostly a marker trait, like `ffi::BridgeHandle` and `jni::BridgeHandle`, but has a Strategy associated type to choose between Immutable-style boxing and Mutable-style boxing.

- BridgeHandleStrategy - chooses between Immutable-style boxing (no extra wrapper) and Mutable-style boxing (wrap in a RefCell for safety). Can also be used to guard a particular function to only work with immutable or mutable bridge handles

- `BoxContentsFor<T>` - shorthand for `T::Strategy::JsBoxContents`, which Rust can't always resolve without fully spelling out the traits involved. In practice, this will either be `T` or `RefCell<T>`.

- BoxedBorrow - a struct that represents a synchronous borrow of a wrapped value. By making the borrow type a generic parameter, we can have `BoxedBorrow<&T>` (immutable borrow of an immutable value), `BoxedBorrow<Ref<T>>` (immutable borrow of a mutable value), and `BoxedBorrow<RefMut<T>>` (mutable borrow of a mutable value).

- PersistentBoxedValue - a struct for *asynchronously* borrowing a wrapped value (by keeping around its JavaScript wrapper as a GC root). This type already existed, but it's been tweaked so that it's more similar to BoxedBorrow and PersistentArrayOfBoxedValues.

- PersistentArrayOfBoxedValues - a struct for borrowing a whole array of boxed values and, uh, hoping JavaScript won't change the array out from under us. This also already existed, but the safety is more clearly documented now.

- `clone_from_wrapper` and `clone_from_array_of_wrappers` - functions to clone a bunch of boxed mutable values, so that they won't change during asynchronous use. Not the most efficient thing, but pretty straightforward, at least.

All of the convert.rs files also need some reorganizing and the contents of this commit message should probably turn into a module-level doc comment once it settles.


rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+    /// The `borrow` callback will be used to get an active reference for the Rust value.
+    /// For example, if the "box contents" for this handle use a [`RefCell`], `borrow` might
+    /// be [`RefCell::borrow_mut`] to get a mutable reference out. For a plain value, `borrow`
+    /// can be the identity function.
+    pub fn from_wrapper(
+        cx: &mut impl Context<'a>,
+        wrapper: Handle<'a, JsObject>,
+        borrow: fn(&'a BoxContentsFor<Borrow::Target>) -> Borrow,
+    ) -> NeonResult<Self> {
+        let boxed_value: Handle<'a, DefaultJsBox<BoxContentsFor<Borrow::Target>>> =
```
peter-signal, line 715: This would still be more clear to me if it were "js_boxed_bridge_handle" to make it clear it's a JsBox and not a normal Box, and we know the contents to be a BridgeHandle.  I know that's in the type, but the type is pretty complex.

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+            Object::get(*wrapper, cx, NATIVE_HANDLE_PROPERTY)?.downcast_or_throw(cx)?;
+        let box_contents: &BoxContentsFor<Borrow::Target> = &*boxed_value;
```
peter-signal, line 717: A name like "borrowed_js_box_contents" would be nice here.  Again, I know it's in the type, but the type is complex.

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+        // FIXME: Workaround for https://github.com/neon-bindings/neon/issues/678
+        // The lifetime of the boxed contents is necessarily longer than the lifetime of any handles
+        // referring to it, i.e. longer than 'a. However, Deref'ing a Handle can only give us a
+        // reference whose lifetime matches a *particular* handle. Therefore, we unsafely (in the
+        // compiler sense) extend the lifetime to be the lifetime of the context, as given by the
+        // Handle. (We also know the object can't move because the compiler can't know how many JS
+        // references there are referring to the JsBox, any of which another thread could
+        // be dereferencing.)
+        let contents_with_extended_lifetime: &'a BoxContentsFor<Borrow::Target> =
```
peter-signal, line 726: Same here with borrowed_js_box_contents_with_extended_lifetime.

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+            .upcast())
+    }
+}
+/// Accesses a boxed Rust value using a custom reference (such as [`std::cell::RefMut`]).
+///
+/// Once created, the value can be accessed using the usual [`Deref`] and [`DerefMut`] operations.
+///
+/// Also stores the JavaScript handle that owns the JsBox to ensure that the Rust value is kept
+/// alive.
+pub struct BoxedBorrow<'a, Borrow: Deref + 'a>
```
peter-signal, line 691: I know it's explicit, but I think this would be much more clear with a name like BorrowedJsBoxedBridgeHandle.  It took me a long time reading this code to figure out what was going on, and I think that name would have helped a lot. In particular, to make it clear that we're borrowing a BridgeHandle that's in a JsBox.

peter-signal: Similarly, I know it's explicit, but instead of "Borrow" being the type name, having it be  BorrowedBoxHandle would have made this a lot more clear.

peter-signal: I think the comment for the class would be more clear with something like

"Contains a JsBoxed BridgeHandle along with a Borrowed version of it.  Can technically borrow from anything of Deref<Target=BridgeHandle> but in practice borrows from a &BridgeHandle,  a RefCell<BridgeHandle>::Ref,  or a RefCell<BridgeHandle>::MutRef"

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+where
+    Borrow::Target: BridgeHandle,
+{
+    _owner: Handle<'a, DefaultJsBox<BoxContentsFor<Borrow::Target>>>,
+    borrow: Borrow,
```
peter-signal, line 696: And here a name like "borrowed_bridge_handle" would be more clear as well.

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+///
+/// Once created, the value can be accessed using the usual [`Deref`] and [`DerefMut`] operations.
+///
+/// Also stores the JavaScript handle that owns the JsBox to ensure that the Rust value is kept
+/// alive.
+pub struct BoxedBorrow<'a, Borrow: Deref + 'a>
+where
+    Borrow::Target: BridgeHandle,
+{
+    _owner: Handle<'a, DefaultJsBox<BoxContentsFor<Borrow::Target>>>,
```
peter-signal, line 695: I think a name like _js_boxed_bridge_handle would be more clear here.  I know it's technically in the type, but it's pretty complex type.

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+        // referring to it, i.e. longer than 'a. However, Deref'ing a Handle can only give us a
+        // reference whose lifetime matches a *particular* handle. Therefore, we unsafely (in the
+        // compiler sense) extend the lifetime to be the lifetime of the context, as given by the
+        // Handle. (We also know the object can't move because the compiler can't know how many JS
+        // references there are referring to the JsBox, any of which another thread could
+        // be dereferencing.)
+        let contents_with_extended_lifetime: &'a BoxContentsFor<Borrow::Target> =
+            unsafe { extend_lifetime(box_contents) };
+        Ok(Self {
+            _owner: boxed_value,
```
peter-signal, line 729: _js_boxed_bridge_handle: js_boxed_bridge_handle
borrowed_bridge_handle: borrowed_js_box_contents_with_extended_lifetime

Would be easier to understand and read here.

rust/bridge/shared/src/node/convert.rs
```diff
@@ -731,182 +858,152 @@ impl<T: Send + Sync + 'static> Finalize for PersistentArrayOfBoxedValues<T> {
...
+
+        // ArgTypeInfo already has a blanket impl (SimpleArgTypeInfo),
+        // so we have to declare this explicitly.
         impl<'storage, 'context: 'storage> node::ArgTypeInfo<'storage, 'context>
-        for &'storage $typ {
+            for &'storage $typ
+        {
             type ArgType = node::JsObject;
-            type StoredType = node::Handle<'context, node::DefaultJsBox<$typ>>;
+            type StoredType = node::BoxedBorrow<'context, &'context $typ>;
```
peter-signal, line 924: BorrowedJsBoxedBridgeHandle here would make it much more clear what's going on.

rust/bridge/shared/src/node/convert.rs
jrose-signal: That one I'm less sure about. It's not *just* the handle that's here, it's also whatever guard is being used to borrow it. But maybe that's implied by the type?

rust/bridge/shared/src/node/convert.rs
```diff
@@ -632,47 +631,170 @@ pub(crate) unsafe fn extend_lifetime<'a, 'b: 'a, T: ?Sized>(some_ref: &'a T) ->
...
+}
+
+/// Mutable handles use a [`RefCell`] for safety.
+impl<T: BridgeHandle> BridgeHandleStrategy<T> for Mutable<T> {
+    type JsBoxContents = RefCell<T>;
+}
+
+/// Shorthand for `T::Strategy::JsBoxContents` when `T: `[`BridgeHandle`], which Rust can't always
+/// resolve.
+type BoxContentsFor<T> = <<T as BridgeHandle>::Strategy as BridgeHandleStrategy<T>>::JsBoxContents;
```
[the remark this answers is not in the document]
jrose-signal: Some concrete examples:

```
BoxContentsFor<PublicKey> = PublicKey
BoxContentsFor<HSMEnclaveClient> = RefCell<HSMEnclaveClient>
```

"Of" doesn't sound as good to me because that feels like it's drilling into the type instead of wrapping it in something. But ultimately this is bikeshedding and whatever makes more sense to you is probably best.

PR #403, opened 2021-10-29
bridge: Add zkgroup APIs

jrose-signal:
These APIs are designed to match the generated "simpleapi" entry points in [the original zkgroup repository](/signalapp/zkgroup), to make it easier to adapt the existing Java, Swift, and TypeScript code to libsignal-client.

The cbindgen-generated signal_ffi.h now includes constants, so that the fixed-size arrays used to serialize zkgroup types can use named constants in Rust. This meant filtering out some constants that were getting picked up but that should not be included.

Note that this commit makes references to Java exception types that will be added in a later commit.

### DIFF
--- a/rust/bridge/ffi/cbindgen.toml
+++ b/rust/bridge/ffi/cbindgen.toml
@@ -17,8 +17,10 @@ style = "type"
 prefix_with_name = true
 
 [export]
-include = ["SignalErrorCode", "FfiDirection", "FfiCiphertextMessageType", "FfiContentHint"]
-item_types = ["enums", "functions", "opaque", "structs", "typedefs"]
+include = ["SignalErrorCode", "FfiDirection", "FfiCiphertextMessageType", "FfiContentHint", "RandomnessBytes"]
+exclude = ["TAG_SIZE", "NONCE_SIZE"]
+item_types = ["enums", "functions", "opaque", "structs", "typedefs", "constants"]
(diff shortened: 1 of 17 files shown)


rust/bridge/ffi/cbindgen.toml
```diff
@@ -17,8 +17,10 @@ style = "type"
 prefix_with_name = true
 
 [export]
-include = ["SignalErrorCode", "FfiDirection", "FfiCiphertextMessageType", "FfiContentHint"]
-item_types = ["enums", "functions", "opaque", "structs", "typedefs"]
+include = ["SignalErrorCode", "FfiDirection", "FfiCiphertextMessageType", "FfiContentHint", "RandomnessBytes"]
+exclude = ["TAG_SIZE", "NONCE_SIZE"]
+item_types = ["enums", "functions", "opaque", "structs", "typedefs", "constants"]
+# FIXME: this doesn't work well with constants in SCREAMING_SNAKE_CASE
```
peter-signal, line 23: What is this referring to?

peter-signal (approved):

rust/bridge/ffi/cbindgen.toml
jrose-signal: The following line, `prefix = "Signal"`. Constants end up as `SignalFOO_BAR` rather than `SIGNAL_FOO_BAR`. But since the C API isn't an end product I decided it wasn't worth spending time on.

jim-signal (approved):

PR #405, opened 2021-11-01
bridge: Add back support for bridging u64, now losslessly

jrose-signal:
- FFI: still maps trivially to uint64_t, no change
- JNI: reinterpret as jlong (i64) to not lose any information
- Node: until Neon supports bigints, convert to/from a big-endian Buffer

Follow-up to #404.

### DIFF
--- a/rust/bridge/shared/src/jni/convert.rs
+++ b/rust/bridge/shared/src/jni/convert.rs
@@ -164,6 +164,14 @@ impl<'a> SimpleArgTypeInfo<'a> for Option<u32> {
     }
 }
 
+/// Reinterprets the bits of the Java `long` as a `u64`.
+impl<'a> SimpleArgTypeInfo<'a> for u64 {
+    type ArgType = jlong;
+    fn convert_from(_env: &JNIEnv, foreign: jlong) -> SignalJniResult<Self> {
+        Ok(foreign as u64)
+    }
+}
+
 /// Supports values `0..=Long.MAX_VALUE`.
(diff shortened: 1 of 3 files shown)


rust/bridge/shared/src/jni/convert.rs
```diff
@@ -164,6 +164,14 @@ impl<'a> SimpleArgTypeInfo<'a> for Option<u32> {
     }
 }
 
+/// Reinterprets the bits of the jlong as a u64.
```
jim-signal, line 167: Do you want it to be consistent with others like:

Reinterprets the bits of the Java \`long\` as a \`u64\`.

jim-signal (approved):
What is the plan for node bigint support?

jrose-signal:
> What is the plan for node bigint support?

If we decide it's important I could spend some time doing the necessary work on the Neon side; if not, converting from Buffer to bigint on the TypeScript side isn't bad, and converting from bigint to Buffer…probably deserves a helper function. Here's what it looks like in the TS zkgroup port:

```typescript
if (receiptExpirationTime < 0 || receiptExpirationTime > UINT64_MAX) {
  throw new RangeError(`invalid receipt expiration time ${receiptExpirationTime}`);
}
const receiptExpirationAsBuffer = Buffer.alloc(8);
receiptExpirationAsBuffer.writeBigUInt64BE(receiptExpirationTime);
```

peter-signal (approved):

PR #412, opened 2021-11-05
swift: Improve zkgroup's ByteArray helper class

jrose-signal:
Don't validate sizes explicitly; they're already provided in the signatures of the *_check_valid_contents functions exposed in the FFI layer. Similar to #408.


jim-signal (approved):

peter-signal (approved):
Nice

PR #416, opened 2021-11-05
zkgroup: Move blob padding/unpadding into Rust

jrose-signal:
Previously this was defined in the app layers, because zkgroup's original codegen didn't support custom exception types. However, we can now move it to a common implementation in Rust.

~~This could live in libsignal-bridge, or in the zkgroup crate proper. I chose libsignal-bridge because padding isn't an inherent requirement for the zkgroup encryption format, but it could go either way.~~

### DIFF
--- a/java/java/src/main/java/org/signal/client/internal/Native.java
+++ b/java/java/src/main/java/org/signal/client/internal/Native.java
@@ -160,11 +160,11 @@ public final class Native {
   public static native byte[] GroupPublicParams_GetGroupIdentifier(byte[] groupPublicParams);
 
   public static native void GroupSecretParams_CheckValidContents(byte[] obj);
-  public static native byte[] GroupSecretParams_DecryptBlob(byte[] params, byte[] ciphertext);
+  public static native byte[] GroupSecretParams_DecryptBlobWithPadding(byte[] params, byte[] ciphertext);
   public static native byte[] GroupSecretParams_DecryptProfileKey(byte[] params, byte[] profileKey, UUID uuid);
   public static native UUID GroupSecretParams_DecryptUuid(byte[] params, byte[] uuid);
   public static native byte[] GroupSecretParams_DeriveFromMasterKey(byte[] masterKey);
-  public static native byte[] GroupSecretParams_EncryptBlobDeterministic(byte[] params, byte[] randomness, byte[] plaintext);
+  public static native byte[] GroupSecretParams_EncryptBlobWithPaddingDeterministic(byte[] params, byte[] randomness, byte[] plaintext, int paddingLen);
   public static native byte[] GroupSecretParams_EncryptProfileKey(byte[] params, byte[] profileKey, UUID uuid);
   public static native byte[] GroupSecretParams_EncryptUuid(byte[] params, UUID uuid);
   public static native byte[] GroupSecretParams_GenerateDeterministic(byte[] randomness);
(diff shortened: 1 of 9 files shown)


java/java/src/main/java/org/signal/zkgroup/groups/ClientZkGroupCipher.java
```diff
@@ -62,35 +62,13 @@ public ProfileKey decryptProfileKey(ProfileKeyCiphertext profileKeyCiphertext, U
...
-
-    byte[] paddedPlaintext = new byte[plaintext.length + 4];
-    System.arraycopy(plaintext, 0, paddedPlaintext, 4, plaintext.length);
-
     byte[] random      = new byte[RANDOM_LENGTH];
-
     secureRandom.nextBytes(random);
-
-    return Native.GroupSecretParams_EncryptBlobDeterministic(groupSecretParams.getInternalContentsForJNI(), random, paddedPlaintext);
+    return Native.GroupSecretParams_EncryptBlobWithPaddingDeterministic(groupSecretParams.getInternalContentsForJNI(), random, plaintext, 0);
```
jim-signal, line 67: Is 0 here the padding length (so no padding) or byte value to pad with? Seems like it should be 4?

jrose-signal: It's the padding length. Specifically, the way this encoding works is that it has a variable number of *trailing* padding bytes, and the number of trailing padding bytes is encoded as a big-endian 32-bit integer prepended to the actual message. The "4" in the old code is just `sizeof(u32)`.

jrose-signal:
> I chose libsignal-bridge because padding isn't an inherent requirement for the zkgroup encryption format, but it could go either way.

I think I've changed my mind on this; putting non-trivial behavior in libsignal-bridge means people have to pay more attention to libsignal-bridge when something goes wrong. Normally I'd look at the app-level interface or the Rust implementation first, then the other, before looking at the bridge layer. Update coming shortly.

jim-signal (approved):

peter-signal (approved):

java/java/src/main/java/org/signal/client/internal/Native.java
```diff
@@ -160,11 +160,11 @@ private Native() {}
   public static native byte[] GroupPublicParams_GetGroupIdentifier(byte[] groupPublicParams);
 
   public static native void GroupSecretParams_CheckValidContents(byte[] obj);
-  public static native byte[] GroupSecretParams_DecryptBlob(byte[] params, byte[] ciphertext);
+  public static native byte[] GroupSecretParams_DecryptBlobWithPadding(byte[] params, byte[] ciphertext);
   public static native byte[] GroupSecretParams_DecryptProfileKey(byte[] params, byte[] profileKey, UUID uuid);
   public static native UUID GroupSecretParams_DecryptUuid(byte[] params, byte[] uuid);
   public static native byte[] GroupSecretParams_DeriveFromMasterKey(byte[] masterKey);
-  public static native byte[] GroupSecretParams_EncryptBlobDeterministic(byte[] params, byte[] randomness, byte[] plaintext);
+  public static native byte[] GroupSecretParams_EncryptBlobWithPaddingDeterministic(byte[] params, byte[] randomness, byte[] plaintext, int padding);
```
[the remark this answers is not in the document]
jrose-signal: Discussed with Trevor, and it sounds like we might want to expose the padding functionality at some point, so I'm going to leave this wired up (but change the name as Peter suggests).

PR #421, opened 2021-11-15
Enable asm implementations for sha1 and sha2

jrose-signal:


jrose-signal:
Worth a shot, but looks like [RustCrypto/asm-hashes]() isn't set up to compile on all the platforms we care about.

PR #425, opened 2021-11-28
Update to prost 0.9

rubdos:
Prost 0.9 contains some performance improvement for varint, and allows developers to work on Apple silicon.

https://github.com/tokio-rs/prost/releases/tag/v0.9.0

Technically this is a breaking change (because `prost::DecodeError` is exposed through `SignalProtocolError`), so you'll want to bump to 0.12 after this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.44"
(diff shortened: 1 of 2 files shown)


jrose-signal:
Thank you! As an aside, the version numbers are more for Signal-internal use, so they track the stability of the Java, Swift, and TypeScript interfaces rather than the Rust crate (which is why that's stuck at 0.1.0).

rubdos:
> Thank you! As an aside, the version numbers are more for Signal-internal use, so they track the stability of the Java, Swift, and TypeScript interfaces rather than the Rust crate (which is why that's stuck at 0.1.0).

Aha, I see. Maybe I should make a separate issue about this, but would it at a certain point be possible to start versioning the crate as well?

jrose-signal:
Circling back to this: after some discussion we're going to do best-effort versioning for the (non-bridge) Rust APIs as well, though for now I think we'll continue treating the version number as a *workspace* version number rather than tracking things on a per-crate level. (That is, a change in signal-crypto's API would bump the overall version number, even if libsignal-protocol didn't change.)

rubdos:
That sounds very usable to me already :-)

PR #426, opened 2021-12-03
NPM: Publish releases for M1 Macs too

jrose-signal:
This requires building on the macOS 11 GitHub runner; the 10.15 runner isn't set up for cross-compiling when not using Xcode.

Also, ensure that when we test the NPM package before publishing, we're using the libraries we're going to submit, rather than ones built on the final runner.

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -22,7 +22,10 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-11]
+        include:
+        - os: macos-11
+          additional-rust-target: aarch64-apple-darwin
 
     steps:
     - uses: actions/checkout@v2
@@ -33,6 +36,7 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
+        target: ${{ matrix.additional-rust-target }}
 
     - name: Get Node version from .nvmrc
       id: get-nvm-version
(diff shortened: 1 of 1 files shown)


jrose-signal:
Thanks again to @dennisameling for getting the initial multi-arch compilation working!

jim-signal (approved):

dennisameling:
@jrose-signal thanks a lot for this! 🎉 May I ask for Windows ARM64 to be added as well? I've been releasing builds of Signal Desktop [since the beginning of June](https://github.com/dennisameling/Signal-Desktop/releases) (and [building the native deps here](https://github.com/dennisameling/signal-native-deps/actions/runs/1529998521)). It's been working very well and stable. Happy to provide a PR if that's OK for you 🙏🏼

jrose-signal:
I'll circle back to this after New Year's, but discussion with the Desktop folks is leaning positive about it.

balupton:
@dennisameling can we get the momentum going and add a Linux ARM64 PR as well? 🙇‍♂️🙇‍♂️🙇‍♂️ ❤️❤️❤️

jrose-signal:
@dennisameling We've discussed this a bit more at Signal and while we probably won't be doing an official ARM64 Linux app any time soon, there doesn't seem to be much harm in enabling it here too. Then you can drop it completely from your unofficial dependencies (well, unless you're building for an even more unusual platform since we last spoke!).

dennisameling:
Thanks! Here's a PR for adding ARM64 Linux prebuilds:

https://github.com/signalapp/libsignal-client/pull/450

PR #429, opened 2021-12-10
GitHub: Move fuzz target checking to the stable Rust toolchain job

jrose-signal:
...because the fuzz target doesn't have a lockfile, and crates don't generally support arbitrary nightly versions. (Specifically, we just hit dtolnay/proc-macro2#307.)


jim-signal (approved):
LGTM

PR #432, opened 2021-12-13
Treat InvalidMacKeyLength as InvalidKeyException

jrose-signal:
...not a generic RuntimeException. Now that it's only used for SignalMessage MAC keys (cf. #430), the only way it could be wrong is if it's provided incorrectly by the user.


jim-signal (approved):

jrose-signal:
Merged as 57c1de64.

PR #433, opened 2021-12-17
Swift: Expose AES-GCM and AES-CTR implementations

jrose-signal:
These have been in the Rust layer for a while, but not exposed as public API to Swift, though Android does use them.

Additionally, clean up the existing AES-GCM-SIV API a little. This is technically a breaking change, but the iOS app isn't actually using SignalClient's Aes256GcmSiv for anything at the moment.


jrose-signal:
Closing for now. We may come back to this later though.

PR #435, opened 2021-12-20
Add signAlternateIdentity and verifyAlternateIdentity operations

jrose-signal:
This can be used for one identity to attest that another identity represents the same account. For instance, a PNI can sign an ACI to demonstrate that the ACI is associated with the phone number in question.


jrose-signal:
See also #434.

PR #436, opened 2021-12-25
rust: bridge: shared: macros: fix parameter typo

ZabeMath:


jrose-signal:
Thank you! This change is trivial (in both a functionality and copyright sense), so I'm just going to merge it.

PR #440, opened 2022-01-11
node: Add Windows ARM64 prebuilds to CI

dennisameling:
This PR adds prebuilds for Windows ARM64 to CI. I've been [building these in my fork](https://github.com/dennisameling/libsignal-node/tree/main/prebuilds/win32-arm64) for a few months and haven't encountered issues with it.

As discussed in https://github.com/signalapp/libsignal-client/pull/426#issuecomment-998233937, "discussion with the Desktop folks is leaning positive about it", so I took the liberty to create this PR 😊

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -25,7 +25,9 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-11]
         include:
         - os: macos-11
-          additional-rust-target: aarch64-apple-darwin
+          arm64-rust-target: aarch64-apple-darwin
+        - os: windows-latest
+          arm64-rust-target: aarch64-pc-windows-msvc
 
     steps:
     - uses: actions/checkout@v2
@@ -36,7 +38,7 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        target: ${{ matrix.additional-rust-target }}
+        target: ${{ matrix.arm64-rust-target }}
 
     - name: Get Node version from .nvmrc
       id: get-nvm-version
@@ -56,7 +58,7 @@ jobs:
 
     - run: npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }} --arch arm64
       working-directory: node
-      if: matrix.os == 'macos-11'
+      if: matrix.arm64-rust-target != ''
 
     - run: npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }}
       working-directory: node


.github/workflows/npm.yml
```diff
@@ -56,7 +58,7 @@ jobs:
 
     - run: npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }} --arch arm64
       working-directory: node
-      if: matrix.os == 'macos-11'
+      if: matrix.os != 'ubuntu-latest'
```
jrose-signal, line 61: Let's make this a little more semantic: we can rename `additional-rust-target` to `arm64-rust-target` and then this can check whether `arm64-rust-target` is present at all.

jrose-signal (commented):
Thanks for the ping! Yes, we can take this, but I'll note this workflow is just for the manually-submitted job to publish to NPM; we're still not compiling for aarch64-apple-darwin or aarch64-pc-windows-msvc on every PR. With macOS I didn't worry about it because we *do* compile the non-bridge crates for aarch64-apple-ios; do you feel confident that aarch64-pc-windows-msvc is sufficiently exercised by the other AArch64 platforms + AMD64 Windows?

dennisameling:
> do you feel confident that aarch64-pc-windows-msvc is sufficiently exercised by the other AArch64 platforms + AMD64 Windows?

I would say yes. I've been [building this in my fork](https://github.com/dennisameling/libsignal-node/tree/main/prebuilds/win32-arm64) for a few months from a x64 host and all is good. `aarch64-pc-windows-msvc` is a [Tier 2 with Host Tools](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools) platform, just like `aarch64-apple-darwin`, so it's also well-supported from the Rust side.

> we're still not compiling for aarch64-apple-darwin or aarch64-pc-windows-msvc on every PR

Would you like me to add `aarch64-pc-windows-msvc` to the regular PR pipeline as well? Shall we simply use `npx prebuildify` for that or would you prefer a different approach? I'm happy to sponsor an [ECS LIVA QC710](https://www.microsoft.com/en-us/d/ecs-liva-mini-box-qc710-desktop/8z247h1h3skp) if you want to run CI tests on Windows ARM64 as well, or would you feel comfortable enough cross-compiling from x64 for now (which is the only architecture for which GH Actions has hosted runners currently)?

.github/workflows/npm.yml
dennisameling: Makes sense! Updated in https://github.com/signalapp/libsignal-client/pull/440/commits/a2b1ae1d960076fbcf46b576d2f038fa5d2ac53d - seems to work correctly: https://github.com/dennisameling/libsignal-client/runs/4893965003?check_suite_focus=true

jrose-signal:
Let's leave the per-PR CI as it is for now, then. Setting up a custom GitHub Actions runner is one more thing we'd have to keep running, and without that we'd just be testing compilation, which is pretty unlikely to break *just* for ARM64 Windows.

Thank you for doing this!

jrose-signal:
Just published v0.12.1 with an ARM64 Windows prebuild. :-)

PR #443, opened 2022-02-08
Update to string encoding for fingerprint

GrishaVar:
I noticed there was a todo left in the comments :)

Should `read5_mod_100k` perhaps use `u64::from_be_bytes`? It would be more readable/understandable.

### DIFF
--- a/rust/protocol/src/fingerprint.rs
+++ b/rust/protocol/src/fingerprint.rs
@@ -41,8 +41,11 @@ fn get_encoded_string(fprint: &[u8]) -> Result<String> {
     }
 
     let s = fprint.chunks_exact(5).take(6).map(read5_mod_100k).fold(
-        String::with_capacity(5*6),
-        |mut s, n| {write!(s, "{:05}", n).ok(); s},
+        String::with_capacity(5 * 6),
+        |mut s, n| {
+            write!(s, "{:05}", n).expect("can always write to a Vec");
+            s
+        },
     );
 
     Ok(s)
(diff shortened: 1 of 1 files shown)


rust/protocol/src/fingerprint.rs
```diff
@@ -35,19 +36,13 @@ fn get_encoded_string(fprint: &[u8]) -> Result<String> {
...
-        "{:05}{:05}{:05}{:05}{:05}{:05}",
-        read5_mod_100k(&fprint[0..5]),
-        read5_mod_100k(&fprint[5..10]),
-        read5_mod_100k(&fprint[10..15]),
-        read5_mod_100k(&fprint[15..20]),
-        read5_mod_100k(&fprint[20..25]),
-        read5_mod_100k(&fprint[25..30])
+    let s = fprint.chunks_exact(5).take(6).map(read5_mod_100k).fold(
+        String::with_capacity(5*6),
+        |mut s, n| {write!(s, "{:05}", n).ok(); s},
```
jrose-signal, line 45: This `ok()` should be `expect("…")`, so we don't ignore errors.

jrose-signal (commented):
Thank you! Just one comment. (And probably a `cargo fmt`.)

rust/protocol/src/fingerprint.rs
```diff
@@ -41,8 +41,11 @@ fn get_encoded_string(fprint: &[u8]) -> Result<String> {
     }
 
     let s = fprint.chunks_exact(5).take(6).map(read5_mod_100k).fold(
-        String::with_capacity(5*6),
-        |mut s, n| {write!(s, "{:05}", n).ok(); s},
+        String::with_capacity(5 * 6),
+        |mut s, n| {
+            write!(s, "{:05}", n).expect("can always write to a Vec");
```
jrose-signal, line 46: Nitpick: String, not Vec!

GrishaVar: Oh, sorry! I copied the message from a different expect

GrishaVar:
Yep, I didn't add it because my conversion to `[u8; 8]` less readable than the current code.

jrose-signal:
Thank you again!

PR #447, opened 2022-02-10
Node: Don't export initLogger and LogLevel directly from Native.d.ts

jrose-signal:
Native.d.ts only applies in TypeScript mode, not plain JavaScript. This also confuses esbuild. Rather than try to fight that, return to having Native not be exported at all, using a translation layer.


indutny-signal (approved):
LGTM

PR #450, opened 2022-02-17
NPM: publish releases for Linux arm64

dennisameling:
As mentioned in https://github.com/signalapp/libsignal-client/pull/426#issuecomment-1035334639

Linux is a bit of a special case in that we need to explicitly define things like the arm64 linker. Without the cross dependencies, I got linker errors: https://github.com/dennisameling/libsignal-client/runs/5229605777?check_suite_focus=true

Example of a successful build: https://github.com/dennisameling/libsignal-client/actions/runs/1858014293

I've been using this setup in my `signal-native-deps` repo for a few months: https://github.com/dennisameling/signal-native-deps/blob/75f22a11a11d18b1ed3b85b51d032816579cce00/.github/workflows/build.yaml#L132-L140

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -28,6 +28,8 @@ jobs:
           arm64-rust-target: aarch64-apple-darwin
         - os: windows-latest
           arm64-rust-target: aarch64-pc-windows-msvc
+        - os: ubuntu-latest
+          arm64-rust-target: aarch64-unknown-linux-gnu
 
     steps:
     - uses: actions/checkout@v2
@@ -56,9 +58,14 @@ jobs:
     - run: yarn install --ignore-scripts --frozen-lockfile
       working-directory: node
 
-    - run: npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }} --arch arm64
+    - run: |
+        if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+          sudo apt-get install -y crossbuild-essential-arm64
+          export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
+        fi
+        npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }} --arch arm64
       working-directory: node
-      if: matrix.arm64-rust-target != ''
+      shell: bash
 
     - run: npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }}
       working-directory: node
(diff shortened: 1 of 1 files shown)


.github/workflows/npm.yml
```diff
@@ -56,9 +58,14 @@ jobs:
     - run: yarn install --ignore-scripts --frozen-lockfile
       working-directory: node
 
-    - run: npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }} --arch arm64
+    - run: |
+        if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+          sudo apt-get install -y crossbuild-essential-arm64
+          export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
+        fi
```
jrose-signal, line 65: Nitpick: I think this would be clearer as its own build step (but the actual command seems fine)

jrose-signal (commented):
I'm happy it's that easy! Your initial PR for cross-compilation support did a great job at laying this groundwork.

.github/workflows/npm.yml
dennisameling: Updated - thanks for the feedback!

Here's another example run to confirm that it works: https://github.com/dennisameling/libsignal-client/runs/5239772453?check_suite_focus=true

.github/workflows/npm.yml
```diff
@@ -58,14 +58,14 @@ jobs:
...
-    - run: |
-        if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
-          sudo apt-get install -y crossbuild-essential-arm64
-          export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
-        fi
-        npx prebuildify --napi -t ${{ steps.get-nvm-version.outputs.node-version }} --arch arm64
+    - name: Install crossbuild for Linux arm64
+      run: |
+        sudo apt-get install -y crossbuild-essential-arm64
+        echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
```
jrose-signal, line 64: Hmm, "This file is unique to the current step and changes for each step in a job." But we could just add this environment variable unconditionally to the next step, it's only going to affect arm64 cross-compiling and doesn't have any paths in it.

jrose-signal (approved):

jrose-signal:
Thank you! I'm planning to do a new tag pretty soon, too, so that build will include the arm64 Linux prebuild.

.github/workflows/npm.yml
dennisameling: Haha yeah I also got a bit confused by the docs - tested it in my example run and the env variable was indeed added to all next steps after it was set :)

PR #466, opened 2022-06-27
make doctests fail on any warnings

cosmicexplorer:
This is a useful setting to have on. We only have one doctest right now (for `ProtocolAddress::new()`), but this is useful when adding documentation in the constituent PRs of #467.

### DIFF
--- a/rust/protocol/src/lib.rs
+++ b/rust/protocol/src/lib.rs
@@ -1,5 +1,5 @@
 //
-// Copyright 2020-2021 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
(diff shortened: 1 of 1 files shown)


jrose-signal:
I'm leery of this for the same reason we don't use `deny(warnings)`: it makes it harder to use different compiler releases (even if this one doesn't affect downstream clients the same way). Is there a way to set this from the command line instead, so we can put it in our CI config but leave local builds alone?

cosmicexplorer:
I totally hadn't considered how similar this is to `deny(warnings)` in general, and it's not clear how `rustdoc` can be made to read from `cargo` config, since `#[doc]` attributes do not seem to obviously have a way to be set except in the source code itself (looking at https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#showing-warnings-in-doctests). I think I'll close this for now as I have had a lot of difficulty trying to figure this out and there doesn't seem to be a lot of benefit from it.

PR #468, opened 2022-06-27
add docs to storage module

cosmicexplorer:
*Split out from #287.*

Add docstrings to `storage.rs`.

### DIFF
--- a/rust/protocol/src/storage/inmem.rs
+++ b/rust/protocol/src/storage/inmem.rs
@@ -1,36 +1,46 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Implementations for stores defined in [super::traits].
+//!
+//! These implementations are purely in-memory, and therefore most likely useful for testing
+//! right now.
+
+use crate::storage::{traits, Context};
 use crate::{
     IdentityKey, IdentityKeyPair, PreKeyId, PreKeyRecord, ProtocolAddress, Result, SenderKeyRecord,
     SessionRecord, SignalProtocolError, SignedPreKeyId, SignedPreKeyRecord,
 };
-
-use crate::storage::{traits, Context};
 
 use async_trait::async_trait;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use uuid::Uuid;
 
+/// Reference implementation of [traits::IdentityKeyStore].
 #[derive(Clone)]
 pub struct InMemIdentityKeyStore {
     key_pair: IdentityKeyPair,
-    id: u32,
+    registration_id: u32,
     known_keys: HashMap<ProtocolAddress, IdentityKey>,
 }
 
 impl InMemIdentityKeyStore {
-    pub fn new(key_pair: IdentityKeyPair, id: u32) -> Self {
+    /// Create a new instance.
+    ///
+    /// `key_pair` corresponds to [traits::IdentityKeyStore::get_identity_key_pair], and
+    /// `registration_id` corresponds to [traits::IdentityKeyStore::get_local_registration_id].
+    pub fn new(key_pair: IdentityKeyPair, registration_id: u32) -> Self {
         Self {
             key_pair,
-            id,
+            registration_id,
             known_keys: HashMap::new(),
         }
     }
 
+    /// Clear the mapping of known keys.
     pub fn reset(&mut self) {
         self.known_keys.clear();
     }
@@ -43,7 +53,7 @@ impl traits::IdentityKeyStore for InMemIdentityKeyStore {
     }
 
     async fn get_local_registration_id(&self, _ctx: Context) -> Result<u32> {
-        Ok(self.id)
+        Ok(self.registration_id)
     }
 
     async fn save_identity(
@@ -94,12 +104,14 @@ impl traits::IdentityKeyStore for InMemIdentityKeyStore {
     }
 }
 
+/// Reference implementation of [traits::PreKeyStore].
 #[derive(Clone)]
 pub struct InMemPreKeyStore {
     pre_keys: HashMap<PreKeyId, PreKeyRecord>,
 }
 
 impl InMemPreKeyStore {
+    /// Create an empty pre-key store.
     pub fn new() -> Self {
         Self {
             pre_keys: HashMap::new(),
@@ -141,12 +153,14 @@ impl traits::PreKeyStore for InMemPreKeyStore {
     }
 }
 
+/// Reference implementation of [traits::SignedPreKeyStore].
 #[derive(Clone)]
 pub struct InMemSignedPreKeyStore {
     signed_pre_keys: HashMap<SignedPreKeyId, SignedPreKeyRecord>,
 }
 
 impl InMemSignedPreKeyStore {
+    /// Create an empty signed pre-key store.
     pub fn new() -> Self {
         Self {
             signed_pre_keys: HashMap::new(),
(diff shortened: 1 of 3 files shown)


cosmicexplorer:
cc @gferon @rubdos

rust/protocol/src/storage/inmem.rs
```diff
@@ -1,38 +1,52 @@
...
-use crate::storage::traits;
-use crate::storage::Context;
-
 use async_trait::async_trait;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use uuid::Uuid;
 
 #[derive(Clone)]
+/// Reference implementation of [traits::IdentityKeyStore].
```
jrose-signal, line 29: Nitpick: Doc comments before attributes looks a little tidier to me.

rust/protocol/src/storage/inmem.rs
```diff
@@ -310,6 +334,8 @@ pub struct InMemSignalProtocolStore {
 }
 
 impl InMemSignalProtocolStore {
+    /// Create an object with the minimal implementation of [traits::ProtocolStore], representing
+    /// the given identity `key_pair` and with the separate random seed `registration_id`.
```
jrose-signal, line 338: "random seed" is a little funny. "randomly chosen", maybe?

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,72 @@
...
+    sender_keys::SenderKeyRecord,
+    state::{PreKeyId, PreKeyRecord, SessionRecord, SignedPreKeyId, SignedPreKeyRecord},
+    IdentityKey, IdentityKeyPair,
 };
 
+/// Handle to FFI-provided context object.
+///
+/// This object is not manipulated in Rust, but is instead passed back to each FFI
+/// method invocation. This argument should just be [None] for all clients of the Rust-only API.
 pub type Context = Option<*mut std::ffi::c_void>;
```
jrose-signal, line 23: I have a personal TODO to do something about this but I'm not sure what yet. :-/

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,72 @@
...
 
+/// Each Signal message can be considered to have exactly two participants, a sender and receiver.
+///
+/// [IdentityKeyStore::is_trusted_identity] uses this to ensure the identity provided is configured
+/// for the appropriate role.
+///
+/// *TODO: consider moving this enum into [crate::protocol]?*
```
jrose-signal, line 30: Hrm. Does not belong in a doc comment, but a good point. On the other hand, it really is a generic concept: it's mostly used here, but it's used for internal hygiene in the sealed sender implementation as well. Maybe it should just go in `util`.

rust/protocol/src/storage/traits.rs
```diff
@@ -1,38 +1,72 @@
...
+    /// Return the single specific identity the store is assumed to represent, with private key.
     async fn get_identity_key_pair(&self, ctx: Context) -> Result<IdentityKeyPair>;
 
+    /// Return a [u32] specific to this store instance.
+    ///
+    /// This local registration id is separate from the per-device identifier used in
+    /// [ProtocolAddress] and should not change run over run.
+    ///
+    /// If the same *device* is unregistered, then registers again, the [ProtocolAddress::device_id]
+    /// will be the same, but the store registration id returned by this method will be regenerated.
```
jrose-signal, line 55: "may be the same" and "should be regenerated", but otherwise well summarized. :-)

rust/protocol/src/storage/traits.rs
```diff
@@ -94,8 +145,10 @@ pub trait SessionStore {
     ) -> Result<()>;
 }
 
+/// Interface for storing sender key records for other users, allowing multiple keys per user.
```
jrose-signal, line 148: We actually use this to store our own records too, using our own address as the `sender`.

rust/protocol/src/storage/traits.rs
```diff
@@ -94,8 +145,10 @@ pub trait SessionStore {
...
 #[async_trait(?Send)]
 pub trait SenderKeyStore {
+    /// Set the entry for `(sender, distribution_id)` to the value of `record`.
```
jrose-signal, line 151: "the value of" is kind of redundant. An in-memory store might save the object as is (cloning it); a database will probably serialize it. And I'm realizing now that it should probably pass the record by value for this reason, but let's do that separately.

rust/protocol/src/storage/inmem.rs
```diff
@@ -1,38 +1,52 @@
...
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Implementations for stores defined in [super::traits].
+//!
+//! These implementations are purely in-memory, and therefore most likely useful for testing
+//! right now.
+//!
+//! *TODO: Define a serialization scheme for all of the in-memory stores so that they can be
+//! persisted to file or elsewhere!*
```
jrose-signal, line 12: Eh. *Someone* can do this, but I don't know if it will be a goal for this library. It's more likely that the library will start including a SQL-based implementation that's separate from the in-memory stores.

rust/protocol/src/storage/inmem.rs
rubdos: IIRC, it's also what rustfmt would do.

cosmicexplorer: Done!

rust/protocol/src/storage/inmem.rs
cosmicexplorer: Done! Also changed "and with" to "along with".

rust/protocol/src/storage/traits.rs
cosmicexplorer: Moved into a non-doc comment and suggested a move to `utils.rs`!

rust/protocol/src/storage/traits.rs
cosmicexplorer: Done! Thanks :D

rust/protocol/src/storage/traits.rs
cosmicexplorer: Thanks! I decided to just remove "for other users" instead of explicitly saying it also stores our own records, since that seemed redundant.

rust/protocol/src/storage/traits.rs
cosmicexplorer: I've shuffled this sentence to say "Assign `record` to the entry for..."!

rust/protocol/src/storage/inmem.rs
cosmicexplorer: Oops! That was a personal note for my own use case. Removed, but thank you for sketching that possible implementation!

rust/protocol/src/storage/inmem.rs
```diff
@@ -1,36 +1,46 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Implementations for stores defined in [super::traits].
+//!
+//! These implementations are purely in-memory, and therefore most likely useful for testing
+//! right now.
```
jrose-signal, line 9: Nitpick: let's strike the "right now". I don't think we generally want to give the impression that a bunch of features are "just around the corner" or anything.

rust/protocol/src/storage/inmem.rs
```diff
@@ -298,6 +316,8 @@ impl traits::SenderKeyStore for InMemSenderKeyStore {
     }
 }
 
+#[allow(missing_docs)]
+/// Reference implementation of [traits::ProtocolStore].
 #[derive(Clone)]
```
jrose-signal, line 321: Nitpick: attribute/doc ordering again

jrose-signal (approved):
Only a few tiny remaining comments, then I'll cherry-pick to the private branch as before. (They're all small and uncontroversial so I might just take care of them.)

cosmicexplorer:
Going to leave this untouched as I assume the "awaiting-release" label means you have applied these yourself, but I agree with these changes! Thanks!

jrose-signal:
Released in v0.22.0!

PR #469, opened 2022-06-27
add docs to utils and expose `constant_time_cmp()` to consumers

cosmicexplorer:
*Broken out of #287.*

This key comparison method is useful to have for downstream consumers to avoid leaking timing information when comparing bytes. If downstream consumers of the `libsignal` crate wish to implement their own structs wrapping secret bytes, they now have a clearly documented example of how to implement that.

### DIFF
--- a/rust/protocol/src/lib.rs
+++ b/rust/protocol/src/lib.rs
@@ -39,7 +39,7 @@ mod session;
 mod session_cipher;
 mod state;
 mod storage;
-mod utils;
+pub mod utils;
 
 use error::Result;
 
--- a/rust/protocol/src/utils.rs
+++ b/rust/protocol/src/utils.rs
@@ -1,7 +1,9 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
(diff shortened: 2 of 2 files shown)


jrose-signal:
I got curious as to why `subtle` *doesn't* have this, since it seems like they could implement it similarly, and found https://github.com/dalek-cryptography/subtle/issues/61. libsignal-protocol's implementation isn't actually lexicographic, it's [length-lexicographic](https://en.wikipedia.org/wiki/Shortlex_order), and (as noted) not constant-time in that case. I'd say that tips it over the edge to being left as an implementation detail; the place for this is really `subtle` (or `subtle-ng`), and just because they don't have it doesn't mean we should provide our own quirky implementation.

cosmicexplorer:
Ok, that is exactly what I was hoping to learn from filing this (my thought was: if we're doing something very wonky, let's document it, and if we're not, let's use the more canonical thing)! I'll close this PR and see what `subtle`/`subtle-ng` thinks about this use case.

cosmicexplorer:
I am hoping to implement this in zkcrypto/subtle-ng#5!

cosmicexplorer:
Note: `subtle-ng` is unmaintained, `subtle` is the correct project.

PR #470, opened 2022-06-27
add docs to identity_key.rs

cosmicexplorer:
*Broken out of #287.*

### DIFF
--- a/rust/protocol/src/identity_key.rs
+++ b/rust/protocol/src/identity_key.rs
@@ -1,7 +1,11 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
+
+//! Wrappers over cryptographic primitives from [`crate::curve`] to represent a user.
+
+#![warn(missing_docs)]
 
 use crate::proto;
 use crate::{KeyPair, PrivateKey, PublicKey, Result, SignalProtocolError};
@@ -15,31 +19,42 @@ use prost::Message;
 const ALTERNATE_IDENTITY_SIGNATURE_PREFIX_1: &[u8] = &[0xFF; 32];
 const ALTERNATE_IDENTITY_SIGNATURE_PREFIX_2: &[u8] = b"Signal_PNI_Signature";
 
+/// A public key that represents the identity of a user.
+///
+/// Wrapper for [`PublicKey`].
 #[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Copy)]
 pub struct IdentityKey {
     public_key: PublicKey,
 }
 
 impl IdentityKey {
+    /// Initialize a public-facing identity from a public key.
     pub fn new(public_key: PublicKey) -> Self {
         Self { public_key }
     }
 
+    /// Return the public key representing this identity.
     #[inline]
     pub fn public_key(&self) -> &PublicKey {
         &self.public_key
     }
 
+    /// Return an owned byte slice which can be deserialized with [`Self::decode`].
     #[inline]
     pub fn serialize(&self) -> Box<[u8]> {
         self.public_key.serialize()
     }
 
+    /// Deserialize a public identity from a byte slice.
     pub fn decode(value: &[u8]) -> Result<Self> {
         let pk = PublicKey::try_from(value)?;
         Ok(Self { public_key: pk })
     }
 
+    /// Given a trusted identity `self`, verify that `other` represents an alternate identity for
+    /// this user.
+    ///
+    /// `signature` must be calculated from [`IdentityKeyPair::sign_alternate_identity`].
     pub fn verify_alternate_identity(&self, other: &IdentityKey, signature: &[u8]) -> Result<bool> {
         self.public_key.verify_signature_for_multipart_message(
             &[
@@ -72,6 +87,9 @@ impl From<IdentityKey> for PublicKey {
     }
 }
 
+/// The private identity of a user.
+///
+/// Can be converted to and from [`KeyPair`].
 #[derive(Copy, Clone)]
 pub struct IdentityKeyPair {
     identity_key: IdentityKey,
@@ -79,6 +97,7 @@ pub struct IdentityKeyPair {
 }
 
 impl IdentityKeyPair {
+    /// Create a key pair from a public `identity_key` and a private `private_key`.
     pub fn new(identity_key: IdentityKey, private_key: PrivateKey) -> Self {
         Self {
             identity_key,
@@ -86,6 +105,7 @@ impl IdentityKeyPair {
         }
     }
 
+    /// Generate a random new identity from randomness in `csprng`.
     pub fn generate<R: CryptoRng + Rng>(csprng: &mut R) -> Self {
         let keypair = KeyPair::generate(csprng);
 
@@ -95,21 +115,25 @@ impl IdentityKeyPair {
         }
     }
 
+    /// Return the public identity of this user.
     #[inline]
     pub fn identity_key(&self) -> &IdentityKey {
         &self.identity_key
     }
 
+    /// Return the public key that defines this identity.
     #[inline]
     pub fn public_key(&self) -> &PublicKey {
         self.identity_key.public_key()
     }
 
+    /// Return the private key that defines this identity.
     #[inline]
     pub fn private_key(&self) -> &PrivateKey {
         &self.private_key
     }
 
+    /// Return a byte slice which can later be deserialized with [`Self::try_from`].
     pub fn serialize(&self) -> Box<[u8]> {
         let structure = proto::storage::IdentityKeyPairStructure {
             public_key: self.identity_key.serialize().to_vec(),
@@ -120,6 +144,7 @@ impl IdentityKeyPair {
         result.into_boxed_slice()
     }
 
+    /// Generate a signature claiming that `other` represents the same user as `self`.
     pub fn sign_alternate_identity<R: Rng + CryptoRng>(
         &self,
         other: &IdentityKey,


rust/protocol/src/identity_key.rs
```diff
@@ -15,31 +24,39 @@ use prost::Message;
 const ALTERNATE_IDENTITY_SIGNATURE_PREFIX_1: &[u8] = &[0xFF; 32];
 const ALTERNATE_IDENTITY_SIGNATURE_PREFIX_2: &[u8] = b"Signal_PNI_Signature";
 
+/// The public identity of a user, used in [IdentityKeyStore].
+///
+/// Wrapper for [PublicKey].
```
jrose-signal, line 29: The "used in IdentityKeyStore" seems a little too specific to me. Maybe just "A public key that represents the identity of a user"?

rust/protocol/src/identity_key.rs
```diff
@@ -15,31 +24,39 @@ use prost::Message;
...
     public_key: PublicKey,
 }
 
 impl IdentityKey {
+    /// Initialize a public-facing identity from a public key.
     pub fn new(public_key: PublicKey) -> Self {
         Self { public_key }
     }
 
+    /// Return a public key representing the public identity.
```
jrose-signal, line 41: Nitpick: I'd say "the public key" and "this identity". One could imagine a system where the same public identity would have multiple keys, but once you've already gotten to an IdentityKey, there's exactly one PublicKey that goes with it.

rust/protocol/src/identity_key.rs
```diff
@@ -15,31 +24,39 @@ use prost::Message;
...
     }
 
+    /// Deserialize a public identity from a byte slice.
     pub fn decode(value: &[u8]) -> Result<Self> {
         let pk = PublicKey::try_from(value)?;
         Ok(Self { public_key: pk })
     }
 
+    /// Verify `signature` matches a provided identity key `other` (*not* `self`)!
     pub fn verify_alternate_identity(&self, other: &IdentityKey, signature: &[u8]) -> Result<bool> {
```
jrose-signal, line 60: This is admittedly a specialized feature Signal isn't using yet, but I think it looks more like this:

```
Given a trusted identity `self`, verify that `other` represents the same user.

See also [IdentityKeyPair::sign_alternate_identity].
```

rust/protocol/src/identity_key.rs
```diff
@@ -72,20 +89,29 @@ impl From<IdentityKey> for PublicKey {
     }
 }
 
+/// The private identity of a user.
+///
+/// This cryptographic identity is used to sign pre-keys in [PreKeySignalMessage], to conceal
+/// sealed-sender [crate::sealed_sender::UnidentifiedSenderMessage]s, or to initiate the key
+/// derivation function for a [SessionRecord].
```
jrose-signal, line 96: I'm again unhappy with including the specific uses. These are some ways identity key pairs are used *today* in *this* crate, but Signal uses them for lots of other things too at the app level, and we might use them more for other things in the future.

rust/protocol/src/identity_key.rs
```diff
@@ -120,6 +150,7 @@ impl IdentityKeyPair {
         result.into_boxed_slice()
     }
 
+    /// Generate a signature for a provided identity key `other` (*not* `self`)!
```
jrose-signal, line 153: Similarly, perhaps "Generate a signature claiming that `other` represents the same user as `self`."

jrose-signal (commented):
Thank you!

rust/protocol/src/identity_key.rs
cosmicexplorer: Done!

rust/protocol/src/identity_key.rs
cosmicexplorer: Done! Thanks so much for the explanation!

rust/protocol/src/identity_key.rs
cosmicexplorer: Done!

rust/protocol/src/identity_key.rs
cosmicexplorer: Ok, thanks for that feedback! I think my motivation for including the specifics was more just as my own notes on figuring out the codebase, and I'll avoid including these in the future.

rust/protocol/src/identity_key.rs
cosmicexplorer: Done! I also clarified that `signature` in `IdentityKey::verify_alternate_identity` must be calculated from `IdentityKeyPair::sign_alternate_identity`.

rust/protocol/src/identity_key.rs
```diff
@@ -1,8 +1,14 @@
 //
-// Copyright 2020 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+//! Wrappers over cryptographic primitives from [`crate::curve`] to represent [identity].
+//!
+//! [identity]: https://en.wikipedia.org/wiki/Online_identity
```
jrose-signal, line 8: This is real abstract, heh. Maybe we can get away with just saying "represent a user"?

rust/protocol/src/identity_key.rs
```diff
@@ -15,31 +21,42 @@ use prost::Message;
...
     }
 
+    /// Deserialize a public identity from a byte slice.
     pub fn decode(value: &[u8]) -> Result<Self> {
         let pk = PublicKey::try_from(value)?;
         Ok(Self { public_key: pk })
     }
 
+    /// Given a trusted identity `self`, verify that `other` represents the same user for the given
+    /// `signature`.
```
jrose-signal, line 57: I'm not sure what you mean by "for the given signature". The signature is used to perform the verification, but if the verification passes, `self` and `other` represent the same user in all cases, not just regarding this signature.

rust/protocol/src/identity_key.rs
cosmicexplorer: Done!

cosmicexplorer: I had been heavily editing the wikipedia page on "privacy" around this time last year and got some wires crossed, I think.

rust/protocol/src/identity_key.rs
cosmicexplorer: I was trying to clarify why the `signature` is needed for this method, and wasn't quite understanding what "alternate" identity was referring to. I have modified this to:
```rust
/// Given a trusted identity `self`, verify that `other` represents an alternate identity for
/// this user.
///
/// `signature` must be calculated from [`IdentityKeyPair::sign_alternate_identity`].
```

It looks like the alternate identity calculation mechanism is consumed in the app somewhere; I'll take a look at documenting that more fully in another PR perhaps.

jrose-signal: It's part of the Phone Number Privacy project but I don't know if anyone's hooked it up end-to-end on the app side yet. But as usual that's just the current use case; the API's designed for future sorts of "alternate identities" as well.

jrose-signal:
Since the last round of PRs we've changed libsignal to match the other Signal repositories in developing mostly in private and doing public releases, with external contributions being cherry-picked into the upcoming release's private branch. That's not so great for an ongoing patch series like this one, though. Perhaps we can make a side branch your PRs are targeted to, `docs-effort` or something, so that you can build on your own work between releases? While meanwhile they are also cherry-picked to our private `main`.

clauz9:
> Since the last round of PRs we've changed libsignal to match the other Signal repositories in developing mostly in private and doing public releases,

I was just admiring the fact that the development for libsignal is a lot more transparent, and publicly accessible, compared to the clients 🤔 I love the fact that I am able to see developers making PR's for features and reviewing each other's code, it's incredibly educational. 

so, can I ask, why the change ?

jrose-signal:
There wasn't one big reason, but a bunch of small things adding up:

- Being inconsistent with the other Signal projects means that it's harder to write automation generic across all Signal-managed projects.

- Being inconsistent with other Signal projects means that Signal employees have to remember to be more circumspect on libsignal reviews (i.e. not mentioning possible future features or security concerns that could get taken out of context, which has historically been quite frustrating). Our releases are scrutinized quite closely throughout the internet already; having our discussion be public too was 90% okay and 10% walking on eggshells. (I think I got congratulated on joining Signal within a few days because of this GitHub account as well as my https://community.signalusers.org account, even though I hadn't said anything elsewhere.)

    This still comes up sometimes with public forks ([signalapp/boring](/signalapp/boring) being the most active example for libsignal at the moment; will be used soon), but that's usually a lot rarer.

- Even in the original development style, we sometimes wanted *specific* reviews to be discussed privately, for similar reasons. Making the *most sensitive* reviews the ones with the *more awkward process* was optimizing for the wrong thing.

I'm a little sad too, but ultimately while it *was* nice to have a "fully open" development, that's not really the main goal of the project. The thing we optimize for is "supporting the Signal client apps" rather than "making a model open-source library", and I do think that's the right choice.

[This is not something that can be argued with, but if you want to keep discussing it please move that discussion to https://community.signalusers.org, so we can go back to talking about the PR content and how to integrate it!]

clauz9:
> Signal employees have to remember to be more circumspect on libsignal reviews (i.e. not mentioning possible future features or security concerns that could get taken out of context, which has historically been quite frustrating)

yes, this makes perfect sense..

> This is not something that can be argued with, but if you want to keep discussing it please move that discussion to https://community.signalusers.org, 

No need, your answer was quite detailed, and makes perfect sense, thanks for taking the time to answer!

cosmicexplorer:
I have updated the description of the tracking issue #487 to note that I am maintaining a branch named `docs-effort` at https://github.com/cosmicexplorer/libsignal-client/tree/docs-effort with all these changes together! @jrose-signal I appreciate your careful description of the new review process and appreciate all the effort Signal puts into ensuring the app remains secure!

PR #471, opened 2022-06-27
add documentation for error cases

cosmicexplorer:
*Broken out of #287.*

While we have a terse description of each case of `libsignal_protocol::Error` to implement `displaydoc::Display`, we may also wish to have more in-depth explanation for the purposes of generated documentation. The `#[ignore_extra_doc_attributes]` flag enables us to provide this additional documentation while retaining the first line for the `displaydoc::Display` impl.

### DIFF
--- a/rust/protocol/src/error.rs
+++ b/rust/protocol/src/error.rs
@@ -1,7 +1,11 @@
 //
-// Copyright 2020-2021 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
+
+//! Errors that may occur during various stages of the Signal Protocol.
+
+#![warn(missing_docs)]
 
 use crate::curve::KeyType;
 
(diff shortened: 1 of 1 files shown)


jrose-signal:
(still on my todo list, but prose is harder to review than code!)

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
 
 use std::panic::UnwindSafe;
 
+#[cfg(doc)]
+pub use crate::{
```
jrose-signal, line 19: Nitpick: does it have to be `pub` for rustdoc to pick it up?

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
+    storage::{IdentityKeyStore, PreKeyStore, SenderKeyStore, SessionStore, SignedPreKeyStore},
+};
+
+/// Return type for all fallible operations in this crate.
 pub type Result<T> = std::result::Result<T, SignalProtocolError>;
 
+/// Error states recognized by the Signal Protocol.
 #[derive(Debug, Display, Error)]
+#[ignore_extra_doc_attributes]
 pub enum SignalProtocolError {
```
jrose-signal, line 37: Heh, this sounds better designed than what's actually going on: the "SignalProtocol" here is really just about the name of the crate, libsignal-protocol. Arguably it could just be Error or something like that; I'd probably document it as "Possible errors produced by the operations in this crate" to match what you did for the Result alias.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// invalid argument: {0}
+    ///
+    /// Raised if an invalid argument is provided to any Signal API methods.
+    ///
+    /// Prefer to use lifetimes, static-sized slices, and dedicated wrapper structs in API
+    /// signatures to minimize the need to raise this error to FFI boundaries.
```
jrose-signal, line 43: Documentation should be client-facing, so comments about API design don't really fit.

jrose-signal (commented):
Ugh, reviewing this is reminding me how much I disagree with my past self and collaborator about the design of our errors, heh. (We were both much newer to Rust then.) Several of these error cases don't need to be distinguished because clients won't treat them differently (as you yourself point out in some of these). Some should probably be panics per Rust conventions. Some don't have *enough* information (which key was the bad key??).

But redoing libsignal's API to be ever-more-perfect is not the best use of my time as a Signal employee, and therefore it'd be better to have documentation than not. I know I've delayed on reviewing this for quite a while, and that'll continue a little longer, but I'll pick it up again when I return from vacation.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
+    /// Prefer to raise [Self::InvalidState] except in methods which directly decode protobufs.
     InvalidProtobufEncoding,
 
     /// ciphertext serialized bytes were too short <{0}>
+    ///
+    /// Raised if some ciphertext was deserialized from a too-small slice.
+    ///
+    /// Prefer to make API method signatures and wrapper structs consume and produce static-sized
+    /// byte slices to minimize the need to raise this error.
     CiphertextMessageTooShort(usize),
```
jrose-signal, line 67: This should be unified with `InvalidEncoding` above; there's nothing a caller would do differently based on which failure happened. If we think the different kinds of failure are important for debugging, we could add a message to `InvalidEncoding`. Failing that, there shouldn't be any mention of "slice"; this just indicates a parameter was too small.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// ciphertext version was too old <{0}>
+    ///
+    /// Raised if the ciphertext version decoded from a protobuf is older than this client.
+    ///
+    /// The current client's ciphertext version is at [CIPHERTEXT_MESSAGE_CURRENT_VERSION].
     LegacyCiphertextVersion(u8),
```
jrose-signal, line 73: This shouldn't mention "protobuf", and the particular current version depends on which protocol is raising this error. Signal, SenderKey, and SealedSender messages all have different versions.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// ciphertext version was unrecognized <{0}>
+    ///
+    /// Raised if the ciphertext version decoded from a protobuf is newer than this client.
+    ///
+    /// The current client's ciphertext version is at [CIPHERTEXT_MESSAGE_CURRENT_VERSION].
     UnrecognizedCiphertextVersion(u8),
```
jrose-signal, line 79: Ditto.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// unrecognized message version <{0}>
+    ///
+    /// Raised if the ciphertext version decoded from a protobuf fails to match the cached version
+    /// for the message chain that message originates from.
+    ///
+    /// *TODO: This case should wrap the same numeric type as [Self::LegacyCiphertextVersion] and
+    /// [Self::UnrecognizedCiphertextVersion]. This dissonance is addressed in
+    /// <https://github.com/signalapp/libsignal-client/pull/289>.*
     UnrecognizedMessageVersion(u32),
```
jrose-signal, line 88: Protobuf is an implementation detail. The version is associated with a session, not a particular chain.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
 
     /// fingerprint version number mismatch them {0} us {1}
+    ///
+    /// Raised if a fingerprint version decoded from a protobuf has an unexpected value.
     FingerprintVersionMismatch(u32, u32),
```
jrose-signal, line 93: Protobuf is an implementation detail, and the value isn't necessarily "unexpected"; it's just not the same as what the local fingerprint is using.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// fingerprint parsing error
+    ///
+    /// Raised if a field in a fingerprint protobuf is invalid in some way.
+    ///
+    /// Similar to [Self::InvalidProtobufEncoding].
     FingerprintParsingError,
```
jrose-signal, line 99: The long-term thing to do here is give fingerprints their own error enums, but short-term agreed, this is basically the same as InvalidEncoding.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
 
     /// no key type identifier
+    ///
+    /// Raised if a [PublicKey] is deserialized from an empty slice.
+    ///
+    /// Prefer to use static-sized slices in API method signatures and struct fields to minimize the
+    /// need to raise this error.
     NoKeyTypeIdentifier,
```
jrose-signal, line 107: Yeah, this is a little silly. No one should mistake an empty slice for a public key.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
+    ///
+    /// Raised if a [KeyType] decoded from a [u8] has an unrecognized value.
     BadKeyType(u8),
     /// bad key length <{1}> for key with type <{0}>
+    ///
+    /// Raised if a [PublicKey] or [PrivateKey] is deserialized from a slice of incorrect length.
+    ///
+    /// Prefer to use static-sized slices in API method signatures and struct fields to minimize the
+    /// need to raise this error.
     BadKeyLength(KeyType, usize),
```
jrose-signal, line 118: ✅

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
 
     /// invalid signature detected
+    ///
+    /// Raised if signature validation fails for a [SignedPreKeyRecord] or a [SenderKeyMessage].
     SignatureValidationFailed,
```
jrose-signal, line 123: Nitpick: it would be better to make it clear that these are examples; many things have signatures that could fail to validate. These are just the two we surface as "signature validation failed".

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
 
     /// untrusted identity for address {0}
+    ///
+    /// Raised if an identity verification check fails in [message_encrypt] or [message_decrypt].
     UntrustedIdentity(crate::ProtocolAddress),
```
jrose-signal, line 128: Again, feels weirdly specific. Anything that takes an IdentityKeyStore could check identities and decide that one is untrusted.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
+    NoSenderKeyState {
+        /// Unique identifier for the sender key session.
+        distribution_id: Uuid,
+    },
 
     /// session with {0} not found
+    ///
+    /// Raised if an [IdentityKeyStore] does not contain an entry for a [ProtocolAddress], or
+    /// alternately if a [SessionStore] does not contain a session for a given [ProtocolAddress].
     SessionNotFound(crate::ProtocolAddress),
```
jrose-signal, line 161: I'd switch the order of these; the *common* use is a missing session (hence the name), but it also gets used for missing identity when one is expected because repairing the session will fetch the identity too. cf. comment in the middle of `sealed_sender_multi_recipient_encrypt`:

```
// Returned as a SessionNotFound error because (a) we don't have an identity error
// that includes the address, and (b) re-establishing the session should re-fetch
// the identity.
```

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
+    ///
+    /// Raised if a [SignalMessage] could not be decrypted or some field had an unexpected value.
+    ///
+    /// *TODO: what differentiates this from [Self::CiphertextMessageTooShort]?*
     InvalidMessage(crate::CiphertextMessageType, &'static str),
 
     /// error while invoking an ffi callback: {0}
+    ///
+    /// Raised to propagate an error from an FFI callback.
     FfiBindingError(String),
```
jrose-signal, line 197: This is not about propagating errors; it's when the callback itself can't be invoked. Think JNI-level errors. (It does not actually come up for pure Rust users and maybe doesn't belong here.)

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// error in method call '{0}': {1}
+    ///
+    /// Raised to propagate an error through to an FFI exception along with a boxed handle.
     ApplicationCallbackError(
         &'static str,
         #[source] Box<dyn std::error::Error + Send + Sync + UnwindSafe + 'static>,
     ),
```
jrose-signal, line 204: This is the one where there was an error from a callback and we're just propagating it out, whether through FFI or from a Rust client that returns Err.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
+    /// Raised if an [crate::sealed_sender::UnidentifiedSenderMessage] could not be
+    /// deserialized successfully.
+    ///
+    /// *TODO: this sounds a lot like [Self::InvalidProtobufEncoding] or [Self::UntrustedIdentity]?*
     InvalidSealedSenderMessage(String),
     /// unknown sealed sender message version {0}
+    ///
+    /// Raised if an version decoded from a [crate::sealed_sender::UnidentifiedSenderMessage]
+    /// was unrecognized.
     UnknownSealedSenderVersion(u8),
```
jrose-signal, line 217: This won't be decoded from UnidentifiedSenderMessage; it's in the outermost envelope.

rust/protocol/src/error.rs
```diff
@@ -11,82 +15,208 @@ use uuid::Uuid;
...
     /// self send of a sealed sender message
+    ///
+    /// Raised if [sealed_sender_decrypt] finds that the message came from this exact client.
     SealedSenderSelfSend,
```
jrose-signal, line 221: ✅

jrose-signal (commented):
Reading this is painful, it reminds me that the error design really isn't that great. As I already remarked above. 😅 But I've owed you this review for a long time, even if I scanned through them pretty quickly.

PR #473, opened 2022-06-27
document a few constants

cosmicexplorer:
*Broken out from #287.*

1. Add documentation to several constants.
1. Put those constants in a `limits` submodule (this part can be reverted if desired).


cosmicexplorer:
I'm aware that constants like these are supposed to be implementation details, but I like the idea of bringing these out into the open because some of them have implications on security. A future change might be to make these constants into default values and allow consumers to override those in these methods in `{group_cipher,session_cipher,session,sender_keys}.rs`.

rust/protocol/src/consts.rs
```diff
@@ -1,10 +1,34 @@
...
-pub const MAX_MESSAGE_KEYS: usize = 2000;
-pub const MAX_RECEIVER_CHAINS: usize = 5;
-pub const ARCHIVED_STATES_MAX_LENGTH: usize = 40;
-pub const MAX_SENDER_KEY_STATES: usize = 5;
+#![warn(missing_docs)]
+
+//! Magic numbers.
+
+/// Various positive integers bounding the maximum size of other data structures.
+pub mod limits {
```
jrose-signal, line 11: If they're all limits then maybe `consts` should just change to `limits`.

rust/protocol/src/consts.rs
```diff
@@ -1,10 +1,34 @@
...
+    /// The maximum number of encrypted messages that the client chain which decrypts Signal
+    /// messages in a [Double Ratchet] instance can retrieve at once (tracked in
+    /// [crate::proto::storage::session_structure::chain::ChainKey::index] as well as a separate
+    /// `counter`).
```
jrose-signal, line 15: This isn't really about how many messages you can decrypt at once; it's about how many messages you can *not* receive and still decrypt further messages on an existing chain. Hard to explain, though!

rust/protocol/src/consts.rs
```diff
@@ -1,10 +1,34 @@
...
+    ///
+    /// [Double Ratchet]: https://signal.org/docs/specifications/doubleratchet/
+    pub const MAX_FORWARD_JUMPS: usize = 25_000;
+    /// The maximum number of per-message keys that can be retained to decrypt messages within
+    /// a specific chain from `message_keys` in [crate::proto::storage::session_structure::Chain].
+    pub const MAX_MESSAGE_KEYS: usize = 2000;
```
jrose-signal, line 21: Similarly, this one is about when you *do* encounter such a jump, how many past keys will the session retain in case you receive some of those missing messages (due to some kind of out-of-order delivery, I guess).

rust/protocol/src/consts.rs
```diff
@@ -1,10 +1,34 @@
...
+    /// The maximum number of temporary backup chains to allow for `receiver_chains` in
+    /// [crate::proto::storage::SessionStructure]. These backup chains corresponds to the [Sesame]
+    /// protocol for syncing a Double Ratchet chain between two users.
+    ///
+    /// [Sesame]: https://signal.org/docs/specifications/sesame/#server
+    pub const MAX_RECEIVER_CHAINS: usize = 5;
```
jrose-signal, line 27: This is just about past chains within the Double Ratchet protocol, again to handle out-of-order delivery.

rust/protocol/src/consts.rs
cosmicexplorer: I have spent some time to distill what it actually does, which ended up taking a couple paragraphs. Please let me know if I should move that documentation elsewhere.

rust/protocol/src/consts.rs
cosmicexplorer: I have spent a little time to rewrite this, referring to the documentation for `MAX_FORWARD_JUMPS` I rewrote above.

rust/protocol/src/consts.rs
cosmicexplorer: I rewrote this to clarify that this only occurs for double ratchet chains.

rust/protocol/src/consts.rs
cosmicexplorer: Great idea! Done!

cosmicexplorer:
So I spent some time to understand what each of these limits actually represents, and I think the information is useful, in particular the description of what each limit actually bounds and why that matters (avoiding denial-of-service attacks, for one). I'm going to spend some more time understanding what an archived session state is.

While these descriptions are quite useful for me to understand what the library is doing (as implementation notes for some parts which aren't fully defined in the Double Ratchet spec), they may be too verbose for this file, so one possibility might be to move this more involved documentation to `{session,group}_cipher.rs` and just leave the pared-down explanation of what each limit actually bounds in these module docs.

If that arrangement seems useful, then we might also consider making `limits` a `pub mod` in order to expose these docs? Not sure right now. Going to figure out how session archiving works first before thinking about where to put these docs.

rust/protocol/src/consts.rs
cosmicexplorer: (this is now in `limits.rs`)

rust/protocol/src/consts.rs
cosmicexplorer: (this is now in `limits.rs`)

rust/protocol/src/consts.rs
cosmicexplorer: (this is now in `limits.rs`)

PR #474, opened 2022-06-27
remove unnecessary Result-wrapping in `crypto.rs`

cosmicexplorer:
*Broken out from #287.*

1. Separate CTR and CBC encryption methods in `ctr` and `cbc` submodules.
1. Define `pub const` sizes for array lengths.
1. Require fixed-size arrays to en/decryption methods to remove  `EncryptionError` and one case of `DecryptionError`.
1. Rename methods to `aes_256_*` and `aes_256_ctr_hmac_*` for consistency.

### TODO
- [ ] Figure out how to apply this comment: https://github.com/signalapp/libsignal/pull/287#discussion_r647031224.


jrose-signal:
These aren't `pub` for a reason: if someone wants to use these operations, they should get them directly from the RustCrypto crates. These are pretty much just convenience wrappers because of how they're used within the protocol. (The hardcoded "first ten bytes as a MAC" is something that a generalized library probably wouldn't hardcode.)

I think you're right about EncryptionError, though. *Someone* has to check that these keys are the right length, but it can happen a level up. And then I think DecryptionError collapses down to a struct instead of a one-case enum (not that it matters so much for an implementation detail).

cosmicexplorer:
> These are pretty much just convenience wrappers because of how they're used within the protocol. 

Ok, so in light of this I've changed it to `pub(crate) mod crypto;` in `lib.rs`, removed the doctests, and re-focused this PR on just removing unnecessary `Result`s by enforcing fixed-size slices where possible.

However, on top of that, I've also kept in:
- the `cbc` and `ctr` module organization
- some docstrings for parts that I thought weren't immediately clear
- new en/decryption test cases using `Rng::gen()`

I can revert any of that as desired.

rust/protocol/src/crypto.rs
```diff
@@ -1,156 +1,244 @@
...
-    BadKeyOrIv,
-}
+/// bad ciphertext: {0}
+///
+/// Either the input is malformed, or the MAC doesn't match on decryption.
+///
+/// These cases should not be distinguished; message corruption can cause either problem.
+#[derive(Debug, Display, Error)]
+#[ignore_extra_doc_attributes]
+pub struct DecryptionError(pub &'static str);
```
jrose-signal, line 26: Nitpick: For some reason displaydoc on structs weirds me out; maybe it's because enum cases often get abbreviated docs anyway in a way that types don't. Mind switching to thiserror-style for the Display representation instead?

rust/protocol/src/crypto.rs
```diff
@@ -1,156 +1,244 @@
...
-pub(crate) enum DecryptionError {
-    /// The key or IV is the wrong length.
-    BadKeyOrIv,
-    /// Either the input is malformed, or the MAC doesn't match on decryption.
-    ///
-    /// These cases should not be distinguished; message corruption can cause either problem.
-    BadCiphertext(&'static str),
-}
+// TODO: Could be nice to have a type-safe library for manipulating units of bytes safely.
+const BITS_PER_BYTE: usize = std::mem::size_of::<u8>() * 8;
```
jrose-signal, line 29: Rust guarantees that `size_of::<u8>()` is 1, so I think it's fine to just write 8 here. (The named constant still makes sense though.)

rust/protocol/src/crypto.rs
```diff
@@ -1,156 +1,244 @@
...
-    let mut cipher = Aes256Ctr::new(key[..].into(), zero_nonce[..].into());
+/// The size of the generated nonce we use for AES encryption in this crate.
+pub const AES_NONCE_SIZE: usize = 128 / BITS_PER_BYTE;
 
-    let mut ctext = ptext.to_vec();
-    cipher.apply_keystream(&mut ctext);
-    Ok(ctext)
-}
+/// Use AES-256 in CTR mode.
+pub mod ctr {
```
jrose-signal, line 38: I don't know if I really see the point of this. Yes, there's a little more organization, but "CTR" and "CBR" don't imply "AES-256" by themselves, and they're all just convenience wrappers for the session implementation in any case.

rust/protocol/src/crypto.rs
```diff
@@ -1,156 +1,244 @@
...
     }
 
-    let mode =
-        Cbc::<Aes256, Pkcs7>::new_from_slices(key, iv).map_err(|_| DecryptionError::BadKeyOrIv)?;
-    mode.decrypt_vec(ctext)
-        .map_err(|_| DecryptionError::BadCiphertext("failed to decrypt"))
-}
+    /// Length in bytes of the [`Hmac`] key used for [`aes_256_ctr_hmac_encrypt`] and
+    /// [`aes_256_ctr_hmac_decrypt`].
+    pub const MAC_KEY_LENGTH: usize = 80 / BITS_PER_BYTE;
```
jrose-signal, line 56: The MAC key isn't 80 bits; only the MAC itself is. The key can in theory be any length and in practice we use a 32-byte key.

rust/protocol/src/crypto.rs
```diff
@@ -1,156 +1,244 @@
...
+        use rand::rngs::OsRng;
+        use rand::Rng;
+
+        #[test]
+        fn static_test() {
+            let key: [u8; AES_256_KEY_SIZE] =
+                hex::decode("603DEB1015CA71BE2B73AEF0857D77811F352C073B6108D72D9810A30914DFF4")
+                    .expect("valid hex")
+                    .try_into()
+                    .expect("correct array size");
```
jrose-signal, line 112: The `hex_literal` crate used elsewhere in the workspace can check these conditions at compile time; if you're adding more tests it might be a good time to switch over (incrementally).

rust/protocol/src/lib.rs
```diff
@@ -24,7 +24,7 @@
 
 mod address;
 mod consts;
-mod crypto;
+pub(crate) mod crypto;
```
jrose-signal, line 27: This doesn't need to be changed, right?

rust/protocol/src/sealed_sender.rs
```diff
@@ -707,12 +707,11 @@ mod sealed_sender_v1 {
...
 
         // Encrypt the sender's public key with AES-256 CTR and a MAC.
-        let sender_static_key_ctext = crypto::aes256_ctr_hmacsha256_encrypt(
+        let sender_static_key_ctext = crypto::ctr::aes_256_ctr_hmac_encrypt(
             &sender_identity.public_key().serialize(),
             &sender_eph_keys.cipher_key,
-            &sender_eph_keys.mac_key,
-        )
-        .expect("just generated these keys, they should be correct");
+            array_ref![&sender_eph_keys.mac_key, 0, crypto::ctr::MAC_KEY_LENGTH],
```
jrose-signal, line 713: …and this is where the MAC key is incorrectly truncated, and our lack of "decrypting this existing message with this session state" tests has come back to bite us.

rust/protocol/src/session_cipher.rs
```diff
@@ -595,23 +574,13 @@ fn decrypt_message_with_state<R: Rng + CryptoRng>(
...
-        Err(crypto::DecryptionError::BadKeyOrIv) => {
-            log::warn!(
-                "{} session state corrupt for {}",
-                current_or_previous,
-                remote_address,
-            );
-            return Err(SignalProtocolError::InvalidSessionStructure(
-                "invalid receiver chain message keys",
-            ));
-        }
```
jrose-signal, line 613: This is the change that convinced me this is a good idea: at first I thought "what happens if those fields are corrupt"? But they can't be, because we already check that when getting the message keys out of the session state, and that wasn't in the type system anywhere.

rust/protocol/src/group_cipher.rs
```diff
@@ -169,22 +176,25 @@ pub async fn group_decrypt(
...
-        sender_key.iv(),
-    ) {
-        Ok(plaintext) => plaintext,
-        Err(crypto::DecryptionError::BadKeyOrIv) => {
+    let cipher_key: [u8; crypto::AES_256_KEY_SIZE] =
+        sender_key.cipher_key().try_into().map_err(|_| {
             log::error!(
-                "incoming sender key state corrupt for {}, distribution ID {}, chain ID {}",
-                sender,
+                "outgoing sender key state corrupt for distribution ID {}",
```
jrose-signal, line 182: Copy/paste error: "incoming" got switched for "outgoing" here and below.

rust/protocol/src/group_cipher.rs
```diff
@@ -40,15 +40,22 @@ pub async fn group_encrypt<R: Rng + CryptoRng>(
...
+            SignalProtocolError::InvalidSenderKeySession { distribution_id }
+        })?;
+    let iv: [u8; crypto::AES_NONCE_SIZE] = message_keys.iv().try_into().map_err(|_| {
+        log::error!(
+            "outgoing sender key state corrupt for distribution ID {}",
+            distribution_id,
+        );
+        SignalProtocolError::InvalidSenderKeySession { distribution_id }
+    })?;
+    let ciphertext = crypto::cbc::aes_256_cbc_encrypt(plaintext, &cipher_key, &iv);
```
jrose-signal, line 58: This duplication is a little annoying, but I can't see a great way to clean it up. Do you think it's worth putting that error generation into a local helper?

PR #475, opened 2022-06-27
rename "djb" type curve to "curve25519" everywhere

cosmicexplorer:
*Broken out from #287.*

1. Rename `KeyType::Djb` to `KeyType::Curve25519`.
1. Rename "DJB_TYPE" to "CURVE_25519_TYPE" in java and swift
1. Add docs and the `Keyed` trait to `curve*.rs`

### TODO

- [ ] See this comment: https://github.com/signalapp/libsignal/pull/287#discussion_r633953757.

### DIFF
--- a/rust/protocol/src/curve.rs
+++ b/rust/protocol/src/curve.rs
@@ -1,56 +1,97 @@
 //
-// Copyright 2020-2021 Signal Messenger, LLC.
+// Copyright 2020-2022 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-pub(crate) mod curve25519;
+//! Cryptographic primitives for asymmetric keys.
+//!
+//! Some example operations:
+//!```
+//! use libsignal_protocol::KeyPair;
+//! use std::collections::HashSet;
+//!
+//! let alice = KeyPair::generate(&mut rand::thread_rng());
+//! assert!(alice == alice.clone());
+//! let bob = KeyPair::generate(&mut rand::thread_rng());
+//! assert!(alice != bob);
+//!
+//! // Keys can be hashed and put in sets.
+//! let key_set: HashSet<KeyPair> = [alice, bob].iter().cloned().collect();
+//! assert!(key_set.contains(&alice));
+//! assert!(key_set.contains(&bob));
+//!```
 
+#![warn(missing_docs)]
+
+pub mod curve25519;
+
+use crate::utils::constant_time_cmp;
 use crate::{Result, SignalProtocolError};
 
 use std::cmp::Ordering;
-use std::convert::TryFrom;
-use std::fmt;
+use std::convert::{TryFrom, TryInto};
+use std::{fmt, hash};
 
 use arrayref::array_ref;
+use displaydoc::Display;
+use num_enum::{IntoPrimitive, TryFromPrimitive, TryFromPrimitiveError};
 use rand::{CryptoRng, Rng};
 use subtle::ConstantTimeEq;
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+/// Map the variant of key being used to and from a [`u8`].
+///
+///```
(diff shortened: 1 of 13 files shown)


rust/protocol/src/curve/curve25519.rs
```diff
@@ -12,17 +23,34 @@ use sha2::{Digest, Sha512};
...
 }
 
+impl fmt::Debug for PrivateKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "PrivateKey {{ secret: {:?} }}", self.secret.to_bytes())
+    }
+}
+
 impl PrivateKey {
+    /// Generate a [`StaticSecret`] and then a [`PublicKey`] from a [`CryptoRng`].
```
jrose-signal, line 53: This is phrased in terms of the implementation. "Generate a new private key from a CryptoRng" is fine.

rust/protocol/src/curve/curve25519.rs
```diff
@@ -100,6 +131,8 @@ impl PrivateKey {
         result
     }
 
+    /// Verify a signature from [`Self::calculate_signature`] against another key pair's public key.
```
jrose-signal, line 134: This is a static method, so there's no "another key pair" to worry about.

rust/protocol/src/curve/curve25519.rs
```diff
@@ -140,10 +173,13 @@ impl PrivateKey {
         bool::from(cap_r_check.as_bytes().ct_eq(&cap_r))
     }
 
+    /// Return the bytes of a [`PublicKey`] that another party can use to validate against
+    /// [`Self::verify_signature`].
     pub fn derive_public_key_bytes(&self) -> [u8; PUBLIC_KEY_LENGTH] {
```
jrose-signal, line 178: That's not the only use, it's also for key agreement! I'd drop the use part completely again.

rust/protocol/src/lib.rs
```diff
@@ -25,7 +25,7 @@
 mod address;
 mod consts;
 mod crypto;
-mod curve;
+pub mod curve;
```
jrose-signal, line 28: This still counts as changing the module structure; it's the public surface of the module that matters, not its implementation. I still think we should stay flat until we discuss it properly and as a whole.

rust/protocol/src/curve.rs
```diff
@@ -1,56 +1,98 @@
...
-    type Error = SignalProtocolError;
-
-    fn try_from(x: u8) -> Result<Self> {
-        match x {
-            0x05u8 => Ok(KeyType::Djb),
-            t => Err(SignalProtocolError::BadKeyType(t)),
-        }
-    }
+/// Interface for structs that perform operations parameterized by values of [`KeyType`].
+pub trait Keyed {
```
jrose-signal, line 81: Hm, this isn't really the right name for the trait. Public keys and private keys and key pairs aren't "keyed", they're…"typed"?

That said, though, I think this is a reasonable trait, but until we have a use case for it I'd rather not add it, and that won't happen until we support another key type. It might turn out we need different operations besides just `key_type`.

rust/protocol/src/curve.rs
```diff
@@ -260,27 +336,48 @@ impl PrivateKey {
...
             }
         }
     }
 
-    pub fn calculate_agreement(&self, their_key: &PublicKey) -> Result<Box<[u8]>> {
+    /// Calculate a shared secret between this private key and the public key `their_key`.
+    pub fn calculate_agreement(
+        &self,
+        their_key: &PublicKey,
+    ) -> Result<[u8; curve25519::AGREEMENT_LENGTH]> {
```
jrose-signal, line 357: This is the place we talked about before: key agreement can't hardcode an output length, because the keys might (someday) not be curve25519 keys.

jrose-signal (commented):
I forgot to check which PRs were still drafts, whoops. Many comments on this one, and I feel a little bad for being like this since this is how we go months without any documentation progress.

PR #478, opened 2022-06-30
demonstrate using subtle-ng-derive with ConstantTimeCmp

cosmicexplorer:
### Problem
In #469 we decided to take a look at whether we could upstream some extensions to the `subtle` crate to support `Ordering` comparisons. After creating zkcrypto/subtle-ng#5, I created this PR to demonstrate how adopting `subtle::ConstantTimeCmp` from that change could improve the code.

### Solution
- I created the [`subtle-ng-derive`](https://crates.io/crates/subtle-ng-derive) crate in order to derive constant-time comparisons for aggregate structs using the technique developed in zkcrypto/subtle-ng#6.
- I pulled in a branch of `subtle-ng` which contained the changes from zkcrypto/subtle-ng#5.

### Result
It is now possible to `#[derive(ConstEq, ConstOrd, ConstantTimeEq, ConstantTimeCmp)]` to get constant-time `Eq` and `Ord` implementations for several structs, which allows us to remove `utils.rs`.


cosmicexplorer:
Note: `subtle-ng` is unmaintained, and the original `subtle` crate should be used. I have raised this use case with upstream `subtle` project at https://github.com/dalek-cryptography/subtle/issues/100 and hope to integrate the derive macros into the upstream `subtle` crate itself.

Issue #484, opened 2022-08-22
Can't load libsignal_jni in windows env


ClementBareth:
java.lang.UnsatisfiedLinkError: ...\AppData\Local\Temp\resource16312657568160574112.so: Can't load this .dll (machine code=0x6666) on a AMD 64-bit platform

Looking at the code, it looks like the extension for windows (.dll) is not correct

`  private static void loadLibrary() {
    try {
      String  osName    = System.getProperty("os.name").toLowerCase(java.util.Locale.ROOT);
      boolean isMacOs   = osName.startsWith("mac os");
      String  extension = isMacOs ? ".dylib" : ".so";

      try (InputStream in = Native.class.getResourceAsStream("/libsignal_jni" + extension)) {
        if (in != null) {
          copyToTempFileAndLoad(in, extension);
        } else {
          System.loadLibrary("signal_jni");
        }
      }
    } catch (Exception e) {
      throw new RuntimeException(e);
    }
  }`

PR #485, opened 2022-08-22
Can't load libsignal_jni in windows env #484

ClementBareth:
#484

### DIFF
--- a/java/shared/java/org/signal/libsignal/internal/Native.java
+++ b/java/shared/java/org/signal/libsignal/internal/Native.java
@@ -49,8 +49,14 @@ public final class Native {
   private static void loadLibrary() {
     try {
       String  osName    = System.getProperty("os.name").toLowerCase(java.util.Locale.ROOT);
-      boolean isMacOs   = osName.startsWith("mac os");
-      String  extension = isMacOs ? ".dylib" : ".so";
+      String extension;
+      if (osName.startsWith("mac os")) {
+        extension = ".dylib";
+      } else if(osName.contains("win")) {
+        extension = ".dll";
+      } else {
+        extension = ".so";
+      }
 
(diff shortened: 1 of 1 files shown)


java/shared/java/org/signal/libsignal/internal/Native.java
```diff
@@ -49,8 +49,14 @@ private static void copyToTempFileAndLoad(InputStream in, String extension) thro
   private static void loadLibrary() {
     try {
       String  osName    = System.getProperty("os.name").toLowerCase(java.util.Locale.ROOT);
-      boolean isMacOs   = osName.startsWith("mac os");
-      String  extension = isMacOs ? ".dylib" : ".so";
+      String extension;
+      if (osName.startsWith("mac os")) {
+        extension = ".dylib";
+      } else if(osName.contains("win")) {
```
jrose-signal, line 55: Does `startsWith` not work? Seems good to keep the check narrow.

jrose-signal (commented):
Thank you for picking this up! This has been broken since, uh, forever, but since libsignal_jni is primarily used for the Signal server and Android app, and none of the server and Android developers are doing their day-to-day work on Windows, we've been letting it slip by.

It's a *little* trickier than what you have here, because the canonical Windows library name we've been shipping is "signal_jni.dll", not "libsignal_jni.dll". But that's just making the check cover the whole filename rather than just the extension.

java/shared/java/org/signal/libsignal/internal/Native.java
ClementBareth: I think the startsWIth should work indeed

jrose-signal:
(closed due to age and no pressing need from within Signal, anyone should feel free to pick it up and finish it off, along with testing on a Windows machine)

PR #486, opened 2022-08-31
fixed a typo in an email address

alex:


jrose-signal:
Thank you for catching this! I think this is a good reminder that we should just have a group author listed for all of these. I'll take care of that for the next tag.

PR #492, opened 2022-10-16
Change getSender method of SenderCertificate to always return the UUID

AsamK:
Currently the only user of this method is the [ProtcolException constructor](https://github.com/signalapp/libsignal/blob/b5d48df116c6f0dfb1fb8ab2365a0be63c44be3f/java/client/src/main/java/org/signal/libsignal/metadata/ProtocolException.java#L23), when an UnidentifiedSenderMessageContent is present. All other instances of ProtocolException already use the sender's UUID as sender. So it would be good to have this consistent.

Also brings this in line with similar methods, like [getSourceIdentifier](https://github.com/signalapp/Signal-Android/blob/0cae15b7fdf1b1baa77a81322dd62d93d575d567/libsignal/service/src/main/java/org/whispersystems/signalservice/api/messages/SignalServiceEnvelope.java#L147) on SignalServiceEnvelope in the Android app.

### DIFF
--- a/java/client/src/main/java/org/signal/libsignal/metadata/certificate/SenderCertificate.java
+++ b/java/client/src/main/java/org/signal/libsignal/metadata/certificate/SenderCertificate.java
@@ -64,7 +64,7 @@ public class SenderCertificate implements NativeHandleGuard.Owner {
   }
 
   public String getSender() {
-    return getSenderE164().orElseGet(this::getSenderUuid);
+    return this.getSenderUuid();
   }
 
   public long getExpiration() {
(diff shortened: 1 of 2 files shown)


jrose-signal (approved):
I checked with the Android team and they concur: while the current behavior made sense during the migration to UUIDs, when local recipient info was potentially in terms of just e164s (or preferring e164s), it's no longer our source of truth, and we don't need to keep this behavior from libsignal-metadata-java moving forward.

I'll cherry-pick this into our private branch and it will go out with the next release.

PR #494, opened 2022-11-08
Add branch for CiphertextMessageType::SenderKey in sealed_sender_decrypt

gferon:
While starting to use sender keys, I realized that the helper function `sealed_sender_decrypt` doesn't have a branch for `CiphertextMessageType::SenderKey` which pushes the responsibility of the (correct) implementation to the caller. This PR fixes this issue.

I looked at the source of official Signal clients, and it looks like a few of them call `sealed_sender_decrypt` and then handle these messages themselves on error.

There might be a reason why this is designed like this (maybe avoiding breaking the API?) and I'd be glad to hear about it if you don't want to go forward with this change!


rubdos:
> but the other thing that's held this back is error reporting: if you manage to decrypt the sealed sender wrapper but fail to decrypt the inner message, you probably want to know who _claimed_ to send the inner message (though at that point you probably shouldn't 100% trust that information, since things are already weird).

What if you model this as a happy path instead of the failure path? The `SealedSenderDecryptionResult` could become an `enum SealedSenderDecryptionResult`, with a variant for decryption failure (for which you can handle the claim) and a variant for successful decryption? Or a `Result<Result<, _>, _>`. Just some thoughts about a potential API, if we're to break it.

PR #503, opened 2023-01-04
lean on stable Ordering repr(i8)

cosmicexplorer:
- *Context:* In https://github.com/dalek-cryptography/subtle/pull/98#issuecomment-1371252490 I make the case to the `subtle` crate that `libsignal` can remove our hand-rolled `constant_time_cmp()` if they support `ConstantTimeOrd` upstream via my change.
- While implementing that for the `subtle` project, I learned that you can lean on [`Ordering`](https://doc.rust-lang.org/core/cmp/enum.Ordering.html) having `repr(i8)` to avoid uses of `match`. In dalek-cryptography/subtle#98, this is used to impl `subtle::ConditionallySelectable` for `Ordering` by delegating to the impl for `i8`.
- It turns out we can avoid a `match` here too in libsignal's FFI method to compare public keys, since our encoding of `Ordering` into `i<whatever>` in `ECPublicKey_Compare` matches the one from the rust stdlib.

### DIFF
--- a/rust/bridge/shared/src/protocol.rs
+++ b/rust/bridge/shared/src/protocol.rs
@@ -1,5 +1,5 @@
 //
-// Copyright 2021-2022 Signal Messenger, LLC.
+// Copyright 2021-2023 Signal Messenger, LLC.
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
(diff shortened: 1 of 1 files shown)


jrose-signal:
This is correct, and shorter, but I'm not sure it's better. It *is* the case that Rust orderings have the raw representations we want here (the ones established by `memcmp` oh so long ago), but you had to write down a comment to say that, because it's [not exposed in the docs](https://doc.rust-lang.org/std/cmp/enum.Ordering.html) (even though it is in the source, and stable). The compiler should absolutely optimize the first into the second anyway, [but it doesn't quite today??](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=61bb40f179272dcbda99f48c07c966f0). I'll file a Rust bug about that.

PR #513, opened 2023-02-15
Update podspec allow other platforms

### LINKED ISSUE #512: [Swift][Pod] The platform of the target (macOS 11.0) is not compatible with `LibSignalClient (0.21.1)`, which does not support `macOS`.
Hi, I'm having an iOS SwiftUI chat app that is using LibSignalClient. Everything works fine until I have a request to make this app work on MacOS (to build a desktop app, later implement it on windows).

At first, I tried to set multiple-platform on my current target, which caused a few problems since cocoapods doesn't support multiple-platform on a single target yet. Then I separated the target into new ones, specific to iOS and MacOS. My pod file looks like this (simplified)

```
def shared
	pod 'LibSignalClient', git: 'https://github.com/signalapp/libsignal.git'
	pod 'SignalCoreKit', git: 'https://github.com/signalapp/SignalCoreKit.git'
end

target 'MyApp' do
	# Pods for iOS
	platform :ios, '15.0'
	shared
end

target 'MyApp-mac' do
	# Pods for Mac
	platform :osx, '11.0'
	shared
end
```

Then I run pod install and got this error:

```
The platform of the target `MyApp-mac` (macOS 11.0) is not compatible with `LibSignalClient (0.21.1)`, which does not support `macOS`.
```

Does this mean it's impossible to develop this idea using SignalProtocol? Otherwise, what should I do next?

EddieLukeAtmey:
resolve #512

### DIFF
--- a/LibSignalClient.podspec
+++ b/LibSignalClient.podspec
@@ -14,7 +14,10 @@ Pod::Spec.new do |s|
   s.source           = { :git => 'https://github.com/signalapp/libsignal.git', :tag => "v#{s.version}" }
 
   s.swift_version    = '5'
-  s.platform         = :ios, '12.2'
+  s.osx.deployment_target     = '10.13'
+  s.ios.deployment_target     = '12.2'
+  s.tvos.deployment_target    = '11.0'
+  s.watchos.deployment_target = '4.0'
 
   s.dependency 'SignalCoreKit'
 


jrose-signal:
Thanks, but there's a little more than just that: you can see in the build settings section that we have to do a manual translation to Rust target names based on the platform and architecture. There are already names for macOS set up, so that should work in theory, but tvOS and watchOS should probably be subsetted out, at least for now.

I'm also realizing that LibSignalClient depends on SignalCoreKit, which doesn't declare support for macOS either, but I think that's mostly just for logging, so maybe that dependency can be moved to a subspec.

I suggest testing this locally to make sure it all works, i.e. `pod lib lint --platforms=macos` should in theory pass once you've built the Rust FFI library.

PR #520, opened 2023-04-25
bump boring in Cargo.lock

selfisekai:
solves https://github.com/signalapp/boring/issues/18


jrose-signal:
I've already done this on our private branch; it will be part of the next release. Thank you!

PR #524, opened 2023-05-21
Add Automatic-Module-Name attribute to jar manifest

AsamK:
This allows using the libraries from java modules.

https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_modular_auto

### DIFF
--- a/java/client/build.gradle
+++ b/java/client/build.gradle
@@ -46,6 +46,12 @@ java {
     withJavadocJar()
 }
 
+tasks.named('jar') {
+    manifest {
+        attributes('Automatic-Module-Name': 'org.signal.libsignal')
+    }
+}
+
 processResources {
     // TODO: Build a different variant of the JNI library for server.
     dependsOn ':makeJniLibrariesDesktop'


jrose-signal:
\*nods\* This is pretty minimal maintainence burden (compared to sitting down and thinking about how we'd actually want to slice up modules for libsignal). What's your reasoning behind making libsignal-client and libsignal-server use the same module name, though? You certainly can't have them both in the same program, but they're also not interchangeable. (And I really am asking; I'm not familiar with Java Modules or any idioms for dealing with these kinds of "variant" products.)

AsamK:
I'm also no expert with Java Modules. There is a recommendation to use the main package name as module name and both libsignal-client and libsignal-server use the same package name.
As you say both can't be used in the same project, I thought it's ok to use the same module name.

That said I think it would also be ok to use `org.signal.libsignalserver` for the server.

https://stackoverflow.com/questions/43192741/how-should-i-name-my-java-9-module
https://mail.openjdk.org/pipermail/jpms-spec-experts/2017-May/000687.html
```
  - Strongly recommend that all modules be named according to the reverse
    Internet domain-name convention.  A module's name should correspond
    to the name of its principal exported API package, which should also
    follow that convention.  If a module does not have such a package, or
    if for legacy reasons it must have a name that does not correspond to
    one of its exported packages, then its name should at least start
    with the reversed form of an Internet domain with which the author is
    associated.
```

AsamK:
How about just adding the Automatic-Module-Name to the client library for now?

jrose-signal:
That's a good compromise. Sorry for letting this fall by the wayside.

jrose-signal:
Included in the 0.32.0 release. Thank you for your patience!

PR #532, opened 2023-08-04
Use explicit rev for dalek

MarcusGrass:
Using an explicit rev makes `zkgroup` importable now that dalek has a 4.0.0 release. It becomes unpatchable without the rev since you'll just get 4.0.0 and an unresolveable dependency availability issue.

Besides, explicit revs are a bit more stable than branch revs, although the lock-file is checked in so it doesn't matter that much

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -680,7 +680,7 @@ dependencies = [
 [[package]]
 name = "curve25519-dalek"
 version = "3.2.1"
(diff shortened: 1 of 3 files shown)


moiseev-signal:
Thanks for reporting the issue!

We have reverted the update to 4.0.0 in lizard2 branch thus making it compatible with all prior versions of libsignal, tagged its latest rev with libsignal-v0.30.2, and updated libsignal to use the tag instead of a branch. We believe this should address the problem you were experiencing.

We will also be using tags for curve25519-dalek fork updates going forward.

PR #539, opened 2023-10-22
Fix wrong shared secret length in KEM

mseewer:
The shared_secret_length function returned the wrong length.


jrose-signal:
Yikes, thanks! Good thing we're not using that particular function (it's just there for completeness). I'll get this into the next release.

jrose-signal:
Pulled into v0.33.0 with some tests as https://github.com/signalapp/libsignal/commit/8c27b74032a38846cba48224fccbb46608e30db4. Thanks again!

PR #561, opened 2024-03-02
Update yanked dependency

julianbraha:
The current version of `iana-time-zone` in `Cargo.lock` is 0.1.59, which was yanked. This PR updates it to the next available version, 0.1.60.


jrose-signal:
Thanks, we'll make sure to include this in our next general dependency update (if not your commit, then at least making sure this crate gets updated).

jrose-signal:
Included in v0.42.0!

PR #565, opened 2024-03-16
bridge: only use cpufeatures on iOS

Kladki:
The `cpufeatures` crate only supports a limited number of architectures, and since a comment explicitly stated that this is for iOS optimization, and since [the same is done for jni](https://github.com/signalapp/libsignal/blob/main/rust/bridge/jni/Cargo.toml#L31), I think this is a sensible change.

### DIFF
--- a/rust/bridge/ffi/Cargo.toml
+++ b/rust/bridge/ffi/Cargo.toml
@@ -32,8 +32,10 @@ signal-pin = { path = "../../pin" }
 signal-media = { path = "../../media" }
 libsignal-bridge = { path = "../shared", features = ["ffi"] }
 async-trait = "0.1.41"
-cpufeatures = "0.2.1" # Make sure iOS gets optimized crypto.
 futures-util = "0.3"
 rand = "0.8"
 log = { version = "0.4", features = ["release_max_level_info"] }
 log-panics = { version = "2.1.0", features = ["with-backtrace"] }
+
+[target.aarch64-apple-ios.dependencies]
+cpufeatures = "0.2.1" # Make sure iOS gets optimized crypto.


Kladki:
I have signed the agreement now.

jrose-signal:
What's the motivation for making this change? This isn't the only dependency on cpufeatures in the workspace; it's just setting a minimum version before where Apple aarch64 CPUs weren't correctly considered as having unconditional support for…something, I forget what.

Kladki:
I am the maintainer of the [mautrix-signal](https://go.mau.fi/mautrix-signal/) package for [Alpine Linux](https://alpinelinux.org/), which depends on libsignal. We build for many architectures, but some of them, such as riscv64 and ppc64le, are being blocked by the cpufeatures crate, since it doesn't support those architectures.

> This isn't the only dependency on cpufeatures in the workspace

This is true, however, the ones that do only use them on architectures that support it, see:

https://lib.rs/crates/sha1
https://lib.rs/crates/aes
https://lib.rs/crates/argon2
https://lib.rs/crates/chacha20
https://lib.rs/crates/curve25519-dalek
https://lib.rs/crates/poly1305
https://lib.rs/crates/polyval
https://lib.rs/crates/sha2

jrose-signal:
A fair point. Obligatory "Signal the organization doesn't support third-party clients and use with a bridge" and "libsignal the library doesn't support direct use of the entry points in libsignal-ffi (they are even more subject to change than the non-bridge Rust crates and the Swift API built on top of them)".

You didn't quite get the iOS triple right, but we can take care of that.

Kladki:
> Obligatory "Signal the organization doesn't support third-party clients and use with a bridge" and "libsignal the library doesn't support direct use of the entry points in libsignal-ffi (they are even more subject to change than the non-bridge Rust crates and the Swift API built on top of them)".

All good :+1: 
> You didn't quite get the iOS triple right, but we can take care of that.

Really? Because I checked against [the rustc book](https://doc.rust-lang.org/rustc/platform-support/arm64e-apple-ios.html) and it looks correct. :thinking:

jrose-signal:
That's *an* iOS triple, but not the normal one, which is just aarch64-apple-ios. arm64e is the experimental variant that enables pointer tagging checks, which Apple hasn't finalized the ABI for.

Kladki:
Thanks for pointing that out, it is now fixed. :+1:

jrose-signal:
Included in v0.42.0!

PR #568, opened 2024-04-05
Fixed use of a wrong function

dtimoshenko99:
PreKeyBundle methods preKeyId and signedPreKeyId used the same function call to signal_pre_key_bundle_get_signed_pre_key_id, so they returned same values. Fixed this, so that preKeyId would use signal_pre_key_bundle_get_pre_key_id.


jrose-signal:
Oh no, thank you! You can see Signal-iOS doesn't actually extract these values after they've built the PreKeyBundle. 😅

dtimoshenko99:
Ahh, I see, is it a valid pr anyway then? 😅

jrose-signal:
Yes, thank you, we'll fix it on our internal release branch. We shouldn't have blatantly incorrect functions exposed in the API even if they're not currently in use. :-)

dtimoshenko99:
Thank you!

jrose-signal:
Fixed in v0.45.0!

PR #590, opened 2024-10-02
Added devcontainer files

8r4n:
I added devcontainer files for users of vscode. This enables users to capture the dependencies as code rather than README instructions.


jrose-signal:
Thanks for the PR! I think since none of us at Signal develop this way, and the container being set up would need to be further customized anyway if you're going to build the Java, Node, or Swift parts, we're not going to take this at this time.

PR #599, opened 2024-12-25
ffi change: Enable SessionRecord_NewFresh, change to cdylib

alexsunday:
Hi,

I am planning to wrap `libsignal` using other languages through FFI and noticed that the `rust/bridge/ffi` part seems to be unmaintained, with the focus primarily on Java/Node/Swift. In the package `ffi`, the `crate-type` is defined as `staticlib`, which doesn't seem to be the usual way FFI is used. Moreover, some methods are not exported only in `ffi` package. I would like to propose changes in this regard, thank you.

For the first simple change, I would like to export `SessionRecord_NewFresh` so that other programming language wrappers for libsignal can construct an instance from scratch. This makes it easier to write test cases for `SessionRecord`.

Additionally, I would like to know if this project accepts external PRs. Does the adjacent storage-service project also accept PRs?

### DIFF
--- a/rust/bridge/ffi/Cargo.toml
+++ b/rust/bridge/ffi/Cargo.toml
@@ -15,7 +15,7 @@ workspace = true
 
 [lib]
 name = "signal_ffi"
-crate-type = ["staticlib"]
+crate-type = ["cdylib"]
 
 [features]
 # Testing the Swift side of this requires compiling with SIGNAL_MEDIA_SUPPORTED enabled for both Swift and C:
--- a/rust/bridge/shared/src/protocol.rs
+++ b/rust/bridge/shared/src/protocol.rs
@@ -925,7 +925,7 @@ fn CiphertextMessage_FromPlaintextContent(m: &PlaintextContent) -> CiphertextMes
     CiphertextMessage::PlaintextContent(m.clone())
 }
 
-#[bridge_fn(ffi = false, node = false)]
+#[bridge_fn(node = false)]
 fn SessionRecord_NewFresh() -> SessionRecord {
     SessionRecord::new_fresh()
 }


jrose-signal:
Welcome! I think you've missed the purpose of the `ffi` crate: it's the C interface through which the Swift APIs talk to the Rust implementations. Most of its contents come from libsignal-bridge-shared, and it's considered an implementation detail rather than rather than a product in its own right (read: we change it around fairly often without any regard to stability). You can see `swift/build_ffi.sh` for how it gets built and its symbols extracted. It is very much in use.

`SessionRecord_NewFresh` was indeed originally exposed to preserve some of the tests from our predecessor library libsignal-protocol-java, but it's not really needed in normal operation, which is why it didn't get exposed for the other bridges. You can't *do* anything with a fresh session; the only way to create valid ones is through the SessionCipher-related APIs. However, because the Android client used libsignal-protocol-java directly, we wanted to be sure that we at least handled such a session correctly. For iOS and Desktop, which previously used non-protobuf representations of sessions, we didn't have to worry about this case, and so we reduced complexity by not exposing it. I think that would apply to any new bridges as well. (Though had the situation been as you described it, it would indeed have been a nice contained place to start!)

As for why the crate is a `staticlib` rather than a `cdylib`, it's because the CocoaPods build will produce a dylib-based framework, and it's both tidier and simpler for the Xcode build system if there's not a *second* dylib to have to package into the iOS app. You could certainly build it as a dylib using `cargo rustc -p libsignal-ffi --crate-type cdylib`, but again, be warned that we consider it an implementation detail.

And for PRs, I'll copy in the relevant portion from our readme:

> Signal does accept external contributions to this project. However unless the change is simple and easily understood, for example fixing a bug or portability issue, adding a new test, or improving performance, first open an issue to discuss your intended change as not all changes can be accepted.
> 
> Contributions that will not be used directly by one of Signal's official client apps may still be considered, but **only if they do not pose an undue maintenance burden** or conflict with the goals of the project.
> 
> Signing a [CLA (Contributor License Agreement)](https://signal.org/cla/) is required for all contributions.

Unfortunately, I think treating the C FFI layer as a (semi-)stable product layer on par with the app languages would count as an "undue maintenance burden" for us, since it would restrict our ability to rearrange things to make the Swift layer better. Still, if you have specific things you need and they're not disruptive, we can consider taking them. (`storage-service` is likely to be even more conservative in taking changes that aren't used by Signal-in-production, but you can still ask. And indeed I see you did.)

alexsunday:
I reviewed the Rust code and found that signal_session_record_new_fresh essentially just creates an empty array of uint8. So, indeed, this is merely a formal issue, and it is entirely possible to create an empty object through signal_session_record_deserialize as well.

Thank you for informing me that the destination/target can be directly changed via  the cargo CLI options. so this commit is no longer necessary. I'll close it. thx.

PR #602, opened 2025-02-24
Avoid duplicate crates

rex4539:


rex4539:
@jrose-signal What do you think?

jrose-signal:
Thanks! We discussed whether we want this—previously we just did it for dependencies that were *already* used by multiple packages—and decided that, yes, we did. I'll rebase this on our internal release branch (with your authorship intact).

Issue #603, opened 2025-03-12
deserializing CiphertextMessage in ffi


saman3d:
We are working on a project using React Native and the libsignal library. Since React Native does not support native handles, we serialize CiphertextMessage to bytes in JavaScript, along with a type identifier, and then deserialize it on the native side based on that identifier.

This approach works well on Android, we use that type identifier to deserialize ciphertext into one of its implementations then use them as the Ciphertext interface in other methods. but on Swift code ciphertext is implemnted differently and can only be constructed from PlaintextContent (not from PrekeySignalMessage, SignalMessage, SenderKeyDistributionMessage ...)

We are facing this issue when trying to construct UnidentifiedSenderMessageContent, which requires a CiphertextMessage.

Is there an existing way to reconstruct CiphertextMessage from its underlying types on iOS?
If not, would you accept a pull request to add this functionality?

saman3d:
Yes thats right. every ciphertext that is constructed by the rust code is kept via a native handle in signal apps. so when passing these ciphertexts back into rust they are derefrenced into their original types. so no ser/de is needed.

However, in our react native app, the application is written in JS and runs on hermes engine which does not support JSI (unlike what you did in the desktop app). Because of this limitation, we use native modules to bridge JavaScript with native code (the bridge layer).

When calling native bridge functions from JavaScript, we must serialize every function argument, then deserialize it inside the native function, process it (e.g., by calling Signal Protocol JNI or FFI code), serialize the output, and return it back to JavaScript. So every ciphertext produced is serialized at the JS layer. and for further use like constructing UnidentifiedSenderMessageContent we need to deserialize it again. which is currently not possible on iOS.

jrose-signal:
Fair enough. I have a vague sense that CiphertextMessage should be validateably correct, so simply reconstructing it from a type and serialized buffer wouldn't guarantee that, but UnidentifiedSenderMessageContent does not have that guarantee (since the same type is used for both sending and receiving). I might be splitting hairs, but it'd still be enough for your purposes to expose a new constructor for USMC, yes? Something like `init(contents:messageType:from:contentHint:groupId:)`? If that's sufficient, we'd probably accept a PR for it.

saman3d:
Yes, that would work for us! I'll prepare a PR, thanks for your help!

PR #604, opened 2025-03-15
add usmc constructor from content and type

saman3d:
regarding #603 

Added an FFI bridge function and a Swift constructor for UnidentifiedSenderMessageContent to allow its construction from type and content.
Included a simple test to verify functionality.

The message content verification is in a separate commit. It can be reverted or skipped if it's considered too expensive or unnecessary.


rust/bridge/shared/src/protocol.rs
```diff
@@ -845,6 +845,48 @@ fn UnidentifiedSenderMessageContent_New(
...
+        }
+        CiphertextMessageType::SenderKey => {
+            SenderKeyMessage::try_from(message_content).err()
+        }
+        CiphertextMessageType::Whisper => {
+            SignalMessage::try_from(message_content).err()
+        }
+    } {
+        return Err(SignalProtocolError::InvalidArgument(format!("invalid message content")));
+    }
```
jrose-signal, line 875: I think I would just skip this step, it turns a relatively simple operation where the only O(N) work is a memcpy to one that's doing protobuf parsing, even if the protobuf parsing isn't *that* complicated. (Any sealed sender message *delivered* to you can turn out to have garbage inside if the sender isn't playing nice.) If you feel strongly about it, though, the `let Some(_)` part is superfluous; the `match` arms can use `is_err()` instead for the same effect.

swift/Tests/LibSignalClientTests/SessionTests.swift
```diff
@@ -624,6 +624,101 @@ class SessionTests: TestCaseBase {
         let bob_session_with_alice = try XCTUnwrap(bob_store.loadSession(for: alice_address, context: NullContext()))
         XCTAssert(try bob_session_with_alice.currentRatchetKeyMatches(XCTUnwrap(bob_error_message.ratchetKey)))
     }
+
+    func testUnidentifiedSealedSenderContentFromContentAndType() throws {
```
jrose-signal, line 628: I think rather than make this a whole separate test, it'd be simpler / more maintainable going forward to insert the new cases inline in the previous test, and then assert that exactly the same bytes get produced when serializing the overall message no matter which constructor gets used. What do you think?

jrose-signal (commented):
Thank you for the tidy and thorough PR, and for including tests. :-)

rust/bridge/shared/src/protocol.rs
saman3d: I thought there was an api API constraint preventing the malformed creation of usmc. now that i double check there's not strong enforcement on this part. I'll remove this validation step. (Thanks for the suggestion is_err is much cleaner! :))

swift/Tests/LibSignalClientTests/SessionTests.swift
saman3d: added serialize method to usmc (ffi function was available) so that i can compare their serialized bytes after creation and asserted that both methods of constructing usmc produced equal ciphertext bytes after encryption.

saman3d: I falsely assumed that identical ciphertexts could be created. I reverted the test changes and only added the serialization assertion. (sorry for multiple reverts)

jrose-signal: Yeah, the sealed sender encryption uses an ephemeral key, and the exposed API doesn't let you control the generation of that key. What you have now looks good!

jrose-signal:
Merged as f536bf74c8f4d0587310f31a9f5c9dabb6019bff in our upcoming branch, thank you!

ImAwsumm (approved):

jrose-signal:
And present in v0.68.0.

PR #608, opened 2025-04-05
swift: add deserialize usmc constructor

saman3d:
As a follow-up to #603 and #604, this adds a deserialization constructor for usmc, enabling its serialization/deserialization across the bridge layer.


jrose-signal:
Reasonable enough! Can you add a simple round-trip test as well?

saman3d:
Thanks for the review! added test to verify that the USMC object remains equal to the original after serialization and deserialization.

jrose-signal:
Thanks, merged into our prerelease branch as 41dd135238fe86cf15d4febc15a60653fb98123e!

Issue #490, opened 2022-10-11
[Feature request] Release on crates.io


werdahias:
I intend to package libsignal for debian. This would require this crate being on crates.io. I'd appreciate it if you could publish it.

Cheers

jrose-signal:
We can certainly consider releasing on crates.io, but I don't think the project is well-suited for a Debian package. As noted in the readme, we don't promise any support for use outside Signal; in particular, "all APIs and implementations are subject to change without notice". We maintain protocol compatibility across our range of currently supported clients, but that's all we guarantee. Combined with Debian's stringent update requirements that leads to packages falling behind their source projects, and you could pretty quickly have something that's no longer consistent with what Signal itself uses. Of course, libsignal can be used as a building block for other programs and protocols, but without compatibility promises updating from version to version has to be understood as a manual step, and making libsignal a standard Debian package dilutes that message somewhat. (It's also a little weird that that package wouldn't come from us, the Signal Foundation.)

Let's take a step back: what are you actually hoping to accomplish?

werdahias:
I intend to package Flare in the long run ( a unofficial GTK Signal client). As it uses libsignal as dependency I'd need to package that first (debian policy calls for no vendored dependencies) . I agree that by the time libsignal would make it into debian stable it might be out of date/incompatible. But because Ubuntu pulls ~70% of the new debian packages in testing, it would be made available for a lot of endusers/developers.

gferon:
I would really love to see `libsignal` on crates.io as it would allow us to publish crates that make use of it, and it's pretty easy to update and make sure everything stays compatible with the official Signal ecosystem.

However, I feel like @jrose-signal really has a point with Debian packages, and I don't see it very fitting either.

werdahias:
The release on crates.io is somewhat separate from the package. A debian/ubuntu rust package provides the same files as in the repo, e.g. `src/lib.rs`. As in regards to possible API breaks: Since most of the users/possible developers in the debian ecosystem run some version of ubuntu because of the newer packages (or debian testing/unstable) I don't see this as a big issue.

werdahias:
Any update? Or are you **not** going to release it?

jrose-signal:
We're still considering, and it's only one priority amidst many. There are a number of things working *against* putting this on crates.io, such as the usual complexity around multiple crates and the dependence on forks of other crates (curve25519-dalek and boring).

[...]

PR #631, opened 2025-09-28
Add Crates.io publishing CI action

BrunoBernardino:
This action also does a dry run if the action isn't triggered manually or from a release.

I hope this PR is not too presumptuous, but I was hoping that having this CI action done for you would nudge you into closing #490.

What's missing here are things I can't do for you, mainly ensuring the requirements from [the publishing docs](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html) are met (I think it's just missing you creating an account, but I'm not 100% sure), as well as [the docs from Trusted Publishing](https://blog.rust-lang.org/2025/07/11/crates-io-development-update-2025-07/#trusted-publishing) (mainly adding this repo for the publishing).

Let me know if there's anything I can do to make this easier for you!

### DIFF
--- a/.github/workflows/crates.yml
+++ b/.github/workflows/crates.yml
@@ -1,0 +1,34 @@
+name: Publish to crates.io
+
+on:
+  workflow_dispatch:
+  push:
+  release:
+
+jobs:
+  publish:
+    environment: release
+    
+    permissions:
+      # Needed for rust-lang/crates-io-auth-action
+      id-token: write
+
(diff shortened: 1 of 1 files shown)


jrose-signal:
The drawbacks of and roadblocks against publishing on crates.io mentioned in #490 still stand, so we won't take this at this time. If we ever change our minds, though, it'll be helpful to have this ready.

PR #643, opened 2026-01-14
refactor: validate_with_trust_roots without indirection

rubdos:
This refactors SenderCertificate::validate_with_trust_roots to take `&[PublicKey]` instead of the double indirection `&[&PublicKey]`. Originally, this seems to have been introduced to be more FFI-friendly, but is rather inconvenient to work with from the Rust-API side.

This does, however, require that FFI-calls reallocate the incoming PublicKey list, so I understand if this is not a desirable change. Feel free to suggest an alternative approach; maybe the FFI side can already supply the `&[PublicKey]` with minimum effort?

For an example for the Rust context, see https://github.com/whisperfish/libsignal-service-rs/pull/385#issuecomment-3748942962


jrose-signal:
Included in v0.86.15.

Issue #659, opened 2026-03-17
Stricer use of the Rust type system


cvengler:
Hey, in accordance with the README:
> unless the change is simple and easily [...] first open an issue to discuss your intended change as not all changes can be accepted.

I read a bit in the source code, most notably `rust/core` and `rust/crypto` and figured out that the codebase makes lots of use with `&[u8]` slices for things where the size is already known at compile-time, most notably public and secret keys.
This pattern paints through various parts of the code base and I think we should change these occurrences from either `&[u8]` to `&[u8; KNOWN_SIZE]` (or the copying variant for non-secrets) or preferably a higher-level wrapper type, such as by exposing the types from dalek-crypto, as they also provide a good zeroization strategy.

I am personally not a fan of the current pattern.
Not just because semantic gets lost, as `&[u8]` can basically mean anything, but also because it introduces runtime bloat that is probably not impactful but still avoidable at compile time.

Any thoughts or feedback?
Is this worth pursuing further?

PR #662, opened 2026-04-02
feat(hpke): add hybrid ML-KEM 1024 + X25519 HPKE cipher suites

pervrosen:
## Summary

- Closes the last PQC gap in libsignal: the HPKE layer in `signal-crypto` previously supported only the classical X25519 variant. PQXDH and SPQR were already PQC-hardened; this brings HPKE to the same level.
- Adds two hybrid cipher suites combining ML-KEM 1024 (NIST FIPS 203, Category 5) with X25519, so security holds as long as either primitive is unbroken — protecting against "harvest now, decrypt later" attacks.
- No self-implemented crypto. All primitives via existing trusted libraries: `libcrux-ml-kem 0.0.8` (Cryspen, formally verified via HAX), `libsignal_core::curve`, `hkdf`+`sha2`, `signal-crypto` AES-256-GCM.

## New cipher suite variants

| Discriminant | Cipher suite |
|---|---|
| `2` | `Hybrid_MlKem1024_X25519_HkdfSha256_Aes256Gcm` |
| `3` | `Hybrid_MlKem1024_X25519_HkdfSha512_Aes256Gcm` |

Wire format: `[type(1)] [mlkem_ct(1568)] [x25519_enc(32)] [aead_ct(n+16)]`

Key derivation: `HKDF(ikm = ss_mlkem ‖ ss_x25519, info = label ‖ [0x00/0x01] ‖ info)`

Secret material (`ikm`, AEAD key/nonce, seeds) wrapped in `Zeroizing`.

## NIST compliance

FIPS 203 (ML-KEM 1024), SP 800-186 (X25519), SP 800-56C (HKDF), SP 800-227 (hybrid KEM), FIPS 197 (AES-256-GCM).

## Test plan

- [ ] 26 new tests in `rust/crypto/src/hpke.rs` covering: round-trips (SHA-256 and SHA-512), empty plaintext, wire format validation, authentication failures (wrong key, tampered ML-KEM ct, tampered X25519 enc, tampered AEAD payload, wrong AAD, wrong info), hybrid combiner properties (each shared secret is load-bearing), HKDF domain separation (key ≠ nonce, SHA-256 ≠ SHA-512), ML-KEM production integration, and X25519 regression
- [ ] `cargo test -p signal-crypto` — all 26 pass, zero failures

### DIFF
--- a/rust/crypto/src/hpke.rs
+++ b/rust/crypto/src/hpke.rs
@@ -4,6 +4,7 @@
 //
 
 use hpke_rs::prelude::*;
+use zeroize::Zeroizing;
 
 mod provider;
 
@@ -16,6 +17,8 @@ pub use hpke_rs::HpkeError;
 #[try_from(repr)]
 pub enum SignalHpkeCiphertextType {
     Base_X25519_HkdfSha256_Aes256Gcm = 1,
+    Hybrid_MlKem1024_X25519_HkdfSha256_Aes256Gcm = 2,
+    Hybrid_MlKem1024_X25519_HkdfSha512_Aes256Gcm = 3,
 }
 
 impl From<SignalHpkeCiphertextType> for u8 {
@@ -37,6 +40,8 @@ impl SignalHpkeCiphertextType {
     fn mode(self) -> HpkeMode {
         match self {
             SignalHpkeCiphertextType::Base_X25519_HkdfSha256_Aes256Gcm => HpkeMode::Base,
+            // Hybrid variants bypass hpke_rs entirely and never call set_up().
+            _ => unreachable!("hybrid variants do not use hpke_rs set_up()"),
         }
     }
 
@@ -45,6 +50,8 @@ impl SignalHpkeCiphertextType {
             SignalHpkeCiphertextType::Base_X25519_HkdfSha256_Aes256Gcm => {
                 hpke_types::KemAlgorithm::DhKem25519
             }
+            // Hybrid variants bypass hpke_rs entirely and never call set_up().
+            _ => unreachable!("hybrid variants do not use hpke_rs set_up()"),
         }
     }
 
@@ -53,6 +60,8 @@ impl SignalHpkeCiphertextType {
             SignalHpkeCiphertextType::Base_X25519_HkdfSha256_Aes256Gcm => {
                 hpke_types::KdfAlgorithm::HkdfSha256
             }
+            // Hybrid variants bypass hpke_rs entirely and never call set_up().
+            _ => unreachable!("hybrid variants do not use hpke_rs set_up()"),
         }
     }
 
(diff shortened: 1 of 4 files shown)


rust/crypto/src/hpke.rs
```diff
@@ -61,6 +70,151 @@ impl SignalHpkeCiphertextType {
...
+    variant: HybridHpkeVariant,
+}
+
+/// Private key for hybrid ML-KEM 1024 + X25519 HPKE decryption.
+///
+/// # Purpose
+/// Decrypts ciphertext produced by the corresponding `HybridHpkePublicKey::seal`.
+/// Handles both SHA-256 (type byte 2) and SHA-512 (type byte 3) variants automatically
+/// by inspecting the type byte in the ciphertext wire format.
+pub struct HybridHpkePrivateKey {
```
cvengler, line 167: I think we can and probably want to derive ZeroizeOnDrop here.

(I know that this pattern is not common across the code base but I don't think this should justify it IMO)

rust/crypto/src/hpke.rs
```diff
@@ -61,6 +70,151 @@ impl SignalHpkeCiphertextType {
...
+
+    /// Generates a hybrid key pair using HKDF-SHA-512 (~256-bit PQ security).
+    pub fn generate_sha512() -> Self {
+        Self::generate_with_variant(HybridHpkeVariant::HkdfSha512)
+    }
+
+    /// Internal generator that uses OS randomness for both X25519 and ML-KEM 1024 key generation.
+    fn generate_with_variant(variant: HybridHpkeVariant) -> Self {
+        use rand_core::RngCore as _;
+        let mut rng = rand::rng();
```
cvengler, line 203: I personally prefer SysRng due to the ThreadRng fork/state leak caveat.

Besides, SysRng is what I would expect when the doc comment says OS randomness, as ThreadRng is a rand abstraction, whereas SysRng offers raw OS access.

rust/crypto/src/hpke.rs
```diff
@@ -112,6 +266,73 @@ impl SimpleHpkeSender for libsignal_core::curve::PublicKey {
...
+    /// remains unbroken. The two shared secrets are combined via HKDF before AEAD encryption.
+    fn seal(&self, info: &[u8], aad: &[u8], plaintext: &[u8]) -> Result<Vec<u8>, HpkeError> {
+        use rand_core::RngCore as _;
+        let mut rng = rand::rng();
+
+        // ML-KEM 1024 encapsulation
+        let mut encaps_seed = Zeroizing::new([0u8; libcrux_ml_kem::ENCAPS_SEED_SIZE]);
+        rng.fill_bytes(encaps_seed.as_mut());
+        let (mlkem_ct, mlkem_ss) = mlkem1024::encapsulate(&self.mlkem, *encaps_seed);
+        debug_assert_eq!(mlkem_ct.as_ref().len(), MLKEM1024_CIPHERTEXT_LEN);
```
cvengler, line 286: You should not need the AsRef.
https://docs.rs/libcrux-ml-kem/latest/libcrux_ml_kem/mlkem1024/type.MlKem1024Ciphertext.html#method.len

rust/crypto/src/hpke.rs
```diff
@@ -161,6 +384,86 @@ impl SimpleHpkeReceiver for libsignal_core::curve::PrivateKey {
...
+        if ciphertext.len() < MIN_LEN {
+            return Err(HpkeError::InvalidInput);
+        }
+
+        let ciphertext_type: SignalHpkeCiphertextType = ciphertext[0]
+            .try_into()
+            .map_err(|_| HpkeError::UnknownMode)?;
+
+        // Parse wire format
+        let mlkem_ct_bytes = &ciphertext[1..1 + MLKEM1024_CIPHERTEXT_LEN];
```
cvengler, line 409: How about removing the +1 and using `..=` instead? Reads much cleaner IMO.

rust/crypto/src/hpke.rs
```diff
@@ -61,6 +70,151 @@ impl SignalHpkeCiphertextType {
...
+    }
+
+    /// Internal generator that uses OS randomness for both X25519 and ML-KEM 1024 key generation.
+    fn generate_with_variant(variant: HybridHpkeVariant) -> Self {
+        use rand_core::RngCore as _;
+        let mut rng = rand::rng();
+        let x25519_pair = libsignal_core::curve::KeyPair::generate(&mut rng);
+        let mut seed = Zeroizing::new([0u8; libcrux_ml_kem::KEY_GENERATION_SEED_SIZE]);
+        rng.fill_bytes(seed.as_mut());
+        let (mlkem_sk, mlkem_pk) = mlkem1024::generate_key_pair(*seed).into_parts();
```
cvengler, line 207: This allocates almost 4KB on the stack; I would prefer to `Box` them.

Also, you lose the `Zeroizing` here due to Deref but I feel the libcrux interface is unfortunately too flawed to do it better :/

rust/crypto/src/hpke.rs
```diff
@@ -112,6 +266,73 @@ impl SimpleHpkeSender for libsignal_core::curve::PublicKey {
...
+
+        // Ephemeral X25519 key agreement
+        let eph_pair = libsignal_core::curve::KeyPair::generate(&mut rng);
+        let ss_x25519 = eph_pair
+            .private_key
+            .calculate_agreement(&self.x25519)
+            .map_err(|_| HpkeError::InvalidInput)?;
+
+        // Combine secrets — hybrid combiner via HKDF; zeroized on drop
+        let mut ikm = Zeroizing::new(Vec::with_capacity(32 + 32)); // ss_mlkem (32) || ss_x25519 (32)
```
cvengler, line 296: I feel this type is better of in an array, given that its length is fixed.

rust/crypto/src/hpke.rs
```diff
@@ -112,6 +266,73 @@ impl SimpleHpkeSender for libsignal_core::curve::PublicKey {
...
+            HybridHpkeVariant::HkdfSha256 => {
+                SignalHpkeCiphertextType::Hybrid_MlKem1024_X25519_HkdfSha256_Aes256Gcm
+            }
+            HybridHpkeVariant::HkdfSha512 => {
+                SignalHpkeCiphertextType::Hybrid_MlKem1024_X25519_HkdfSha512_Aes256Gcm
+            }
+        };
+
+        // AES-256-GCM encrypt
+        let mut enc = crate::aes_gcm::Aes256GcmEncryption::new(&*aead_key, &*aead_nonce, aad)
```
cvengler, line 315: You are giving up on the `Zeroize` here by this, potentially introducing a leak of sensitive information, as `Deref` on `Zeroizing<[u8;32]>` returns you a copy of the inner `[u8; 32]` with the copy **not** being backed by `Zeroizing` anymore.

IMO the interface of aes_gcm is unfortunate in that regard.

Trying this should work.
```suggestion
        let mut enc =
            crate::aes_gcm::Aes256GcmEncryption::new(aead_key.as_ref(), aead_nonce.as_ref(), aad)
                .map_err(|_| HpkeError::InvalidConfig)?;
```

rust/crypto/src/hpke.rs
```diff
@@ -161,6 +384,86 @@ impl SimpleHpkeReceiver for libsignal_core::curve::PrivateKey {
...
+                .map_err(|_| HpkeError::InvalidInput)?;
+        let ss_x25519 = self
+            .x25519
+            .calculate_agreement(&x25519_pub)
+            .map_err(|_| HpkeError::InvalidInput)?;
+
+        // Re-derive AEAD key and nonce; ikm is zeroized on drop
+        let mut ikm = Zeroizing::new(Vec::with_capacity(32 + 32)); // ss_mlkem (32) || ss_x25519 (32)
+        ikm.extend_from_slice(mlkem_ss.as_ref());
+        ikm.extend_from_slice(&ss_x25519);
```
cvengler, line 433: The Zeroizing on ikm gives you nothing unfortunately if `ss_x25519` is a `Box<[u8]>` like it is right now.

rust/crypto/src/hpke.rs
```diff
@@ -161,6 +384,86 @@ impl SimpleHpkeReceiver for libsignal_core::curve::PrivateKey {
...
+                derive_hybrid_keys_sha256(&ikm, info)
+            }
+            SignalHpkeCiphertextType::Hybrid_MlKem1024_X25519_HkdfSha512_Aes256Gcm => {
+                derive_hybrid_keys_sha512(&ikm, info)
+            }
+            _ => return Err(HpkeError::UnknownMode),
+        };
+
+        // AES-256-GCM decrypt + authenticate
+        let mut dec = crate::aes_gcm::Aes256GcmDecryption::new(&*aead_key, &*aead_nonce, aad)
```
cvengler, line 446: See my note above on why this is insecure.

cvengler (changes requested):
My comments mostly center around incorrect use of `zeroize`.

I hope external reviews are accepted and appreciated, if not, feel free to discard :)

pervrosen:
> My comments mostly center around incorrect use of `zeroize`.
> 
> I hope external reviews are accepted and appreciated, if not, feel free to discard :)

Of course! Much appreciated. Give me time to work everything through.

rust/crypto/src/hpke.rs
pervrosen: Since this comment is correct and vital, I started with this problem. But then I realized I cannot use SysRng as it is not available in rand:0.9.4 (as it was introduce in a bump to 0.10). I can only use OsRng as you correctly noted. And OsRng from rand_core@0.9.3 implements TryRngCore (returns Result), not RngCore — so it can't be used with KeyPair::generate() or fill_bytes() directly. And now I feel get out on too deep waters. Do you have any thoughts on how to fix this (I am out on too deep water here) or should I delete the PR?

cvengler: I personally think it is fine to panic given that ThreadRng is doing the same internally when it fails to initialize using OsRng.

```rust
thread_local!(
    // We require Rc<..> to avoid premature freeing when ThreadRng is used
    // within thread-local destructors. See #968.
    static THREAD_RNG_KEY: Rc<UnsafeCell<ReseedingRng<Core, OsRng>>> = {
        let rng = ReseedingRng::new(THREAD_RNG_RESEED_THRESHOLD,
                                    OsRng).unwrap_or_else(|err|
                panic!("could not initialize ThreadRng: {}", err));
        Rc::new(UnsafeCell::new(rng))
    }
);
```

Besides, the only reason why OsRng may fail is if you are on a system that offers no system call but only access to `/dev/urandom`, in which case the error would indicate an I/O error of `/dev/urandom`.
I don't think this error can happen in reality on modern system, as almost every system includes something like a getrandom(2) system call, thereby making I/O access to `/dev/urandom` unnecessary.

pervrosen: Still I agree about your concern. Should I close this remark and work further on the rest you think? (I learn a lot so I do not mind at all.)

rust/crypto/src/hpke.rs
pervrosen: You are correct: 

286 -         debug_assert_eq!(mlkem_ct.as_ref().len(), MLKEM1024_CIPHERTEXT_LEN);
286 +        debug_assert_eq!(MlKem1024Ciphertext::len(), MLKEM1024_CIPHERTEXT_LEN);

The len() method is actually an associated function (no self), so the fix is MlKem1024Ciphertext::len() — no AsRef needed as you mentioned.

rust/crypto/src/hpke.rs
pervrosen: Is going from 

```
        let ss_x25519 = self
            .x25519
            .calculate_agreement(&x25519_pub)
            .map_err(|_| HpkeError::InvalidInput)?;

        // Re-derive AEAD key and nonce; ikm is zeroized on drop
        let mut ikm = Zeroizing::new(Vec::with_capacity(32 + 32)); // ss_mlkem (32) || ss_x25519 (32)
        ikm.extend_from_slice(mlkem_ss.as_ref());
        ikm.extend_from_slice(&ss_x25519);
```

to 

```
        let mut ss_x25519 = self
            .x25519
            .calculate_agreement(&x25519_pub)
            .map_err(|_| HpkeError::InvalidInput)?;

        // Re-derive AEAD key and nonce; ikm is zeroized on drop
        let mut ikm = Zeroizing::new(Vec::with_capacity(32 + 32)); // ss_mlkem (32) || ss_x25519 (32)
        ikm.extend_from_slice(mlkem_ss.as_ref());
        ikm.extend_from_slice(&ss_x25519);
        // Zero the Box<[u8]> before dropping to avoid leaking the shared secret
        zeroize::Zeroize::zeroize(&mut ss_x25519);
```

what you wanted? Please advise (since the learning surve is steep!)

rust/crypto/src/hpke.rs
pervrosen: Aes256GcmDecryption stores Aes256Ctr32 and GcmGhash which both contain the AES key material (round keys derived from the key). When dropped, neither struct has explicit zeroing, so the key material lingers in memory even though Zeroizing zeroes the source aead_key buffer.

However, this is an issue in the aes_gcm module itself, not something we can fix from hpke.rs. The fix would need to be:

```
// In aes_gcm.rs - add zeroing drop to Aes256GcmDecryption
impl Drop for Aes256GcmDecryption {
    fn drop(&mut self) {
        // Zero the GHASH state which contains the key-derived H value
        self.ghash.ghash = GHash::new(GenericArray::from_mut_slice(&mut [0u8; 16]));
        self.ghash.ghash_pad = [0u8; TAG_SIZE];
        self.ghash.msg_buf = [0u8; TAG_SIZE];
    }
}
```

Or wrapping the cipher state in Zeroizing. This is a separate fix in aes_gcm.rs, not hpke.rs. 
But I don't think it is wise for me as an outsider to touch this part of the code as all. Do you agree? Or have I misinterpreted the situation again? (Again a steep learning curve for me so have patience.)

rust/crypto/src/hpke.rs
pervrosen: I will use this.

rust/crypto/src/hpke.rs
cvengler: You don't have to close this PR, just add new commits that fix the existing one.

I don't think opening a new PR is a good idea as review context gets often lost there.

rust/crypto/src/hpke.rs
cvengler: I was thinking about wrapping the right-hand-side of `let mut ss_x25519` into `Zeroizing::new`.

rust/crypto/src/hpke.rs
cvengler: Please replace the `&*` in the parameters with `.as_ref()`.

I agree on the part with the aes_gcm module. It does not really embrace any sort of compile time safety really which is a bit sad given that Rust offers strong guarantees for that if used properly.

See #659

pervrosen:
Please look at the updated code.

moiseev-signal:
We do not have any plans to utilize or support this particular set of cipher suites in this manner, and keeping the codebase focused and constrained is an important design consideration too. Although this isn't a good fit for libsignal right now, we appreciate your interest in the project and willingness to help.

PR #668, opened 2026-04-19
doc: Remove `multilingual` key

cvengler:
This key was removed in mdBook 0.5.

According to the changelog:
> - Removed the `book.multilingual` field. This was never used.
>   [#2775](https://github.com/rust-lang/mdBook/pull/2775)

See also: <https://github.com/rust-lang/mdBook/pull/2775>

### DIFF
--- a/doc/book.toml
+++ b/doc/book.toml
@@ -1,4 +1,3 @@
 [book]
 language = "en"
-multilingual = false
 title = "libsignal"


cvengler:
Besides, it is currently impossible to run `mdbook serve` on current git-main using mdbook 0.5.2

jrose-signal:
Very belatedly pulled into v0.99.2 as c99362c, thank you for bringing it up.

PR #671, opened 2026-05-25
Migrate to upstream dalek

rozbb:
curve25519-dalek v5.0.0 will have Lizard encoding/decoding algorithms, which means libsignal can finally migrate to upstream dalek. This upgrade required some other relatively easy RustCrypto upgrades, namely `sha1`, `sha2`, `hmac`, and `hkdf`.

This PR is a work in progress. We're about to cut v5.0.0-rc.0. I'll mark it as ready once we release v5.0.0 proper.

cc @tarcieri


rust/account-keys/src/hash.rs
```diff
@@ -125,14 +125,14 @@ pub fn verify_local_pin_hash(encoded_hash: &str, pin: &[u8]) -> Result<bool> {
 #[cfg(test)]
 mod test {
     use const_str::hex;
-    use hmac::{Hmac, Mac};
+    use hmac::{HmacReset, KeyInit, Mac};
     use sha2::Sha256;
```
rozbb, line 129: Explanation: `Hmac` no longer has a `reset()` method. If you want to reset state, you need to use an `HmacReset`

rust/bridge/shared/types/Cargo.toml
```diff
@@ -85,10 +85,3 @@ jni-invoke-annotated = []
 extra-jni-checks = ["jni-type-tagging", "jni-invoke-annotated"]
 node = ["neon", "linkme", "signal-neon-futures"]
 metadata = ["linkme", "serde/derive"]
-
-[target.'cfg(not(any(windows, target_arch = "x86")))'.dependencies]
-# sha2's asm implementation uses standalone .S files that aren't compiled correctly on Windows,
-# and aren't linked correctly on x86 Android.
-# This will be fixed in sha2 0.11, which also removes the "asm" feature and turns it on by default.
-# So when sha2 0.11 is released, this section will go away.
-sha2 = { workspace = true, features = ["asm"] }
```
rozbb, line 97: Explanation: As mentioned in the comment, sha2 v0.11 no longer ships with ASM

rust/protocol/src/incremental_mac.rs
```diff
@@ -5,7 +5,7 @@
 
 use aes::cipher::Unsigned;
 use hmac::Mac;
-use hmac::digest::generic_array::{ArrayLength, GenericArray};
```
rozbb, line 8: Explanation: All of RustCrypto moved from `generic_array` to `hybrid_array` (exported as `array` by `digest`)

rust/zkgroup/src/crypto/profile_key_encryption.rs
```diff
@@ -77,38 +78,51 @@ impl ProfileKeyEncryptionDomain {
         let M4 = key_pair
             .decrypt_to_second_point(ciphertext)
             .map_err(|_| ZkGroupVerificationFailure)?;
-        let (mask, candidates) = M4.decode_253_bits();
+        let candidates = M4.map_to_curve_inverse();
```
rozbb, line 81: Explanation: These are the same function.

From Signal's [repo](https://github.com/signalapp/curve25519-dalek/blob/7c6d34756355a3566a704da84dce7b1c039a6572/curve25519-dalek/src/lizard/lizard_ristretto.rs#L87):

```rust
    pub fn decode_253_bits(&self) -> (u8, [[u8; 32]; 8]) {
        let mut ret = [[0u8; 32]; 8];
        let (mask, fes) = self.elligator_ristretto_flavor_inverse();


        for j in 0..8 {
            ret[j] = fes[j].as_bytes();
        }
        (mask, ret)
    }
```

From upstream [dalek](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b796f38c805311b296450f73a644732438d8ef9/curve25519-dalek/src/lizard/lizard_ristretto.rs#L210):

```rust
    pub fn map_to_curve_inverse(&self) -> [CtOption<[u8; 32]>; 16] {
        // Compute the inverses
        let fes = self.elligator_ristretto_flavor_inverse();
        // Serialize the field elements
        let it = fes.map(|fe| fe.map(|f| f.to_bytes()));
        Array::<_, U16>::from_iter(it).0
    }
```

The only difference is that the new function returns 16 elements instead of 8. This is because the old version of `elligator_ristretto_flavor_inverse` returned just the 8 positive solutions, whereas the new one returns 8 positive and 8 negative.

rust/zkgroup/src/crypto/profile_key_struct.rs
```diff
@@ -27,7 +27,7 @@ impl ProfileKeyStruct {
         encoded_profile_key[0] &= 254;
         encoded_profile_key[31] &= 63;
         let M3 = Self::calc_M3(Self::seed_M3(), profile_key_bytes, uid_bytes);
-        let M4 = RistrettoPoint::from_uniform_bytes_single_elligator(&encoded_profile_key);
+        let M4 = RistrettoPoint::map_to_curve(encoded_profile_key);
```
rozbb, line 30: Explanation: these are the same.

From Signal's [repo](https://github.com/signalapp/curve25519-dalek/blob/7c6d34756355a3566a704da84dce7b1c039a6572/curve25519-dalek/src/lizard/lizard_ristretto.rs#L26)
```rust
    pub fn from_uniform_bytes_single_elligator(bytes: &[u8; 32]) -> RistrettoPoint {
        RistrettoPoint::elligator_ristretto_flavor(&FieldElement::from_bytes(bytes))
    }
```

From upstream [dalek](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b796f38c805311b296450f73a644732438d8ef9/curve25519-dalek/src/ristretto/elligator.rs#L62):
```rust
    pub fn map_to_curve(bytes: [u8; 32]) -> RistrettoPoint {
        // MAP must clear the top bit. This is done in `from_bytes` for every field backend, so we
        // don't have to do it ourselves here
        let fe = FieldElement::from_bytes(&bytes);
        RistrettoPoint::elligator_ristretto_flavor(&fe)
    }
```

rust/account-keys/src/hash.rs
jrose-signal: We largely *don't* need `reset()`, it was just the convenient thing at the time.

rust/zkgroup/src/crypto/profile_key_encryption.rs
```diff
@@ -77,38 +78,51 @@ impl ProfileKeyEncryptionDomain {
...
-                    pk[31] |= 0x40;
-                }
-                let M3 =
-                    profile_key_struct::ProfileKeyStruct::calc_M3(seed_sho.clone(), pk, uid_bytes);
-                let candidate_retval = profile_key_struct::ProfileKeyStruct { bytes: pk, M3, M4 };
-                let found = M3.ct_eq(&target_M3) & is_valid_fe;
-                retval.conditional_assign(&candidate_retval, found);
-                n_found += found.unwrap_u8();
-            }
+            retval = retval.or_else(|| {
```
jrose-signal, line 93: Curious why you decided to rewrite using `or_else` instead of `conditional_assign` (both should have the correct behavior, of course).

rust/zkgroup/Cargo.toml
```diff
@@ -21,9 +21,7 @@ libsignal-core = { workspace = true }
 poksho = { workspace = true }
 zkcredential = { workspace = true, features = ["rayon"] }
 
-# Use our fork of curve25519-dalek for zkgroup support.
-curve25519-dalek-signal = { workspace = true, features = ["serde"] }
-
+curve25519-dalek = { workspace = true, features = ["lizard"] }
```
jrose-signal: Nitpick: now that this is a normal dependency we'll want to sort it normally!

jrose-signal (commented):
Aw, thanks! I suspect we'll want to do some extra integration steps on our end even beyond this (updating the remaining RustCrypto deps, maybe), but we can certainly take care of that using this as base, we won't bother you to keep adjusting the PR ad infinitum.

rust/zkgroup/src/crypto/profile_key_encryption.rs
rozbb: I tried but it was annoying. Conditional assign requires a decomposed CtOption (ie a value and a Choice). So we’d have to decompose all the values we generate, which is kinda messy. This solution is also messy in its own way though :)

rozbb:
Wonderful thank you! I’ll just update it then once 5.0.0 drops

rust/attest/src/snow_resolver.rs
```diff
@@ -3,11 +3,12 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
+use blake2::digest::Digest as _;
```
rozbb, line 6: Explanation: I have to do this because the current `blake2` version doesn't use `digest` v0.11 like `sha2` does. This will be resolved once `blake2` gets a release

rozbb:
dalek v5 dropped today!

jrose-signal:
(This is likely to hang out a bit longer as the rest of the RustCrypto updates drop too, to minimize duplicates in our dependency tree. Thank you again for setting it up though!)

rozbb:
Npnp, that should be my last edit. Feel free to build on this or throw it out if you have a refactor internally

jrose-signal:
Included in v0.97.5. Thank you again for, well, everything in this process! :-)

PR #672, opened 2026-05-25
Respect custom Cargo target dir in JNI build

### LINKED ISSUE #655: allow user-defined cargo target dir
i have, unsuccessfully, tried to build this with target-dir configured.

at least https://github.com/signalapp/libsignal/blob/main/java/build_jni.sh#L120 assumes the target dir to just be target.

`cargo metadata --format-version 1 --no-deps --offline | jq -r '.target_directory'` reliably finds the actual target directory.
i have changed that but still did not manage a successful build. something about a zip error and a jar not being found.

dannyward630:
## Summary
- resolve Cargo's actual target directory with `cargo metadata`
- copy desktop/server JNI libraries from that resolved directory instead of hard-coded `target/...`
- keep Android artifact handling unchanged because it already uses `--artifact-dir`

## Why
`java/build_jni.sh` respected Cargo's configured target directory during the build, but then copied desktop/server libraries from `target/$triple/...`. That breaks users who set `CARGO_TARGET_DIR` or configure `build.target-dir`, as reported in #655.

## Validation
- `bash -n bin/build_helpers.sh java/build_jni.sh`
- `git diff --check`
- fake-toolchain smoke: ran `java/build_jni.sh desktop` with a fake `cargo`/`rustc` and `CARGO_TARGET_DIR` set to a temp custom directory; the script copied `libsignal_jni_amd64.dylib` and `libsignal_jni_testing_amd64.dylib` from that custom target directory into `java/client/src/main/resources`

I could not run the full Rust/JNI build in this local environment because `cargo` is not installed on this machine.

Fixes #655.

### DIFF
--- a/bin/build_helpers.sh
+++ b/bin/build_helpers.sh
@@ -74,3 +74,8 @@ rust_remap_path_options() {
     echo -n "--remap-path-prefix ${prefix}= "
   done
 }
+
+cargo_target_directory() {
+  cargo metadata --format-version 1 --no-deps --offline |
+    python3 -c 'import json, sys; print(json.load(sys.stdin)["target_directory"])'
+}
--- a/java/build_jni.sh
+++ b/java/build_jni.sh
@@ -19,6 +19,7 @@ SERVER_LIB_DIR=java/server/src/main/resources
 # Fetch dependencies first, so we can use them in computing later options.
 # But allow this to fail in case we're offline.
 cargo fetch || true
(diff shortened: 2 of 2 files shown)


dannyward630:
Rechecked current head `87f7fff8` against issue #655. The JNI script resolves Cargo's target directory through `cargo metadata` while preserving Android artifact handling. `cla-server/validation` is green, the branch is mergeable, and there are no unresolved review comments; remaining step is maintainer review/merge.

PR #674, opened 2026-06-05
Fix poksho squeeze overflow

### LINKED ISSUE #480: poksho: an impossibly contrived overflow can theoretically happen
The `while i * HASHLEN < outlen` pattern in `squeeze_and_ratchet` for both `ShoSha256` and `ShoHmacSha256` can overflow if given an extremely large `outlen`, leading to a panic in `debug` mode or an infinite memory-eating loop in `release` mode. A tweaked implementation using `u32` and a test case that trips the overflow is available at https://github.com/jpdoyle/libsignal/blob/fb6a53de6c7b26863d1feab1d83c62291ac1be0d/rust/poksho/src/shosha256.rs#L131

So far as I can tell this code path is completely unreachable in any practical scenario, and is at worst able to cause denial-of-service. On 64-bit systems I think you use more RAM than has ever been manufactured (and certainly more than you can fit in most processors' physical address spaces) before it triggers.

The only case I can imagine is a system with a 32-bit `usize` which will still let you allocate a `(2^32 - 63)`-length `Vec` (unlikely unless `sizeof(void*) != sizeof(size_t)`), plus an input to `squeeze_and_ratchet` where `((outlen/HASHLEN)+1)*HASHLEN >= (1<<32)`.

I initially thought that if you have a poksho statement with 67108863 scalar parameters and all these other things line up, you can potentially trigger the overflow via https://github.com/signalapp/libsignal/blob/aa5f7c2b14315e94fa54fbdbda02b3938b218925/rust/poksho/src/statement.rs#L206 -- but it turns out that by including the encoded statement into the transcript, `to_bytes` gets called and that will fail an assert if there are more than 256 scalars.

Ap4sh:
## summary

fixes #480

- avoid `i * HASH_LEN` in the SHO squeeze loops
- iterate over the remaining output buffer instead, so the loop cannot overflow before the final block
- keep the same `u64` block counter encoding and final ratchet length

## testing

- `cargo test -p poksho`
- `cargo test -p poksho --all-features`
- `cargo test -p poksho --release`
- `cargo clippy -p poksho --all-targets --all-features -- -D warnings`
- `cargo clippy --workspace --all-targets --all-features --keep-going -- -D warnings`
- `git diff --check`

### DIFF
--- a/rust/poksho/src/shohmacsha256.rs
+++ b/rust/poksho/src/shohmacsha256.rs
@@ -65,13 +65,13 @@ impl ShoApi for ShoHmacSha256 {
         let outlen = target.len();
         let output_hasher_prefix =
             Hmac::<Sha256>::new_from_slice(&self.cv).expect("HMAC accepts 256-bit keys");
-        let mut i = 0;
-        while i * HASH_LEN < outlen {
+        let mut i = 0u64;
+        while !target.is_empty() {
             let mut output_hasher = output_hasher_prefix.clone();
-            output_hasher.update(&(i as u64).to_be_bytes());
+            output_hasher.update(&i.to_be_bytes());
             output_hasher.update(&[0x01]);
             let digest = output_hasher.finalize().into_bytes();
-            let num_bytes = cmp::min(HASH_LEN, outlen - i * HASH_LEN);
+            let num_bytes = cmp::min(HASH_LEN, target.len());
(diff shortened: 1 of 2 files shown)


jrose-signal:
Merged as 2c258202df009c4fcf3f7a6499df6faaa7399cff and released in v0.99.2, thank you!

PR #675, opened 2026-06-05
Reject oversized keytrans commitment inputs

Ap4sh:
## summary

- return a typed error when keytrans commitment inputs exceed the encoded length limits
- propagate that error through search verification instead of panicking
- keep the commitment bytes unchanged for valid inputs

## testing

- `git diff --check`
- `cargo test -p libsignal-keytrans`
- `cargo test -p libsignal-keytrans --all-features --release`
- `cargo clippy -p libsignal-keytrans --all-targets --all-features -- -D warnings`
- `cargo clippy --workspace --all-targets --all-features --keep-going -- -D warnings`


Ap4sh:
still relevant and ready for review

PR #678, opened 2026-06-11
poksho: Avoid theoretical overflow in squeeze_and_ratchet

### LINKED ISSUE #480: poksho: an impossibly contrived overflow can theoretically happen

celikgo:
Fixes #480.

## Problem

Both `ShoSha256` and `ShoHmacSha256` drive `squeeze_and_ratchet_into` with:

```rust
while i * HASH_LEN < outlen { ... }
```

For an extremely large `outlen`, `i * HASH_LEN` can overflow `usize`, panicking in debug builds or producing a runaway memory-eating loop in release builds. As noted in the issue, this is unreachable in any practical scenario (and at worst a DoS), but it's still a latent footgun.

## Fix

Iterate over `target.chunks_mut(HASH_LEN).enumerate()` instead. This removes the `i * HASH_LEN` multiplication entirely — and with it the `outlen - i * HASH_LEN` subtraction for the final partial chunk, which becomes just `chunk.len()`. The manual `split_at_mut` plumbing and the now-unused `std::cmp` import go away too.

## Why it's safe

- The block counter `i` fed into the hash (`(i as u64).to_be_bytes()`) is unchanged.
- The number of bytes copied per chunk is unchanged.
- An empty `target` yields zero chunks, matching the old `0 * HASH_LEN < 0` being false.

Output is therefore byte-identical. The existing `test_vectors` tests (which assert exact output for lengths 63/64/65/127/128/129 plus chained ratchets) still pass:

```
running 5 tests
test statement::tests::test_statement_encoding ... ok
test shosha256::tests::test_vectors ... ok
test shohmacsha256::tests::test_vectors ... ok
test sign::tests::test_signature ... ok
test statement::tests::test_complex_statement ... ok

test result: ok. 5 passed; 0 failed
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rust/poksho/src/shohmacsha256.rs
+++ b/rust/poksho/src/shohmacsha256.rs
@@ -3,8 +3,6 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 //
 
-use std::cmp;
-
 use hmac::{Hmac, Mac};
 use sha2::Sha256;
 
@@ -60,22 +58,20 @@ impl ShoApi for ShoHmacSha256 {
         self.mode = Mode::RATCHETED;
     }
 
-    fn squeeze_and_ratchet_into(&mut self, mut target: &mut [u8]) {
+    fn squeeze_and_ratchet_into(&mut self, target: &mut [u8]) {
         assert!(self.mode == Mode::RATCHETED);
         let outlen = target.len();
         let output_hasher_prefix =
             Hmac::<Sha256>::new_from_slice(&self.cv).expect("HMAC accepts 256-bit keys");
-        let mut i = 0;
-        while i * HASH_LEN < outlen {
+
+        // Iterating over chunks (rather than computing `i * HASH_LEN`) avoids a
+        // theoretical `usize` multiplication overflow for very large `outlen`.
+        for (i, chunk) in target.chunks_mut(HASH_LEN).enumerate() {
             let mut output_hasher = output_hasher_prefix.clone();
             output_hasher.update(&(i as u64).to_be_bytes());
             output_hasher.update(&[0x01]);
             let digest = output_hasher.finalize().into_bytes();
-            let num_bytes = cmp::min(HASH_LEN, outlen - i * HASH_LEN);
-            let (output, tail) = target.split_at_mut(num_bytes);
-            output.copy_from_slice(&digest[..num_bytes]);
-            target = tail;
-            i += 1
+            chunk.copy_from_slice(&digest[..chunk.len()]);
         }
 
         let mut next_hasher = output_hasher_prefix;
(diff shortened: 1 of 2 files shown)


celikgo:
Friendly nudge on this one — small, self-contained fix for #480. It removes the `i * HASH_LEN` overflow footgun in `squeeze_and_ratchet_into` by iterating over `chunks_mut(HASH_LEN).enumerate()`, and is verified byte-identical (the existing `test_vectors` exact-output tests still pass). Net −9 lines. Happy to adjust if you'd prefer a different approach. Thanks for taking a look!

x64x2:
The original code wasn't broken, it was correct. `i * HASH_LEN` is a trivial operation that will won't overflow in any practical scenario. The fix replaces an efficient loop with a more iterator based approach that does the exact same thing under the hood. 

The block counter _i_ is unchanged but  you're relying on `chunks_mut` to do the heavy lifting. That's fine until someone changes the chunk size or the hash length and now you've got bug that's hard to trace. The original code was explicit.

The test results are meaningless. Of course the output is identical, you're still hashing the same data the same way.

If you want to fix something, fix the actual problem, add a comment explaining why `i * HASH_LEN` will never overflow in practice. Don't replace working code with slower more convoluted hokum.

PR #680, opened 2026-06-21
protocol: fix ML-KEM parameter key type

Ap4sh:
## Summary

- fixes the ML-KEM 1024 `Parameters::KEY_TYPE` value so it matches `KeyType::MLKEM1024`
- extends the existing KEM parameter-constant test so each enabled KEM checks its static `KEY_TYPE` against the dynamic `KeyType` dispatch

## Notes

The ML-KEM implementation still used the Kyber1024 key type in its `Parameters` impl, while `KeyType::MLKEM1024` dispatches to that impl

`Parameters::KEY_TYPE` is used when KEM encapsulation or decapsulation errors are mapped into `SignalProtocolError`. This keeps ML-KEM 1024 self-consistent with that mapping

This does not change serialized keys, ciphertexts, or successful KEM behavior

## Test plan

- `cargo test -p libsignal-protocol`
- `cargo test -p libsignal-protocol --all-features`
- `cargo clippy -p libsignal-protocol --all-targets --all-features -- -D warnings`
- `cargo test -p libsignal-protocol --features mlkem1024 test_dyn_parameters_consts`
- `cargo test -p libsignal-protocol --all-features kem::tests`
- `git diff --check`

### DIFF
--- a/rust/protocol/src/kem.rs
+++ b/rust/protocol/src/kem.rs
@@ -674,21 +674,32 @@ mod tests {
 
     #[test]
     fn test_dyn_parameters_consts() {
-        assert_eq!(
-            kyber1024::Parameters::SECRET_KEY_LENGTH,
-            kyber1024::Parameters.secret_key_length()
-        );
-        assert_eq!(
-            kyber1024::Parameters::PUBLIC_KEY_LENGTH,
-            kyber1024::Parameters.public_key_length()
-        );
-        assert_eq!(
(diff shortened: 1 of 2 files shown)


jrose-signal:
Thank you, we've included this one in v0.96.3 as https://github.com/signalapp/libsignal/commit/6733123fc92fb37613f2293983d5a51ed713693b.

